PR #27, opened 2015-05-31
There are some bug fix and change

DelightRun:
Hi, thanks for your excellent project, it help me a lot.. However, since your os is windows(I guess), there are some inconvenient config in `wscript` file while I compile tiny-cnn on my Mac OS. Here's a change list:

+ update waf file to the newest waf-1.8
+ on *nix, many dynamic libs' header and dylib file are under `/usr/local/include/` and `/usr/local/lib/`,
   so we don't need to configure `BOOST_ROOT` and `TBB_ROOT`, just use `-ltbb` and `-lboost_timer`.
   But I still add a default value for *nix system.
+ add os detect and link to libc++ instead of libstdc++ on Mac OS
+ in my Mac OS, `--march=native` will lead to compile errors for both `gcc` and `clang`, so I removed it
+ mv data files to a new folder named `data`
+ fix some typo about `SSE` and `AVX` config in `wscript` under root folder
+ in *nix system, waflib fold is `.waf-*`(yes, there's a dot), so I add it to `gitignore`
+ use waf's `compiler_cxx` config instead of set `g++` manually

And up to now, tiny-cnn can compile and run on my Mac OS X smoothly. I think my config should work on most of Linux system as well. LOL~


nyanp:
@DelightRun 
Excellent! I don't know much about Mac OS and *nix environments, so it helps me a lot. 
Moving MNIST data to data/ dir is also good change. Thanks :)

DelightRun:
@nyanp 
You are welcome. How do you think about adding a new kind of softmax layer to tiny-can?

nyanp:
@DelightRun 
It's good if we can use softmax function, especially for the final layer. I'm wondering if we should implement it as a kind of layer (easy to implement), or activation-function (need for changing interfaces of activation::function::f and df, but it seems natural that tan-h, sigmoid, softmax belong same inheritance-tree). What do you think about it?

DelightRun:
Caffe, CXXNET and Theano are both implement Softmax as a new kind of Layer. And I think softmax function is not a kind of activation function, because it's result relied on the sum of all exp(wx+b) for each neuron's w, or in other words, other neuron's output

PR #46, opened 2015-10-13
added missing include in network.h

vgrabe:
added missing include of <iostream> to make std::cout known, fixes compiler error in line 142


nyanp:
@vgrabe thanks! :+1:

PR #47, opened 2015-10-13
Bug fix in layers.h

vgrabe:
Bug fix, shared pointer cannot be casted directly - fixes compiler error with example code from wiki


nyanp:
@vgrabe :+1:

PR #50, opened 2015-10-22
update util.h (isfinite)

shenck0:
the origin implement of isfinite will return true when x is -nan(ind)
test on vs2015 (compile in x64, release!)

### DIFF
--- a/tiny_cnn/util.h
+++ b/tiny_cnn/util.h
@@ -33,6 +33,7 @@
 #include <cassert>
 #include <cstdio>
 #include <cstdarg>
+#include <float.h>
 
 #ifdef CNN_USE_TBB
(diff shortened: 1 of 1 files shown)


patrikhuber:
You should not use `_isnan` and `_isfinite`, that looks highly non-standard (or C). Use `std::isnan` and `std::isfinite` instead.

Also, these functions are in `<cmath>`, not `<float.h>`.

nyanp:
@shenck0 @patrikhuber 
msvc2012 doesn't have ```std::isfinite```. They locate ```isfinite``` in global namespace, but it is non-standard way...This is why there's a ```tiny_cnn::isfinite``` (I want to support both gcc4.7- and msvc2012- in a potable manner).

PR #52, opened 2015-11-02
fixed include path in example wscript

balintfodor:


nyanp:
@balintfodor
nice! :)

PR #54, opened 2015-11-02
fix for signed / unsigned comparison error on GCC.

dicroce:


dicroce:
My gcc version is 4.8.4... The error I was getting was: 

../tiny_cnn/network.h:636:38: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
             for (int k=0;k<sizepatch*sizepatch;k++)
                                      ^
cc1plus: all warnings being treated as errors

nyanp:
@dicroce
Oh, I haven't noticed. Thanks! :)

PR #58, opened 2015-11-11
CMake support

rmsalinas:
Cmake for compiling test and train. Tested in msvc12, g++,mingw

Since the test and train programs using cmake might be in any locations, I have forced to use command line indicating the location of data directory.


nyanp:
@rmsalinas great works! :+1: I'll update build section in readme.md.

PR #67, opened 2016-01-06
Modified paths in includes, CMake

mrazekv:
Updated paths for include
Modified CMake to reflect these changes
Add OpenCV package requirement to CMakeList

The current version is not working on Linux with gcc+cmake, so I have modified the code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,10 @@ if(Boost_FOUND)
 ENDIF()
 
 
+
+find_package( OpenCV REQUIRED )
+
+
 IF(USE_TBB)
     find_package(TBB REQUIRED)
     add_definitions(-DCNN_USE_TBB)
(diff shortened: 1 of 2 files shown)


patrikhuber:
The changes to the paths to the headerfiles in CMakeLists.txt might indeed be necessary, but using "../" in include-paths can lead to problems and I think I would consider it bad practice.
[Related SO question](http://stackoverflow.com/questions/597318/what-are-the-benefits-of-a-relative-path-such-as-include-header-h-for-a-hea?lq=1). 

Having the includes as `#include "tiny_cnn/layers/fully_connected_layer.h"` is standard and it works, given the rest of the setup is working properly.

nyanp:
@mrazekv 
Thanks a lot for nice PR, and I agree with @patrikhuber . Could you rebase header files?

mrazekv:
@patrikhuber I agree, I have forgotten to set the direct paths. I'm not sure why the gcc had a problem with it. I will revert these files back in next week and then I will re-push these changes. Thank you for your work!

mrazekv:
@nyanp Hi, I have changed CMakeFile definition and opencv include in the pull request. I allso add include definition to prevent using "../" in includes. Everything works fine for me. I hope there will be no problem with pulling.

PR #74, opened 2016-01-19
Added assignment operator and copy constructor.

dicroce:


nyanp:
@dicroce
Thank you very much!

PR #77, opened 2016-01-24
Pull Request for ELU and Initialization

tolgabirdal:
I have found ELU activation units to be working quite good when initialized with the implemented method. The loss is multiclass cross entropy, which is optimized via Adam. I think it's useful to have it in tiny-cnn. Here is a summary:

- ELU (Exponential linear units) activation function added (http://arxiv.org/pdf/1511.07289v3.pdf).
- He et. al. weight initialization added (http://arxiv.org/pdf/1502.01852v1.pdf).
- for_ was raising a compile error on VC 2012 with TBB (due to type ambiguity). This is fixed via a cast operation.

### DIFF
--- a/tiny_cnn/activations/activation_function.h
+++ b/tiny_cnn/activations/activation_function.h
@@ -79,9 +79,8 @@ public:
 
 class elu : public function {
 public:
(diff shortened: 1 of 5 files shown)


nyanp:
@tolgabirdal 
ELU and suggested paper is very interesting. Thanks :)

PR #80, opened 2016-01-29
Fix COPY_PARAM() macro in layer_factory_impl.h

Abai:
- On OSX and Ubuntu 14.04 caffe_coverter.cpp does not compile due to
  invalid macro processing token error. This should fix it.


nyanp:
@Abai 
Nice fix, thanks! :)

PR #81, opened 2016-01-31
Some fixes and additions

pansk:
- fixed a potential bug in train_onebatch
- added constructor for rectangular window in convolutional_layer
- fixed hiding when overloading the out_size function in  convolutional_layer
- fixed hiding when overriding just one of the overloaded df method in activation_functions
- fixed the move semantics in a couple of places where move constructor/assignment were silently deleted
- added explicit deletion for classes containing reference
- removed a bunch of useless files


nyanp:
@pansk 
Oh, it's a big and greate PR! Thank you very much :)

PR #83, opened 2016-02-01
Fixes to previous PR #81

pansk:
My previous PR broke pre-VS2013 compatibility (VS2013 doesn't allow defaulted move constructor).
Added a small utility for this in util.h (defining CNN_DEFAULT_MOVE_CONSTRUCTOR_UNAVAILABLE and CNN_DEFAULT_ASSIGNMENT_OPERATOR_UNAVAILABLE in this case).

### DIFF
--- a/tiny_cnn/activations/activation_function.h
+++ b/tiny_cnn/activations/activation_function.h
@@ -35,9 +35,13 @@ class function {
 public:
     function() = default;
     function(const function &) = default;
+#ifndef CNN_DEFAULT_MOVE_CONSTRUCTOR_UNAVAILABLE
     function(function &&) = default;
+#endif
     function &operator =(const function &) = default;
+#ifndef CNN_DEFAULT_ASSIGNMENT_OPERATOR_UNAVAILABLE
     function &operator =(function &&) = default;
+#endif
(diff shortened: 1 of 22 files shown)


patrikhuber:
What about just dropping support for VS2012? It's getting pretty old. Most people should be on VS2015, or at least 2013. Would make the code simpler and more readable.

pansk:
+1

I just noticed that my comment is a bit ambiguous.
Defaulting move constructor/assignment is unavailable in VS2013 too (only VS2015 supports that C++11 feature).

stereomatchingkiss:
Please do not give up the support of vs2013, because cuda do not support vs2015 yet, it would be quite tedious to maintain two versions of lib/dll on windows platform.

pansk:
Same problem here, that's the main reason of this PR. Unfortunately it will drop the move semantics on these objects from VS2013 (but at the moment having move semantics for these objects is not important in this version of tiny-cnn).

nyanp:
+1
I want to keep support for vs2013 (portability is important for tiny-cnn) , but I think stop supporting vs2012 is acceptable. Thanks for nice PRs :)

PR #90, opened 2016-02-04
Fix type cast warning

stereomatchingkiss:
Fix another type cast warning, I do some change on for_  function in the util.h too, please do not merge that part since it is over complicated, just merge following line

` max_index = static_cast<int>(i);`

### DIFF
--- a/tiny_cnn/network.h
+++ b/tiny_cnn/network.h
@@ -171,7 +171,9 @@ public:
             if (optimizer_.requires_hessian())
                 calc_hessian(in);
             for (size_t i = 0; i < in.size(); i+=batch_size) {
-                train_once(&in[i], &t[i], std::min(batch_size, in.size() - i), n_threads);
+                train_once(&in[i], &t[i],
+                           static_cast<int>(std::min(batch_size, in.size() - i)),
+                           n_threads);
                 on_batch_enumerate();
 
                 if (i % 100 == 0 && layers_.is_exploded()) {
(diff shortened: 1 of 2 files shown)


pansk:
I'm not referring to constexpr or compile-time constants obtained via metaprogramming, but plain non-constexpr bool-typed function parameters or variables that are used as condition (so the condition is necessarily evaluated at run time). To be completely clear, I'm talking about this:

template<typename Func>  
void for_(bool parallelize, int begin, int end, Func f, int grainsize = 100) {  
    parallelize ? parallel_for(begin, end, f, grainsize) : xparallel_for(begin, end, f);  
}  

No metaprogramming and no constexpr are involved here, just a plain condition between parallel_for and xparallel_for. Here a if-else makes exactly the same job as a ?:, and actually ?: is supposed to be an expression, but here it's used as a statement (a non-expressions, that's what I find confusing), while if-else is clearly just a statement.

stereomatchingkiss:
Sorry for misunderstand your question.

As far as I remember, original codes prefer that solution, it is not my decision, maybe the author like that kind of style better than more common if...else. In my opinion, both of them are easy to read, it is just a matter of taste

pansk:
I confirm that the code is from the original code (not just from this commit), and as I said I didn't noticed it before (but it was there).
I just used this commit, which added or modified the code around one of these lines, to ask if there was a reason that I cannot see behind that choice.

nyanp:
@stereomatchingkiss 
Thank you very much! I've merged it :)

PR #91, opened 2016-02-04
Add travis-ci.org support

Abai:
- Compiles and runs tiny-cnn/test/test.cpp picotest
- Tests under linux with gcc-4.8 and clang 3.4
- Tests under OSX with Apple LLVM clang 6.0 (based on clang 3.5)
- Tests master branch only
- Uses CMake for config and compilation

**TODO:**
- Host homebrew-cache.tar.gz somewhere like s3 to speed up the build on
  OSX (Currently 15 minutes)
- The picotest exists with code 0 on failure. Should be code 1 for the
  build to break on travis-ci.
- Check why some picotests fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,51 @@
+sudo: false
+language: cpp
+os:
+  - linux
+  - osx
+compiler:
+  - gcc
+  - clang
+cache:
+  directories:
+  - $HOME/opencv_3_1_0 # Cache is only available on travis container-based boxes
+addons:
+  apt:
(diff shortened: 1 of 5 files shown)


Abai:
@nyanp I couldn't figure out whether you are planning to test on mnist or the test.cpp used in the PR. It's up to you to change it :)

nyanp:
@Abai 
Thank you very much!! I'm a beginner of Travis CI, so I'll read .travis.yml and official documents before changing :)

PR #92, opened 2016-02-08
Nyanp master

pansk:
Changed definition of max_index to use max_element
Introduced fprop_max and fprop_max_index protected functions, and their respective public counterpart predict_max_value and predict_label (better naming suggestions are welcome).
Added configuration option to opt-in the use of exceptions (CNN_USE_EXCEPTION, enabled by default in config.h), added inclusion of config.h from util.h.
Moved the inclusion of future and thread under a ifndef CNN_USE_OMP, to avoid warnings related to use of exception mechanism in VS2015 with exception handling disabled (and warning level 4).

### DIFF
--- a/tiny_cnn/config.h
+++ b/tiny_cnn/config.h
@@ -47,6 +47,11 @@
 //#define CNN_USE_OMP
 
 /**
+ * define to use exceptions
+ */
+#define CNN_USE_EXCEPTIONS
+
+/**
  * number of task in batch-gradient-descent.
  * @todo automatic optimization
  */
--- a/tiny_cnn/layers/layer.h
+++ b/tiny_cnn/layers/layer.h
@@ -260,7 +260,7 @@ private:
     }
 
     void set_size(layer_size_t in_dim, layer_size_t out_dim, size_t weight_dim, size_t bias_dim) {
-        try {
+        CNN_TRY {
             in_size_ = in_dim;
             out_size_ = out_dim;
 
@@ -275,7 +275,7 @@ private:
             for (auto& p : prev_delta_) p.resize(in_dim);
             for (auto& dw : dW_) dw.resize(weight_dim);
             for (auto& db : db_) db.resize(bias_dim);
-        } catch (const std::bad_alloc&) {
+        } CNN_CATCH (const std::bad_alloc&) {
             throw nn_error(
                 format_str("memory allocation failed: layer size too large!\nin:%d,out:%d,weights:%d,biases:%d",
                            in_dim, out_dim, weight_dim, bias_dim));
--- a/tiny_cnn/layers/layer.h
+++ b/tiny_cnn/layers/layer.h
@@ -260,26 +260,20 @@ private:
     }
 
     void set_size(layer_size_t in_dim, layer_size_t out_dim, size_t weight_dim, size_t bias_dim) {
-        CNN_TRY {
-            in_size_ = in_dim;
-            out_size_ = out_dim;
+        in_size_ = in_dim;
+        out_size_ = out_dim;
 
-            W_.resize(weight_dim);
-            b_.resize(bias_dim);
-            Whessian_.resize(weight_dim);
-            bhessian_.resize(bias_dim);
-            prev_delta2_.resize(in_dim);
+        W_.resize(weight_dim);
+        b_.resize(bias_dim);
+        Whessian_.resize(weight_dim);
+        bhessian_.resize(bias_dim);
+        prev_delta2_.resize(in_dim);
 
-            for (auto& o : output_)     o.resize(out_dim);
-            for (auto& a : a_)          a.resize(out_dim);
-            for (auto& p : prev_delta_) p.resize(in_dim);
-            for (auto& dw : dW_) dw.resize(weight_dim);
-            for (auto& db : db_) db.resize(bias_dim);
-        } CNN_CATCH (const std::bad_alloc&) {
-            throw nn_error(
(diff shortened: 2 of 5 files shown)


patrikhuber:
I don't like making exceptions optional and using these macros, it makes the code unnecessarily more complex and complicated to read. Not using exceptions is inadequate these days in my opinion, except for very special circumstances, and I'm having a hard time seeing tiny-cnn be applied in such circumstances. I prefer not having this additional "non-standard" abstraction layer. What's your use-case?

pansk:
I see your point, and I would have agreed with you until few months ago, before I started working in game industry, where disabling RTTI and exceptions is the rule of thumb.
Another solution I can propose is to remove this try-catch altogether and make tiny-cnn completely exception-agnostic: since the incriminated try-catch is the only one present in all the codebase, we could delegate the user's code to handle the bad_alloc exception and figure out what was wrong.

stereomatchingkiss:
>Not using exceptions is inadequate these days in my opinion, except for very special circumstances

+1, the rule of thumb of game industry do not mean it is the rule of thumb of convolution neural network.

Could pansk explain under what kind of circumstances, disable exception is a reasonable choice for tiny-cnn?

patrikhuber:
@pansk: Games industry, I see your point too ;-) Your other proposed solution doesn't sound too bad either, though the detailed error message would be lost in this case. Let's hear @nyanp's opinion.

micahpearlman:
I'm from the games industry as well. In the "old days" there was a very measurable penalty in performance and memory simply enabling exceptions and RTTI. Not clear on how much of a penalty it is with more modern (C++11) compilers. Considering that I'm potentially targeting ARM M4F series processors   where even a 2% loss of performance and memory is critical it would be nice to be able to turn off exceptions.

stereomatchingkiss:
>where even a 2% loss of performance and memory is critical it would be nice to be able to turn off exceptions.

Modern compiler use zero cost model to handle exception. If no exception happen, it would not cost you anything(that means, it could be more efficient than if...else solution if everything are fine), but you will pay enormous performance penalty if exception do happen.

Exception, as the term said, this kind of cases are rarely happen, is it worth to add more complexity just to handle those rare cases when modern compiler support zero cost model?

nyanp:
Thanks for PR, @pansk  :)
It's interesting topic whether or not exception handling in tiny-cnn should be optional.
Portability is very important characteristics of tiny-cnn, which is designed for hard circumstances (including embedded systems, IoT devices, and game industries), where memory footprint of RTTI and exception still matter.
I prefer @pansk 's 2nd solution, or just remove ```try...catch``` block. Macro solution is a bit tricky and it seems to be going away from c++11 standard way. Moreover, now this try block never throws ```std::bad_alloc```, because ```vec_t```'s custom allocator throws ```nn_error``` instead of ```bad_alloc``` (in ```aligned_allocator::allocate```). There is almost no disadvantage to remove this ```try...catch```, except for using user-defined custom allocator.

pansk:
Ok, later this evening I'll fix the code and do a new commit. Could we agree that no function should do exception handling? (throwing is good, it will fall to abort() when exceptions are disabled, and throws in tiny-cnn are used wisely, only to signal misconfiguration or other crazy things happening, so it's pretty safe).

On a different topic, I don't have the problems that AppVeyor, later I'll also verify my version of VS2013 wrt the one provided by AppVeyor.

Marco

pansk:
Please note that tiny_cnn_test project file contained in the repo is still compiling with VS2012, that's why appveyor is complaining. I'm committing the updated project file.

pansk:
Appveyor is now failing for a different reason (the lack of boost/filesystem.hpp)

nyanp:
@pansk 

> Could we agree that no function should do exception handling? (throwing is good, it will fall to abort() when exceptions are disabled, and throws in tiny-cnn are used wisely, only to signal misconfiguration or other crazy things happening, so it's pretty safe).

I agree with that ;) When someone disable exceptions, he/she should assert that such misconfiguration or out-of-memory don't occur. Does anyone else?

> On a different topic, I don't have the problems that AppVeyor, later I'll also verify my version of VS2013 wrt the one provided by AppVeyor.

Thank you very much, It's a great help for me :) I failed to configure AppVeyor and haven't enough time to fix it...

>Appveyor is now failing for a different reason (the lack of boost/filesystem.hpp)

boost is [pre-installed in Appveyor container](https://www.appveyor.com/docs/installed-software#languages-libraries-frameworks), So if you add ```C:\Libraries\boost_1_59_0``` into include path before build event, maybe it works (but I'm a newbie of AppVeyor and I have no idea how to configure it)

patrikhuber:
I would recommend you add an `appveyor.yml` file, and not configure AppVeyor via the GUI.

For the boost headers in AppVeyor, you can pass `-DBOOST_ROOT=C:\Libraries\boost_1_59_0` to the call to cmake. I think tiny-cnn also requires linking against some of the boost libraries (filesystem, timer), so an additional `-DBOOST_LIBRARYDIR=C:\Libraries\boost_1_59_0\lib64-msvc-14` is needed. I have a pretty simple `appveyor.yml` in [my repo](https://github.com/patrikhuber/4dface/blob/master/appveyor.yml), you can copy from it. I do a bit more fancy stuff there (e.g. clone submodules, get the newest cmake version, etc.), which tiny-cnn doesn't require, so if you strip it down, the whole config should be 10 lines only or so.

nyanp:
@patrikhuber 
Thanks for your help :) I've copied your appveyor.yml and it works fine.

PR #98, opened 2016-02-21
Switch test train phase

stereomatchingkiss:
1 : Provide api to switch between test phase and train phase. 
2 : Switch to train phase when training; switch to test phase when training

### DIFF
--- a/tiny_cnn/network.h
+++ b/tiny_cnn/network.h
@@ -171,7 +171,9 @@ public:
             if (optimizer_.requires_hessian())
                 calc_hessian(in);
             for (size_t i = 0; i < in.size(); i+=batch_size) {
-                train_once(&in[i], &t[i], std::min(batch_size, in.size() - i), n_threads);
+                train_once(&in[i], &t[i],
+                           static_cast<int>(std::min(batch_size, in.size() - i)),
+                           n_threads);
                 on_batch_enumerate();
 
                 if (i % 100 == 0 && layers_.is_exploded()) {
--- a/tiny_cnn/network.h
+++ b/tiny_cnn/network.h
@@ -35,6 +35,7 @@
 
 #include "tiny_cnn/util/util.h"
 #include "tiny_cnn/layers/layers.h"
+#include "tiny_cnn/layers/dropout_layer.h"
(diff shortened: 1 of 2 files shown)


patrikhuber:
I think you two are talking about different things here. @pansk was saying the conditional (or ternary) operator was hard to read compared to a traditional `if`-`else` statement. I would be very surprised if they didn't get compiled to exactly the same assembly. @stereomatchingkiss on the other hand you explained compile-time branching, which is entirely different.

pansk:
Indeed, I completely agree on the convenience of compile-time branching, even if I'd prefer to do it using plain "if"s and "constexpr"s, rather than with SFINAE.
Unluckily "consterpr"s are not supported in VS2013.

stereomatchingkiss:
Sorry, these comments are refer to old issues, the file I want to commit is "network.h" but not "util.h". My mistake, please omit the changed related to "util.h".

>even if I'd prefer to do it using plain "if"s and "constexpr"s, rather than with SFINAE.

Nice idea, I did not think about it. 
Besides, I think it is called "tag dispatch" but not SFINAE?

>Unluckily "consterpr"s are not supported in VS2013.

After cuda support vc2015, maybe we could consider give up support on vs2013?

pansk:
You can use both std::enable_if (which uses SFINAE) or tag_dispatch to pick the right function at compile time.

I tend to find std::enable_if syntax(es) more clear (despite the fact there're at least three, for different situations), but that's probably just me :)

nyanp:
@stereomatchingkiss 
Thank you for your PRs, and I would appreciate if you open a new pull request about util.h.
```std::enable_if``` and tag-dispatch are both good idea, and I'd be happy with either :)

pansk:
What about moving the set_context up and implementing it as a no-op for all the other layers?
This way the network doesn't need to know there's a special layer (and every other special layer can use the same mechanism).
Moreover, I think the code has a bug: it iterates all the layers, but always accesses layers_[0].

nyanp:
@pansk 
Thanks for your idea and pointing it out :) It would be better to move ```set_context``` up to ```layer_base``` class.

PR #99, opened 2016-02-21
added comments

andreh7:
Added comments in some places in the code according to my understanding what I think the code does, potentially useful for others. No code changes.


nyanp:
@andreh7 
Thank you very much! It's helpful for everyone :)

PR #100, opened 2016-02-22
fix ambuigity of for_ function

stereomatchingkiss:
Fix ambuigity of for_function(#98)


nyanp:
Thank you very much ;)

PR #101, opened 2016-02-23
Fixed table formatting / type in example

Teemperor:
The table is not valid github markdown and ends up being weird text. Making three dashes instead of two should fix this.

Also removed the semicolon in the example.


nyanp:
@Teemperor 
Oh, thanks :)

PR #102, opened 2016-02-29
Parametrizing VECTORIZE_TYPE

pansk:
I parametrized the macro VECTORIZE_TYPE (when inspecting possible actions for cuDNN implementation, I found out that I had to use T as generic type for my template).

One drawback of this approach is that it cannot be used if T is a type whose representation includes a comma (such as a templated type with two template parameters).

A different (more C++11ish) approach would be to use templated usings, such as

template<typename T> using vectorize_type = detail::avx<T>;

and use vectorize_type<T> in place of VECTORIZE_TYPE (or VECTORIZE_TYPE(T) in my PR).

I remember a year ago a PR was partially rejected when replacing typedefs with usings, but the reason was compatibility with VS2012, which we dropped.
Let me know if I should produce a different version with using(s).

### DIFF
--- a/tiny_cnn/util/product.h
+++ b/tiny_cnn/util/product.h
@@ -281,38 +281,38 @@ inline void reduce_aligned(const typename T::value_type* src, unsigned int size,
 } // namespace detail
 
 #if defined(CNN_USE_AVX)
-#define VECTORIZE_TYPE detail::avx<T>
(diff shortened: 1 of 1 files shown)


nyanp:
@pansk
Thanks! Now we can use alias template (and I think it's more C++ish too), but macro solution also works well. I'll merge current version :)

PR #103, opened 2016-02-29
Fix build problem due to missing header for size_t

Teemperor:
I get this error when compiling with clang 3.7.1:

    ./tiny_cnn/config.h:76:9: error: 'size_t' does not name a type
     typedef size_t cnn_size_t;

Including the header for size_t should fix this.

### DIFF
--- a/tiny_cnn/config.h
+++ b/tiny_cnn/config.h
@@ -25,6 +25,7 @@
     SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 #pragma once
+#include <cstddef>
 
 /**
  * define if you want to use intel TBB library


nyanp:
@Teemperor
Nice PR, thank you very much!

patrikhuber:
Also I'm pretty sure it should be `std:.size_t`!

Teemperor:
@patrikhuber +1, but `grep -R "[^_:]size_t" | wc -l -> 202` tells me that problem should be addressed in another dedicated (and bigger) patch

patrikhuber:
@Teemperor Right :-)

Teemperor:
The CI-OS X is failing due to a homebrew error. Should be resolved if [this](https://github.com/Homebrew/homebrew/pull/49667) is merged.

PR #105, opened 2016-03-01
Made convolutional_layer::out_size/out_length static

Teemperor:
clangs UBSAN report:

    tiny_cnn/tiny-cnn/tiny_cnn/layers/convolutional_layer.h:101:60: runtime error: member call on address 0x7fffffffcfe0 which does not point to an object of type 'convolutional_layer'
    0x7fffffffcfe0: note: object has invalid vptr
     00 00 00 00  f0 cf ff ff ff 7f 00 00  00 00 00 00 00 00 00 00  00 d1 ff ff ff 7f 00 00  58 5e f9 f7
                  ^~~~~~~~~~~~~~~~~~~~~~~
                  invalid vptr

Not calling a member function before the base class constructor is called should fix this. I made the functions static (which doesn't seem to break the API as far as I can see) for this reason.

### DIFF
--- a/tiny_cnn/layers/convolutional_layer.h
+++ b/tiny_cnn/layers/convolutional_layer.h
@@ -509,11 +509,11 @@ private:
         return pad_type == padding::same ? (in_length + window_size - 1) : in_length;
     }
 
-    cnn_size_t out_length(cnn_size_t in_length, cnn_size_t window_size, cnn_size_t stride, padding pad_type) const {
+    static cnn_size_t out_length(cnn_size_t in_length, cnn_size_t window_size, cnn_size_t stride, padding pad_type) {
         return pad_type == padding::same ? (cnn_size_t)ceil((double)in_length / stride) : (cnn_size_t)ceil((double)(in_length - window_size + 1) / stride);
     }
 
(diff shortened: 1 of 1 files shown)


nyanp:
@Teemperor
Thanks :) This change will work great, but I have a question: Should we avoid calling non-virtual functions in constuctor? As far as I know, non-virtual functions can be called without using vptr and I wonder why UBSAN report like this...

Teemperor:
I think the problem is that g++/ubsan tries to verify by the vptr-content whether the pointer actually points to an class of the expected type on every member function call. I've played around with this issue and it seems that g++/ubsan on GCC 5.3 is [malfunctioning in this scenario](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70035). As clang++/ubsan still works and it seems to be a GCC bug, I'll close this PR.

nyanp:
@Teemperor
Ok :)

Teemperor:
Well, it seems it's actually undefined behaviour to call a member before base-class constructor (see comment 3 in the GCC bugreport), so I'll reopen the PR.

pansk:
I agree, it is UB.
+1 for making those members static

nyanp:
Thanks for information, I've understand it is definitely undefined behavior.

PR #106, opened 2016-03-01
Fixed unsigned underflow in lrn_layer.h

Teemperor:
Should fix the failing tests. There are probably nicer ways to do this, but I guess just throwing big integers at the problem is ok for now :).


nyanp:
@Teemperor
Oh, it was careless miss... thanks! ;)

PR #107, opened 2016-03-01
Fixed anchor links in readme

Teemperor:
Githubs Markdown processor generates lowercase anchors which breaks navigation in browsers that care for case-sensitive anchor ids (such as Chrome). This should fix this :)


nyanp:
@Teemperor
Good job :) Thanks!

Issue #85, opened 2016-02-01
Consider adding picotest as git submodule


Abai:
This would easily propagate changes and fixes to https://github.com/nyanp/picotest
:)

pansk:
+1, I never worked with git submodules, and I'd love to see them in action :)

PR #111, opened 2016-03-03
PR for #85

pansk:
picotest as a submodule


nyanp:
@pansk 
Nice, thanks :)

nyanp:
@pansk 
CI builds break because `git submodule update` fails:

fatal: reference is not a tree: ab703fb03ddeb494a6468b389fe5172893900ca2
Unable to checkout 'ab703fb03ddeb494a6468b389fe5172893900ca2' in submodule path 'test/

Do you have a solution?

pansk:
Checking, seems like it's referencing a non-existent commit (which probably was related to the PR on picotest - but apparently I forgot to send that PR...)

Issue #115, opened 2016-03-25
Problem with topology of convolution network


Denzelmon:
I need to generate network for face recognition.
I use this topology

net << convolutional_layer<tan_h>(240, 240, 5, 1, 16)
<< average_pooling_layer<tan_h>(236, 236, 16, 2)
<< convolutional_layer<tan_h>(118, 118, 3, 16, 40)
<< average_pooling_layer<tan_h>(116, 116, 40, 2)
<< convolutional_layer<tan_h>(58, 58, 3, 40, 60)
<< average_pooling_layer<tan_h>(56, 56, 60, 2)
<< convolutional_layer<tan_h>(28, 28, 5, 60, 100)
<< average_pooling_layer<tan_h>(24, 24, 100, 2)
<< convolutional_layer <tan_h> (12, 12, 3, 100, 150)
<< average_pooling_layer <tan_h> (10, 10, 150, 2)
<< convolutional_layer <tan_h> (5, 5, 5, 150, 400)
<< fully_connected_layer <tan_h> (400, 12);

But program doesn't work. What problem is?

pansk:
Can you elaborate the part "program doesn't work"? :)
It's really hard to understand otherwise.
Does it hang?
Does it throws?
Is something appearing on the console?
Have you tried to break down the process in single "net << something(...)" operations, and execute them step-by-step in a debugger? (or interleave them with cout << "something" << endl; to see when a line is failing)

Denzelmon:
When I try to define a variable "net" i get an error message "Microsoft Visual C++ Runtime Library Expression: vector subscript out of range"

pansk:
Replace the line with

net << convolutional_layer(240, 240, 5, 1, 16);
net << average_pooling_layer(236, 236, 16, 2);
net << convolutional_layer(118, 118, 3, 16, 40);
net << average_pooling_layer(116, 116, 40, 2);
net << convolutional_layer(58, 58, 3, 40, 60);
net << average_pooling_layer(56, 56, 60, 2);
net << convolutional_layer(28, 28, 5, 60, 100);
net << average_pooling_layer(24, 24, 100, 2);
net << convolutional_layer (12, 12, 3, 100, 150);
net << average_pooling_layer (10, 10, 150, 2);
net << convolutional_layer (5, 5, 5, 150, 400);
net << fully_connected_layer (400, 12);

and then run them step by step, trying to see which line fails.
[...]

PR #116, opened 2016-03-25
Move semantics for layers

pansk:
- Enabling move semantics on layer_base (GCC or VS14+)
- Fixing a bug on operator << for layers (which prevented move semantics to kick in)

(problem originated from #115)


pansk:
This partially solves #115.
To solve it completely we should eventually review the value for CNN_TASK_SIZE when CNN_USE_OMP is defined

nyanp:
@pansk 
Thanks! :+1:

PR #123, opened 2016-04-07
Improve load performace

TohnoSakuya:
Hi,
When I use this framework in my program, the time of a loading a 8-layers-weights-file(1400258 params) runs up to 27 second.

So I add fast_load function. Same task, the time goes down to 0.8 second. 
I think this maybe helpful for someone who cares about running time.

Just a little code, easy to merge

### DIFF
--- a/tiny_cnn/layers/layer.h
+++ b/tiny_cnn/layers/layer.h
@@ -161,6 +161,11 @@ public:
         for (auto& w : W_) is >> w;
         for (auto& b : b_) is >> b;
     }
+    
+    virtual void load(std::vector<double> weights, int& idx) {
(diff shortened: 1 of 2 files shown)


pansk:
The CI is failing, but I think it's quite an easy fix, you just need to add a namespace qualifier in line 294

std::vector<double> data;

nyanp:
@TohnoSakuya 
Thank you very much, it'a great improvement :+1: 
Could you fix the problem with std:: qualifier?

TohnoSakuya:
I am really sorry for the late reply.
I have fix the problem.

nyanp:
@TohnoSakuya 
Thank you very much! :)

PR #127, opened 2016-04-16
Problem: the xor function test fails in my environment

reunanen:
Solution: increase the number of training samples
  (alternatively, increasing the number of epochs appeared to work as well)


nyanp:
@reunanen 
Ok, Thanks! :+1:

PR #130, opened 2016-04-17
Need: Ability to weigh individual targets (e.g., samples of a rare class)

reunanen:
Solution: Add capability to pass a target costs matrix, whose shape equals that of the t matrix


nyanp:
@reunanen 
It's useful for unbalanced training data, great! thank you :)

reunanen:
@nyanp : Thank *you*, for this very nice library! Really looking forward to be able to use it in practice.

reunanen:
Note that this can also be used to implement _emphasized denoising autoencoders_ (see [Vincent et al., 2010](http://www.jmlr.org/papers/volume11/vincent10a/vincent10a.pdf), Section 3.4, pages 3381-3382).

PR #135, opened 2016-04-23
Fix issue 132: size the worker-specific vectors correctly.

reunanen:
Fix issue 132: size the worker-specific vectors correctly.

### DIFF
--- a/tiny_cnn/layers/convolutional_layer.h
+++ b/tiny_cnn/layers/convolutional_layer.h
@@ -320,7 +320,7 @@ public:
     {
         copy_and_pad_input(in_raw, static_cast<int>(worker_index));
 
-        worker_specific_storage& ws = get_worker_storage(worker_index);
+        auto& ws = this->get_worker_storage(worker_index);
         vec_t &a = ws.a_; // w*x
         vec_t &out = ws.output_; // output
         const vec_t &in = *(conv_layer_worker_storage_[worker_index].prev_out_padded_); // input
@@ -376,7 +376,7 @@ public:
     }
 
     const vec_t& back_propagation(const vec_t& curr_delta, size_t index) override {
-        worker_specific_storage& ws = get_worker_storage(index);
+        auto& ws = this->get_worker_storage(index);
         conv_layer_worker_specific_storage& cws = conv_layer_worker_storage_[index];
 
         const vec_t& prev_out = *(cws.prev_out_padded_);
--- a/tiny_cnn/layers/convolutional_layer.h
+++ b/tiny_cnn/layers/convolutional_layer.h
@@ -107,14 +107,14 @@ public:
         pad_type_(pad_type),
         w_stride_(w_stride), h_stride_(h_stride)
     {
-        init();
+        set_worker_count(CNN_TASK_SIZE); // calls init()
     }
 
     /**
     * constructing convolutional layer
     *
-    * @param in_width         [in] input image width
-    * @param in_height        [in] input image height
+    * @param in_width      [in] input image width
+    * @param in_height     [in] input image height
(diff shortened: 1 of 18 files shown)


tiny_cnn/layers/layer.h
```diff
@@ -238,55 +240,79 @@ class layer_base {
...
+    }
+
 protected:
     cnn_size_t in_size_;
     cnn_size_t out_size_;
     bool parallelize_;
 
     layer_base* next_;
     layer_base* prev_;
-    vec_t a_[CNN_TASK_SIZE];          // w * x
```
reunanen: This is the beef of the pull request. These should not be fixed-size, unless n_threads given to train() is somehow limited. (Currently it isn't, so we may overflow these arrays – in particular when not using OpenMP, in which case `CNN_TASK_SIZE` is only 8.)

tiny_cnn/layers/convolutional_layer.h
```diff
@@ -489,15 +493,21 @@ class convolutional_layer : public layer<Activation> {
...
+        conv_layer_worker_storage_.resize(worker_count);
+        init();
+    }
+
 private:
     void init() {
-        for (cnn_size_t i = 0; i < CNN_TASK_SIZE; i++) {
+        for (conv_layer_worker_specific_storage& cws : conv_layer_worker_storage_) {
             if (pad_type_ == padding::same) {
-                prev_out_buf_[i] = new vec_t(in_padded_.size(), float_t(0));
```
reunanen: I wonder if there was a memory leak here. Looks like the allocated `prev_out_buf_` was never `free`d?

Anyway fixed now by avoiding manual memory management.

tiny_cnn/layers/dropout_layer.h
```diff
@@ -43,45 +43,39 @@ class dropout_layer : public layer<activation::identity> {
           dropout_rate_(dropout_rate),
           scale_(float_t(1) / (float_t(1) - dropout_rate_))
     {
-        mask_ = new bool[in_size_ * CNN_TASK_SIZE];
```
reunanen: Avoiding manual memory management here as well.

pansk:
Thank you for your clarification, I think we got the point: n_threads is supposed to be in the range [1, CNN_TASK_SIZE]. If it's not enforced (as it seems), that's where the bug is (all layer classes work like that, if we decide to make it dynamic, we should change it everywhere).
The allocated prev_out_buf_ could be memory leak, I'll look at it.

reunanen:
@pansk : Yes, well, I tried to change it (to be dynamic) pretty much everywhere, though it's quite possible that I missed something.

pansk:
@reunanen My fault, I saw few changes, but they were comments. Apparently it's complete: :+1: for me :)
I don't know @nyanp 's opinion on this, but from my point of view this could be a good improvement (and allow the user to have a better control over allocated memory). I'd just differentiate (if such a distinction makes sense) memory allocated for inference-only from that used by training, and add a function to it to release all the training-related memory, and to set up an inference threading level of N (with default to one), in order to have a slim single-threaded inference-only object. I can take care of this if the PR goes through.

reunanen:
@pansk : Yep, there were so many changes that I thought it'd be best to highlight (at least to some extent) what is important and what is boilerplate.

nyanp:
@reunanen 
Thanks, I think it's a great improvement! I've wanted to shrink memory footprint after training phase (memory footprint is always important for embedding systems), and your PR is very useful for it :+1: 

Encapsulation weights, diffs as `storage` will probably be a good first step for further improvement such as weights sharing over layers (it is necessary for lstm), and gpu supporting.

PR #144, opened 2016-05-12
fix scheduling issue

edgarriba:
- Modified std::all_of for std::any_of since some layer such convolutional and pooling have several backward and forward connections. Not sure if that's the proper solution to check connectivity during the scheduling.
- Refactor some typedefs such *node_ptr* and *layer_ptr*


nyanp:
@edgarriba 
Oh, I'm sorry but my last commit is conflicted with your refactoring... could you merge it?

edgarriba:
sure!

nyanp:
@edgarriba
I've confirmed that your modification is correct! Good job :+1:

PR #160, opened 2016-05-31
update caffe proto

edgarriba:


nyanp:
@edgarriba
Nice, thx :)

PR #163, opened 2016-06-01
fixes allocation issue

edgarriba:
fixes https://github.com/nyanp/tiny-cnn/pull/156#issuecomment-222526548


edgarriba:
yes! we're good to go

PR #166, opened 2016-06-04
Abstract Convolutional Layer

edgarriba:
First commit for device abstraction:

- The convolution class has been abstracted
- Added a PlantUML in order to show the "big picture" of the current implementation.
- Removed test warnings


tiny_cnn/layers/convolutional_layer.h
```diff
@@ -335,6 +355,71 @@ class convolutional_layer : public base_conv_layer<Activation> {
...
+            static_cast<size_t>(this->weight_.width_),
+            static_cast<size_t>(this->weight_.height_)
+        };
+
+        /*const nnp_padding padding = {.top = static_cast<size_t>(pad.cpu_data()[0]),
+                                     .right = static_cast<size_t>(pad.cpu_data()[1]),
+                                     .bottom = static_cast<size_t>(pad.cpu_data()[0]),
+                                     .left = static_cast<size_t>(pad.cpu_data()[1])};*/
+        const nnp_padding padding = { 0, 0, 0, 0 };
+
```
edgarriba: @nyanp Could you help with that?

edgarriba:
@nyanp Do you think that we need to enable NNPACK in Travis?

nyanp:
@edgarriba /cc:@bhack
Good jobs! :) I'll merge it after the merging computational-graph branch into base. I'll also check out the nnpack's functionalities.

- Could you tell me what the "session" in your UML means? (the same meaning to TensorFlow's session?)
- Class hierarchy of convolutional layers is ok. It would be very good if we can extract the core math operations of these classes as free functions from a unit-testing point of view.

edgarriba:
@nyanp 

> - Could you tell me what the "session" in your UML means? (the same meaning to TensorFlow's session?)

Yes. The idea is to follow @naibaf7 thoughts and have a tracking for all the available devices in order to be able to pass data between them. We need to iterate this, that's only a first version of the design.

> - Class hierarchy of convolutional layers is ok. It would be very good if we can extract the core math operations of these classes as free functions from a unit-testing point of view.

Here we need to collaborate in order to find a clean solution. This is only a proof of concept to check that at least we are able to abstract engines. The reason why I chose this design is because by having different classes we need to pass a lot of parameters each time (maybe we can unify this in a struct). Any suggestion for a clean solution?

nyanp:
@edgarriba
Here is another idea: Instead of class hierarchy of conv, extract backend as independent class:

```cpp
// just a sketch
class math_backend {
public:
    // context holds solution-dependent parameters
    // context should be able to hold any types of structures (like boost::any)
    math_backend(context *ctx);

    // core math functions

    virtual void conv2d(...) {
        // provides default implementation (extracted from current tiny-cnn's implementation)
    }

    virtual void matmul(...);
    virtual void maxpool(...);

protected:
    context* ctx_;
};

class nnpack_math_backend : public math_backend {
..
};

////

class convolutional_layer {
public:
    // performance critical layers delegate their core calculations to the backend
    // yes, it is a strategy design pattern
private:
    math_backend *kernel_;
};
```

Layers are responsible to calculate forward/backward pass correctly. Backends are responsible to calculate primitive math-functions efficiently (and probably manage device-dependent parameters). The merit of this architecture is we don't need to create derived classes per layer (nnpack_conv_layer, nnpack_fully_connected_layer, etcetc).

edgarriba:
@nyanp  that looks interesting.

Then how do we specify/initialize different e.g. `convolutional_layer`engines? Is there where we should keep tracked different devices as @naibaf7 pointed out?

nyanp:
@edgarriba 
I agree that we should track the list of layers, to specify which device owns them. Maybe ```Device``` class should hold the list of ```Backend``` (instead of ```Layer``` in your model):

device-abstraction-uml

(I don't think it is the best idea, and we need some iteration to find a deep model)

There are some ways to specifying/initializing backend for each layer:
- Preprocessor macro (like ```CNN_USE_NNPACK```)
    - Bind a backend at the preprocess phase
    - Simple, but not clean and not flexible (Maybe it is not suitable for opencv's backend. We don't want to recompile opencv every time we switch backends)
- Additional constructor's parameter
    - ```convolutional_layer::convolutional_layer(..., backend_type backend = backend_type::default)```
    - Simple, but it is difficult to feed additional customizable parameters into backends
- Dependency Injection
    - ```convolutional_layer::convolutional_layer(..., backend* backend = backend::default())```
- Select the fastest backend automatically
    - I'm not sure if we can

Did you learn how tensorflow manage this problems?

@bhack Do you have any option regarding @edgarriba 's proposal?

bhack:
@nyanp I've commented some critical section here https://github.com/BVLC/caffe/issues/4187#issuecomment-223598905.
In this first iteration all the devices are CPU so there could be not so much device transfer overhead.
The main problem I see is that parameters support is not homogeneous between backends. What we want to do if tinycnn doesn't support padding on pooling and nnpack support it or viceversa?

mtamburrano:
@nyanp I think we could go for a mixed version of your 2nd and 4th proposals.
I'd index all backends with a ranking ID to let the system to automatically choose the best one, but the user could specify a "preferred backend" as a constructor's parameter.
The constructor expects all the possible parameters with a default invalid value (e.g. `convolutional_layer::convolutional_layer(..., stride = -1, pad = -1, ..., backend = backend::nnpack())`).
If all the specified parameters are supported from the preferred backend, than that backend is used and it gets _chosen_, otherwise the system automatically uses as fallback a backend (with higher possible rank) that supports the given parameters. If the given parameters are not supported all at once from any backend, an exception is raised. An exception is also raised if an user specifies an invalid value for a parameter, of course.
The not specified parameters are switched to a default (e.g. `stride = 1`) or ignored if the chosen backend doesn't support them.

Why the default is an invalid value? To avoid cases where we have two backends, let's say B1 and B2, and 2 parameters, P1 and P2. B1 supports P1 but not P2, and B2 supports P2 but not P1. If a users specifies both valid values for P1 e P2, we can easily detect that we have no available backends that support both of them to raise an exception. If the default value is a valid one instead, an user could specify the same default and we couldn't detect if the parameter was passed or it simply has the default value.

I'd also add a BackendParameters in the Backend constructor, to add specific backend parameters, so to instantiate a layer with a preferred backend one can write: `convolutional_layer::convolutional_layer(..., backend_type backend = backend_type::nnvpack backend_params = nnvpack_params(p1, p2))`. In this way we could avoid to flood the layer constructor with all possible parameters of all backends.
Another option would be to set backend's parameters as a global session variable but this is useful if we want that all layers for a specific phase use those parameters , but could be an issue to differentiate single layers.

I think this could work easily while we use CPU only, when mixing CPU and GPU backends, automatically choose between them could not be so easy. So we need some additional plan for that case

edgarriba:
@mtamburrano I like your proposal. I think that can be useful to specify for each layer the backend to use. However, the drawback of this ranking system will be the spaghetti set of IF rules.

naibaf7:
> The not specified parameters are switched to a default (e.g. stride = 1) or ignored if the chosen backend doesn't support them.

I would disagree with this part of the proposal. Ignoring a parameter will change model properties. Here, the option is to fallback to a backend that does support it (if available, eventually with warning to the user), or a fatal error and stop the network.

nyanp:
@bhack 
Tensorflow's whitepaper is very interstring resource, I would recommend @edgarriba to read it :) The overhead of copying tensors between devices might be critical in the next iteration, but I feel that current idea is good enough as the starting point. I agree that "inpedance mismatch" (or 
homogeneousness) between backends is the problem.

@mtamburrano 
Thanks for the idea! Mixed approach is interesting :) The ranking system seems overkill for me, I think the following rules covers the most of use cases:
- If user explicitly specify the backend, each layer try to use it
- If user don't specify, 
    - Use gpu backend if the system detect any gpu
    - Use cpu backend otherwise
- In all cases, the system will throw an exception if the selected device doesn't support parameter

Implicit fallbacking to other backend would cause unexpected problem, so I like this simpler rule. What do you think?

mtamburrano:
@nyanp 
my suggestions were about different CPU backends, like with or without NNPACK.
Speaking about GPU, probably a simpler approach as you proposed could be the best one, at least for this gsoc, because probably it's not easy to mix GPU layers with CPU ones, or if it is easy would be probably very slow most of the time without an accurate design.
But if the user wants simply use CPU, why not allow him to use the best choice for a specific layer?
Maybe he wants to use a pooling layer that is only supported on NNPACK, but he also wants to use stride on conv layers and NNPACK doesn't support stride.
We could also drop the ranking system and let the user to manually specify each backend for each layer and raise an exception each time he specifies a parameter non supported without fallback to a different backend, but I don't think that automatic fallbacking is a big issue while staying on CPU.
But maybe I'm wrong and it is a big mess, I never tested it, this is just gut feeling

bhack:
@naibaf7 Have you any feedback for this first stage? Next target will be libdnn (do you plan to support other datatypes?)

naibaf7:
@bhack 
FP16 is no big effort, can be compiled with FP16 support soon and requires only minimal autotuning effort. As with 64bit, atomic kernels may not be available on 16bit depending on the device capabilities. 8bit probably not in the immediate future.

When do you guys need a standalone libDNN library? If it is urgent I have to reschedule my development plans a bit, as it's currently part of Caffe.
Minimal dependencies to compile libDNN will be C++11 and (ViennaCL or NVRTC) depending on OpenCL or CUDA release.

Wangyida:
@bhack I a not so familiar with other backend data with 8FP and 16FP, but I find that even ristrreto network modified from Caffe is still a concept of 8PF in precision and the data structure is still 32 bits. I think the precision adjustment strategy in ristrreto could be applied directly in tiny-cnn which is mainly embedded in the conv layer, deconv layer and fully connected layer.

bhack:
@naibaf7 What is your original timeplan? We are going on integrating a CPU only backend like nnpack in the first iteration. Your has a little bit more impact at infrastructural level cause can consume CPU and GPU devices.

naibaf7:
@bhack 
I wanted to spend a bit more time on optimizing the kernels for AMDs GCN architecture until end of June and provide the standalone library in mid/end-July together with a system that allows users to exchange autotunings (since it can take a bit of time).

bhack:
GSoC midterm end 27/6. Probably we need to approach to libdnn before midterm. Performance are important but we need to do that at infrastructural level.

naibaf7:
@bhack 
OK I will see if I find time to make it standalone until then.
But if not, the interface will stay about the same as it is now, so you can at least write the interface already.

nyanp:
@mtamburrano 
> But if the user wants simply use CPU, why not allow him to use the best choice for a specific layer?
Maybe he wants to use a pooling layer that is only supported on NNPACK, but he also wants to use stride on conv layers and NNPACK doesn't support stride.

Definitely. Implicit switching between GPU and CPU would be difficult, but switching between CPU-based implementation may be user-friendly :) Perhaps selecting the "best" one from many cpu-engines still isn't easy (what does "the best" mean? minimum clock cycle, lowest memory bandwidth or minimum memory usage?), but if we have only 2 engines per platform, we can avoid most of the problem.

bhack:
Ehm @naibaf7's engine can consume GPU and CPU devices. So in a strict sense it is also a CPU engine. Best allocation it is an open topic cause could involve also costrains (memory, locality etc...) partially discussed in the updated [Tensorflow paper](http://arxiv.org/abs/1605.08695).

naibaf7:
@bhack 
Wow not so fast, libDNN is CUDA + OpenCL, not CPU :) if CPU, then over OpenCL, but the way the shared memory GEMM works is only suitable for high performance on GPU architectures like VLIW4/5, GCN and Fermi, Kepler, Maxwell, Pascal and maybe Intel GPUs with different tunings.

It's not very suitable for CPUs and mobile GPUs without shared memory, for which a different algorithm is required for peak performance (that may come later).

bhack:
Yes for mobile CPU we will looking to https://github.com/google/gemmlowp. Probably in some embedded configuration like APU your engine could work fine with some adaptation.

naibaf7:
@bhack 
Yes for AMD APUs definitely, but I haven't gotten my hands on one yet for tuning or testing. Same with modern Intel APUs.

bhack:
Yes I'm really interested to see how things will run on [hUMA](http://www.amd.com/en-us/innovations/software-technologies/compute-cores)

edgarriba:
hey, check last commit just to see if we are in the good path. Still need some cleaning

nyanp:
@edgarriba
It seems good :) Let's keep on going :+1:

tiny_cnn/core/tiny_backend.h
```diff
@@ -0,0 +1,150 @@
...
+              float_t * pa  = &a[params_.out.get_index(0, 0, o)];
+              float_t * paa = pa + params_.out.width_ * params_.out.height_;
+              std::for_each(pa, paa, [&](float_t& f) { f += bias[o]; });
+          }
+      });
+
+      /*for_i(this->parallelize_, params.out.size(), [&](int i) {
+          out[i] = this->h_.f(a, i);
+      });*/
+  }
```
edgarriba: @nyanp not sure if it should be done here or outside the kernel. Besides, check how do I use conv params. Also not sure if they should be in kernel signature.

tiny_cnn/layers/convolutional_layer.h
```diff
@@ -415,108 +343,125 @@ class convolutional_layer : public feedforward_layer<Activation> {
...
-            else {
+        // TODO: got a seffault at somepoint here
+        for (conv_layer_worker_specific_storage& cws :
+                conv_layer_worker_storage_) {
+            if (params_.pad_type == padding::same) {
+                cws.prev_out_buf_.resize(params_.in_padded.size(), float_t(0));
+                cws.prev_delta_padded_.resize(
+                    params_.in_padded.size(), float_t(0));
+            } else {
                 cws.prev_out_buf_.clear();
```
edgarriba: @nyanp something weird is happening here when we allocate workers and we pass them to kernels. A segfault raises in this case.

nyanp: @edgarriba
Let's consider the following code:

```
net << conv<tan_h>(...);
```

- A temporary object of conv-layer (A) is created
- operator<< is called
- a move construction of conv-layer occured in ```sequential```, and another conv-layer (B) is created (every member of B is moved from A)
- A is destructed at the end of the line
- B holds the conv_param shared from A, but the conv_param holds A's member which is already destructed

backend has similar problem because of this line:

```
Backend::backend_->set_layer(this);
```

edgarriba: @nyanp I'm used with move constructors, learning it right now. Does it means that we should add extra move constructors for conv_params and math_backend classes?

nyanp: @edgarriba 
> I'm used with move constructors, learning it right now. 

Good :) If you want to learn more deeply, Effective Modern C++ is a great book for you.

> Does it means that we should add extra move constructors for conv_params and math_backend classes?

Yes we'll be able to solve the segfault by it. I'm sorry but I haven't catch up your whole code yet, I'll check it within a day.

edgarriba: @nyanp I solved it or at least tests pass again. Check commit 82c9412

tiny_cnn/core/nnp_backend.h
```diff
@@ -56,40 +58,40 @@ class nnp_backend : public math_backend {
...
-              this->conv_layer_worker_storage_[index].prev_out_padded_);
+          reinterpret_cast<const float*>(&in_data[0]->at(0));
 
       const float* kernel_pointer =
-          reinterpret_cast<const float*>(&in_data[0]->at(0));
+          reinterpret_cast<const float*>(&in_data[1]->at(0));
 
       const float* bias =
-          reinterpret_cast<const float*>(&in_data[1]->at(0));
+          reinterpret_cast<const float*>(&in_data[2]->at(0));
```
edgarriba: @nyanp Could you check that I pass correctly the parameters to nnpack? The inference is running but I got zeros in the output. 

@nyanp @bhack @mtamburrano 
Besides, I would like to start to implement Device and Session classes. Any suggestion what should be doing each one?

mtamburrano: @edgarriba 
I would keep on hold the implementation of the Session, it could be useful to optimize some parallel runs of a same model like Tensorflow does, but I think it needs a redesign of model declaration to take advantages of sessions.
For the Device class, I'll start to add some discovery methods on derived classes (like to find number of devices for OCLDevice or the number of cores for CPUDevice). You can also implement some general method on the Base class (like, set the device ID). 
I see a Device instance will be passed to a backend, so we could, for example, limit the computation to a specific GPU or to a certain number of cores.
If some backend doesn't support device filtering (I'm not sure NNPACK does it) the device param will be ignored (with an alert)

nyanp: @edgarriba 

- Let's change the following line in config.h to correct type mismatch.
```cpp
//typedef double float_t;
typedef float float_t
```

- ```in_padded``` in tiny-cnn is size of input images after padding, and ```nnp_padding``` in NNPack is size of padding itself (in_padded.width = in.width + padding.left + padding.right). so value of ```padding``` should be corrected.

- Let's change the index of ```output_pointer```. out_data[0] is used for the output data after activation. And I think moving these code to convolutional_layer is better. (These indexes depend on the implementation detail of convolutional layer, so the implementation of the layer and the engine is tightly coupled in current design)

```cpp
      //float* output_pointer = reinterpret_cast<float*>(&out_data[0]->at(0));
      float* output_pointer = reinterpret_cast<float*>(&out_data[1]->at(0));
```

tiny_cnn/core/nnp_backend.h
```diff
@@ -88,7 +93,7 @@ class nnp_backend : public math_backend {
           reinterpret_cast<const float*>(&in_data[2]->at(0));
 
       float* output_pointer =
-          reinterpret_cast<float*>(&out_data[0]->at(0));
+          reinterpret_cast<float*>(&out_data[1]->at(0));
```
edgarriba: @nyanp I followed your suggestions but still getting same results

nyanp: @edgarriba
Hmm..Let's check values of weight vector in your debugger if you still get zero vector after fixing dx/dy.

tiny_cnn/core/nnp_backend.h
```diff
@@ -117,6 +114,10 @@ class nnp_backend : public math_backend {
           threadpool,
           profile);
 
+      if (status != nnp_status_success) {
+          throw nn_error("Could not succeed with nnp_convolution_inference");
+      }
+
       // TODO: embed it into a class
```
edgarriba: @nyanp The result is the same.

@bhack @mtamburrano 
Besides, I added a status checking and I realized that it raises an error each time. Maybe we are not configuring well NNPACK. I'll squash my commits and after that you can merge the PR.

edgarriba:
@nyanp Since we changed float_t to float, now picotest crashes.

nyanp:
@edgarriba 
A data race occurs in tiny_backend: a local variable ```idx``` is write-accessed from multiple thread. It may crashes the program. Unfortunately I have no idea about CMP0015 error...

tiny_cnn/layers/convolutional_layer.h
```diff
@@ -406,7 +310,31 @@ class convolutional_layer : public feedforward_layer<Activation> {
...
+    void init_backend(backend_t backend_type) {
+        using namespace std::placeholders;  // for _1, _2, _3...
+        switch (backend_type) {
+            case backend_t::tiny_cnn:
+                Base::backend_ = std::make_shared<core::tiny_backend>(&params_,
+                    std::bind(&convolutional_layer::copy_and_pad_input, this, _1, _2),
+                    std::bind(&convolutional_layer::copy_and_unpad_delta, this, _1, _2),
+                    std::bind(&Base::backward_activation, this, _1, _2, _3),
+                    &conv_layer_worker_storage_);
+                Base::backend_->set_layer(this);
```
patrikhuber: Can you not use lambas here instead of `bind`? `std::bind` is pretty much discouraged, I believe it's [this talk](https://youtu.be/zt7ThwVfap0?t=25m8s) that is a good reference.

tiny_cnn/core/tiny_backend.h
```diff
@@ -54,13 +54,13 @@ class tiny_backend : public math_backend {
...
 
-      cnn_size_t idx = 0;
       std::fill(a.begin(), a.end(), float_t(0));
 
       for_i(layer_->get_parallelize(), params_->out.depth_, [&](int o) {
           for (cnn_size_t inc = 0; inc < params_->in.depth_; inc++) {
               if (!params_->tbl.is_connected(o, inc)) continue;
 
+              cnn_size_t idx = 0;
               idx = params_->in.depth_ * o + inc;
```
edgarriba: @nyanp Putting the idx var inside the parallel for should fix the data race issue, right?

nyanp: @edgarriba
Yes :)

edgarriba:
@nyanp Any idea why gradient checking brakes the tests?

nyanp:
@edgarriba
- ```delta``` in network::calc_delta is too small for float - it is always rounded to zero. it should be adjusted for actual type of float_t. I think std::numeric_limit<float_t>::epsilon() is useful for this purpose.
- simply error-threshold is too strict for float.

edgarriba:
> simply error-threshold is too strict for float. 

@nyanp  so, which threshold do you suggest?

nyanp:
@edgarriba 
I can't have confidence in that, but 1e-2 or 1e-3 would fit.

Adding following code into testhelper.h:
```cpp
template <typename T>
inline T epsilon() {
    return 0;
}

template <>
inline float epsilon() {
    return 1e-2f;
}

template <>
inline double epsilon() {
    return 1e-4;
}
```
and replacing floating point literal to epsilon() will work.
```cpp
//EXPECT_TRUE(nn.gradient_check<cross_entropy>(&a, &t, 1, 1e-4, GRAD_CHECK_ALL));
EXPECT_TRUE(nn.gradient_check<cross_entropy>(&a, &t, 1, epsilon<float_t>(), GRAD_CHECK_ALL));
```

edgarriba:
@nyanp no luck with your solution. I'll check that this binded functions haven't broken the backward pass.

tiny_cnn/layers/convolutional_layer.h
edgarriba: @patrikhuber I'm not used with lambdas.
After a first search I would go for the following approach:
   
       Base::backend_ = std::make_shared<core::tiny_backend>(
           &params_,
           static_cast<std::function<void(const vec_t&, int)>>(copy_and_pad_input),
           static_cast<std::function<void(const vec_t&, vec_t&)>>(copy_and_unpad_delta),
           static_cast<std::function<void(const vec_t&, const vec_t&, vec_t&)>> (backward_activation),
           &conv_layer_worker_storage_);

where

      auto copy_and_pad_input   = [](const vec_t& in, int worker_index) { }
      auto copy_and_unpad_delta = [](const vec_t& delta, vec_t& dst) { }

nyanp:
@edgarriba
taking the following step is good to isolate the problem:

1. run tests on the older version of con-layer with float - If tests still fail, we should fix the broken tests at first.
2. after fixing tests, replace conv-layer to your version, and run tests again with  CNN_SINGLE_THREAD option (this macro supresses parallelization in for_i) - If tests fail, something happened in your code. A binding is suspicious if test fails only when the test uses temporary object for layer construction.
3. finally, run tests without  CNN_SINGLE_THREAD option - if tests fail in this step, data race happens somewhere.

edgarriba:
@nyanp The problem is float_t as float. I guess we need to find a proper casting type for testing.

bhack:
@nyanp Have you tried to handle FLT_EVAL_METHOD?

edgarriba:
@nyanp @bhack good to go!

nyanp:
@edgarriba 
Excellent! :+1:

PR #167, opened 2016-06-05
An option to be indepedent of OpenCV

beru:
It's hard to deny that OpenCV is a de-facto standard library in computer vision.
But it's rather huge for my PC-9801 computer.


nyanp:
@beru 
Excellent! Attempts to reduce dependencies are always welcome in tiny-cnn :)
Could you extract stb version of test program as the independent file? I want to keep example source code as clean as possible, for both OpenCV&stb users.

nyanp:
@beru 
And I'd very happy if you solve the AppVeyor's fail. (I'll fix it if you don't have enough time)

beru:
@nyanp 
I updated the branch. Would you check it please?

CMakeLists.txt
```diff
@@ -54,7 +56,7 @@ OPTION(BUILD_DOCS "Set to ON to build documentation" OFF)
 #   Find Dependencies
 # ----------------------------------------------------------------------------
 
-IF(BUILD_EXAMPLES)
+IF(USE_OPENCV)
```
nyanp: @beru
In this PR CMake requires opencv even if we don't compile the mnist example (This is why AppVeyor failed to build). We need opencv only when BUILD_EXAMPLS && USE_OPENCV. Could you fix it?

beru:
Thank you for pointing out the problem. I updated CMakeLists.txt file.
Could you please take a look at it again?

nyanp:
@beru 
Perfect :) I'll merge it. Thank you very much!

edgarriba:
@beru nice contribution!
Could you implement a find_package cmake instead of adding the header files?

Take a look how I solved this for NNPACK: https://github.com/edgarriba/tiny-cnn/tree/feat/generic-computational-graph-device-abstraction/cmake/Modules

Extra link: Writing find modules https://cmake.org/Wiki/CMake:How_To_Find_Libraries

edgarriba:
@nyanp This need some fixes for new branch.

beru:
@edgarriba 
Thank you for the kind messages.
I've noticed the problem but I'm slow to react. Sorry for the delay.
I posted another PR #189 recently. I wish that can fix the problem.

Talking about using CMake find_package command feature for stb_image.
Since stb_image is a single-file headers library and it doesn't have a Makefile so I'm not sure if it is designed to be installed into certain path. Probablly stb_image library is not registered to repositories of famous package managers too. So I'd rather the files to be accompanied with tiny-cnn.

PR #171, opened 2016-06-08
Duplicate symbols fix and test

rolandpersson:
I discovered some link errors due to duplicate symbols when I upgraded to 0.1.0. This happens for me because I'm using the tiny cnn headers in more than one translation unit. 

This pull request contains some inline keywords to fix the link errors and an added .cpp file included in the tiny_cnn_test in order to try and catch this type of error when they are introduced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ ENDIF()
 
 
 IF(BUILD_TESTS)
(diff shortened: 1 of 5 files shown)


nyanp:
@rolandpersson 
Oh it's a critical bug. Adding the test is very good idea, thank you veru much :)

PR #173, opened 2016-06-08
split tests and reorganize cmake

edgarriba:


nyanp:
@edgarriba 
Could you tell me the purpose of this PR? If it is related to [picotest#3](https://github.com/nyanp/picotest/issues/3), I feel that replacing picotest to gtest is the better choice to enable us to execute specific test cases (The only merit of picotest is single-header).

edgarriba:
yes we can think in alternatives in order to add more functionalities at developing time.

PR #174, opened 2016-06-09
Implement Caffe related dependency in CMakeList and modification for deconv between tiny-cnn and Caffe

Wangyida:
The cmake style of compiling proto file of caffe is embedded in the main CMakeList.txt, this will generate caffe.proto.h and caffe.proto.cc automatically in make process rather than using protoc command in the the io folder.

Padding strategy between tiny_cnn and Caffe is modified as the same way in deconvolution layer so the caffe_converter will work in tiny_cnn with trained models.

Already tested, able to be merged.


bhack:
Thank you. We have talked about this some days ago with @edgarriba.

edgarriba:
oh, nice! @Wangyida Is PROTOBUF_GENERATE_CPP generating the sources to tiny_cnn/io/caffe? In CMake docs* its said that it can only be called in the same directory where sources exist. Besides, it's out of CMake policy to generate files outside of the building directory. Apart from that it will be really useful, even for testing.

*https://cmake.org/cmake/help/v3.0/module/FindProtobuf.html

Wangyida:
@bhack @edgarriba Hi, Thanks for your tests, PROTOBUF_GENERATE_CPP is a command provided by FindProtobuf.cmake. PROTO_SRCS and PROTO_HDRS are variables in FindProtobuf.cmake. This will put caffe.pb.cc and caffe.pb.h to these two variables. I also tested another way which compile caffe.pb.cc and caffe.pb.h as a standalone lib for example_caffe_converter to link with. Both ways could do the same thing.

Wangyida:
@nyanp During cmake, caffe.pb.* will be generated in the standard path as they are compiled with protoc in io folder. Both building in root and ./build is available for source and header. May this be useful for a clean build.

nyanp:
@Wangyida 
It's a great advance for me. Thanks :+1:

bhack:
@nyanp I suggest you to ask squashing before merge for having more linearity in the master history.

Wangyida:
@bhack Squashing on commits are necessary, I will squash or fixup them next time~

PR #178, opened 2016-06-13
Removed a repeated include.

jamseams:
Just removed a repeated include that has been overlooked.


nyanp:
@jamseams 
Nice, Thanks! :)

PR #181, opened 2016-06-14
Refactor layer to handle minibatch at once

reunanen:
For [Batch Normalization purposes](https://github.com/nyanp/tiny-cnn/issues/147#issuecomment-225482682), refactor layer to handle a minibatch at once.

Not saying that I completely knew what I was doing, so please be critical!

### DIFF
--- a/test/test_network.h
+++ b/test/test_network.h
@@ -156,9 +156,11 @@ TEST(network, multi_out) {
     auto hidden = std::make_shared<layers::fc<tan_h>>(2,4);
     auto out1 = std::make_shared<layers::fc<tan_h>>(4,2);
     auto out2 = std::make_shared<layers::fc<tan_h>>(4,2);
+    auto out3 = std::make_shared<layers::fc<tan_h>>(4,2);
 
     in << hidden << out1;
     hidden << out2;
+    hidden << out3;
 
     for (size_t i = 0; i < tnum; i++) {
         bool in[2] = { bernoulli(0.5), bernoulli(0.5) };
@@ -166,13 +168,14 @@ TEST(network, multi_out) {
         data.push_back({ in[0] * 1.0, in[1] * 1.0 });
 
         out.emplace_back(std::vector<vec_t>{
-            {(in[0]&&in[1])*1.0, (in[0]||in[1])*1.0}, // 1st output train and/or function
-            {(in[0]^in[1])*1.0, (in[0]==in[1])*1.0} // 2nd output train xor/eq function
+            {(in[0]&&in[1])*1.0, (in[0]||in[1])*1.0}, // 1st output: train and/or function
+            {(in[0]^in[1])*1.0, (in[0]==in[1])*1.0},  // 2nd output: train xor/eq function
+            {-1, 0.5}                                 // 3rd output: train constants (purpose: make output channel count different from output dimension)
         });
     }
 
-    // construct single input, dual output network
-    construct_graph(net, {in}, {out1,out2});
+    // construct single input, three output network
+    construct_graph(net, {in}, {out1,out2,out3});
 
     optimizer.alpha *= 10;
 
@@ -184,14 +187,16 @@ TEST(network, multi_out) {
         std::vector<vec_t> actual = net.predict(std::vector<vec_t>{{ in[0] * 1.0, in[1] * 1.0 }});
         vec_t actual_out1 = actual[0];
         vec_t actual_out2 = actual[1];
+        vec_t actual_out3 = actual[2];
 
         EXPECT_NEAR(actual_out1[0], in[0]&&in[1], 0.1);
         EXPECT_NEAR(actual_out1[1], in[0]||in[1], 0.1);
-        EXPECT_NEAR(actual_out2[0], in[0]^in[1], 0.1);
+        EXPECT_NEAR(actual_out2[0], in[0]^in[1],  0.1);
         EXPECT_NEAR(actual_out2[1], in[0]==in[1], 0.1);
+        EXPECT_NEAR(actual_out3[0], -1,  0.05);
+        EXPECT_NEAR(actual_out3[1], 0.5, 0.001);
     }
 }
-
 
 TEST(network, train_predict) {
     // train xor function
--- a/test/test_network.h
+++ b/test/test_network.h
@@ -156,11 +156,9 @@ TEST(network, multi_out) {
     auto hidden = std::make_shared<layers::fc<tan_h>>(2,4);
     auto out1 = std::make_shared<layers::fc<tan_h>>(4,2);
     auto out2 = std::make_shared<layers::fc<tan_h>>(4,2);
-    auto out3 = std::make_shared<layers::fc<tan_h>>(4,2);
 
     in << hidden << out1;
     hidden << out2;
-    hidden << out3;
 
     for (size_t i = 0; i < tnum; i++) {
         bool in[2] = { bernoulli(0.5), bernoulli(0.5) };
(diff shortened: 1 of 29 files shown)


tiny_cnn/network.h
```diff
@@ -672,10 +672,18 @@ class network
...
+            if (start_index < end_index) { // any data for this thread to process?
+                // process data points in this batch assigned to thread i
+                std::vector<tensor_t> in_batch(&in[start_index], &in[end_index]);
+                std::vector<tensor_t> t_batch(&t[start_index], &t[end_index]);
+                std::vector<tensor_t> t_cost_batch
+                    = t_cost
+                    ? std::vector<tensor_t>(&t_cost[start_index], &t_cost[end_index])
+                    : std::vector<tensor_t>();
+
+                bprop<E>(fprop(in_batch, i), t_batch, i, t_cost_batch);
```
reunanen: This `bprop`s only those samples from the minibatch that have been assigned to this thread. I guess this will need to be changed when implementing Batch Normalization, so that the whole minibatch goes to the layer as a single object. The parallelism can then be handled inside each layer. This may be a good thing, because the optimal parallelism strategy may be different for each layer type (at least in theory).

tiny_cnn/network.h
```diff
@@ -732,17 +745,15 @@ class network
...
-            });
-        }
+    void bprop(const std::vector<vec_t>& out, const std::vector<vec_t>& t, int idx, const std::vector<vec_t>& t_cost) {
+        bprop<E>(std::vector<tensor_t>{out}, std::vector<tensor_t>{t}, idx, std::vector<tensor_t>{t_cost});
+    }
 
+    template <typename E>
+    void bprop(const std::vector<tensor_t>& out, const std::vector<tensor_t>& t, int idx, const std::vector<tensor_t>& t_cost) {
+        tensor_t delta = gradient<E>(out, t, t_cost);
         net_.backward(delta, idx);
```
reunanen: This too may require some further changes when implementing Batch Normalization.

edgarriba:
@reunanen it looks nice! be aware that soon there will be some changes in the layers API.
Check this PR for device abstraction: https://github.com/nyanp/tiny-cnn/pull/166

test/test_network.h
```diff
@@ -156,26 +156,23 @@ TEST(network, multi_out) {
...
         bool in[2] = { bernoulli(0.5), bernoulli(0.5) };
         label_t expected = (in[0] ^ in[1]) ? 1 : 0;
         data.push_back({ in[0] * 1.0, in[1] * 1.0 });
 
         out.emplace_back(std::vector<vec_t>{
-            {(in[0]&&in[1])*1.0, (in[0]||in[1])*1.0}, // 1st output: train and/or function
-            {(in[0]^in[1])*1.0, (in[0]==in[1])*1.0},  // 2nd output: train xor/eq function
-            {-1, 0.5}                                 // 3rd output: train constants (purpose: make output channel count different from output dimension)
+            {(in[0]&&in[1])*1.0, (in[0]||in[1])*1.0}, // 1st output train and/or function
+            {(in[0]^in[1])*1.0, (in[0]==in[1])*1.0} // 2nd output train xor/eq function
```
edgarriba: @reunanen please cast the results to `float_t` as done [here](https://github.com/edgarriba/tiny-cnn/blob/8756f3f8a9a98420ae5b69c96d7394866f6c3d6e/test/test_network.h#L166) in order to have a consistent code

reunanen: Ok. But actually I reverted this change already. Didn't really mean to push it in the first place – just added it to more easily track how the different dimensions are supposed to work.

edgarriba:
@reunanen @nyanp this PR is to master. We should define a quick protocol for development branches

reunanen:
Damn, I only tested with OpenMP enabled. Apparently the tests fail without it. Will try to fix.

bhack:
@nyanp Yes we need to find a policy for branches and pull request in this period of heavy refactoring

edgarriba:
my PR can wait after we can merge this minibatch approach. Otherwise, it will be a pain to adapt it.

nyanp:
@reunanen 
Great step! Did you execute a performance profiling? I'm interested in how this PR will affect the speed of forward/backward pass.

@edgarriba 
Oops I've accidentally merged your first PR. As far as I see it is not so hard to merge this PR into current feat/generic-computational-graph, so I'll do it.

edgarriba:
no worries

@nyanp @reunanen @bhack 
Do you think that with the new API we could implement this minibatches routines inside `layer::forward/backward_pass()` ?

nyanp:
@edgarriba 
Yes, and we'll have more opportunities for improve the training time with minibatch parallelism inside each layers.

reunanen:
Unfortunately still struggling with correctness. @nyanp, good question (re: perf), compared it for the first time now. Seems to be poorer at the moment (about 10% or so); will try to look into that as well.

tiny_cnn/layers/dropout_layer.h
```diff
@@ -134,22 +142,26 @@ class dropout_layer : public layer {
...
     }
 
     virtual void set_worker_count(cnn_size_t worker_count) override {
         Base::set_worker_count(worker_count);
         dropout_layer_worker_storage_.resize(worker_count);
 
         for (dropout_layer_worker_specific_storage& dws : dropout_layer_worker_storage_) {
-            dws.mask_.resize(in_size_);
+            if (dws.mask_.empty()) {
+                dws.mask_.resize(1, std::vector<uint8_t>(in_size_));
```
nyanp: Size of ```mask_``` allocated here is [1, in_size_], but what we need is [sample_size, in_size_]. Possible solution is one of them: (I think the former is better)

- Add new virtual function ```set_minibatch_count``` into layer classes, and call it from ```network``` at the beginning of training
- Check the size of minibatch in fprop and resize ```mask_``` on demand - It requires a exclusive locking while resizing

nyanp:
@reunanen 
> Seems to be poorer at the moment (about 10% or so); will try to look into that as well.

Ok. I think loss around 10% is acceptable for master branch :) but it's great if you improve it better!

> So even if the tests pass, I don't think everything is correct. Probably better not merge just yet.

Current tests do almost nothing about minibatch, so it would be great if you add/extend a minibatch version of ```gradient_check``` method.

edgarriba:
besides, if we parallelize the for loops for mini-batches I think that it will be a huge boosting

bhack:
For convolution there are differents batched gemm approaches. Make a google search on this topic. /cc @naibaf7

tiny_cnn/layers/dropout_layer.h
reunanen: @nyanp: Correct. For other layer types I took the latter approach, though (see e.g. [here](https://github.com/reunanen/tiny-cnn/blob/80272e443b10c3f8ed53d6d28da054d38aee00b8/tiny_cnn/layers/convolutional_layer.h#L494-L495) – although I think this may be one of the places that are not correct yet, but let's not focus on that right now).

Not sure why it requires a lock, though, as each vector (to be resized) is still worker-specific.

nyanp: @reunanen
Ah I misunderstood, you are correct. it doesn't require a lock while resizing.

reunanen: Fixed by resizing in fprop.

reunanen:
Regarding the discussion on perf – after all the point is to be able to do Batch Normalization, which should let us learn more in a single epoch. But maybe this shouldn't be merged to master, before BN actually exists?

nyanp:
@reunanen 
As far as I know, all other major dnn libraries(tensorflow, caffe, mxnet, keras) handle minibatch as this PR - it is a good sign for us. Since we may be heading in the right direction regardless of BN, I can merge this PR without BN implementation.

Could you add a gradient-check test for mini-batch?

reunanen:
@nyanp : Ok, I'll try to add a gradient check for minibatch.

Also, I think the code [here](https://github.com/reunanen/tiny-cnn/blob/80272e443b10c3f8ed53d6d28da054d38aee00b8/tiny_cnn/network.h#L612-L624) should be changed to now run in parallel `n_threads` minibatches – one on each thread. Any comments or objections?

nyanp:
@reunanen
Thanks!

> I think the code here should be changed to now run in parallel n_threads minibatches – one on each thread. Any comments or objections?

I agree. We don't need this parallelism any more. Would you change it?

reunanen:
@nyanp : Sure, I'll have a look at it.

reunanen:
@nyanp
In fact, running the minibatches in parallel would change the learning results. Why? In the current version, the weights are updated after each minibatch, so results from minibatch 1 are already in use when processing minibatch 2. In my opinion, now that I think of it, changing this behavior is a major modification that may well break existing applications. So after all I think it should be introduced in another PR, or perhaps not at all.

nyanp:
@reunanen 
Thank you very much for pointing out the error!

>  In the current version, the weights are updated after each minibatch, so results from minibatch 1 are already in use when processing minibatch 2.

Hmm all weights are updated at [outside of parallel-for loop](https://github.com/nyanp/tiny-cnn/blob/master/tiny_cnn/network.h#L681), so weight should be unchanged within minibatch. Could you tell me where did you find the data-dependency?

Anyway I agree that this issue should be separated from this PR, so I'll merge it sooner.

reunanen:
@nyanp 
Yes, I meant that the current implementation is correct, while running multiple minibatches in parallel (as I [suggested](https://github.com/nyanp/tiny-cnn/pull/181#issuecomment-226988470)) might not be such a great idea after all.

nyanp:
@reunanen
Ah I misunderstood. Ok I'll merge this PR as is :)

PR #189, opened 2016-06-17
fix test_with_stb_image.cpp and typos (tiny_cnn_hrds => tiny_cnn_hdrs…

beru:
…) in CMakeLists.txt

This PR relates to another PR #167 
Somehow merge commit 73ff7ce02a753c4f4b7bdc106cae66dae658c342 didn't work out quite well for some files.
So this addresses the issue which was also mentioned by edgarriba.

And it seems to me that symbol name tiny_cnn_hrds in CMakeList.txt is a typo, so I changed it to tiny_cnn_hdrs.


nyanp:
@beru 
Thanks to nice fix! :)

PR #191, opened 2016-06-17
fix a compile error and warnings when the type float_t is a typedef of float

beru:
Hi, this PR fixes problems when the calculation data type is a typedef of float.


nyanp:
@beru 
:+1:

PR #192, opened 2016-06-17
Backend API refactor

edgarriba:
*GOAL*: split kernels into different files


edgarriba:
@nyanp Would you mind to list those layers that we should split their kernel? Currently it's done with convolutional, maxpool and fully connected. I think there are some layers like input or linear that doesn't make sense to split the kernels since they have few lines of code.

nyanp:
@edgarriba
> Currently it's done with convolutional, maxpool and fully connected

I feel it's enough now :) Probably we should split the kernel of deconv, depool and batch-normalization layer after they will be merged/implemented.

edgarriba:
@nyanp 
IMO, this PR can be merged as this.
Regarding deconv and depool @Wangyida is working on that.

@Wangyida 
Once we have merged this PR you can rebase and adapt your code.
After that we'll see if we need further modifications in the API.

nyanp:
@edgarriba
Ok! I'll merge it. Good job :)

PR #193, opened 2016-06-17
Speed optimization

beru:
This PR contains a lot of changes for speed optimization.
There are still a lot of rooms for improvement and I'm willing to follow suggestions to adjust this PR.


tiny_cnn/layers/arithmetic_layer.h
```diff
@@ -51,8 +51,8 @@ class elementwise_add_layer : public layer {
     }
 
     void forward_propagation(cnn_size_t worker_index,
-                             const std::vector<vec_t*>& in_data,
-                             std::vector<vec_t*>& out_data) override {
+                             const vec_t** in_data,
+                             vec_t** out_data) override {
```
patrikhuber: I am not really in favour of these changes. Do we really have to go into "2 star programming"? One of the reasons I like and use tiny-cnn is because of it's modern, clear code, and if we start introducing these pointers-to-pointers, everything becomes much harder to reason about.

1) Since we're passing around references, I don't see why there would be dynamic allocations in the first place? If there are, can we not solve it in a different way? The memory has to be allocated somewhere, can we not use `::reserve()` there, for example?

2) If this is the only way, can you show a speed difference of these isolated changes, showing that the loss of clarity and ability to reason about the code would be worth it?

3) If we were to adopt these changes, please at least let us add a a clear documentation to each function and the parameters.

nyanp:
@beru 
Thanks! I'll check your PR in detail after merging #181. Here are some impressions/requests about this PR, at the moment:

- All of highly optimized, x86 dependent codes should be isolated to independent file/folder, from both readability (as @patrikhuber pointed out) and portability point of view.
- Since convolution&matrix-multiply should consume most of execution time in practical dnn models, I have no idea if we would really need hardcore optimizations around activation functions. Also, these codes require fmath library - it is a bad news for tiny-cnn (header-only & dependency-free is it's identity). Could you share us your profiling experience?
- #if 0 statement and commented codes should be cleanup.
- I'd like you to use 4-spaces instead of tab :)

edgarriba:
In the case we want to introduce this optimizations I would add it in the new kernels API format as it's been done with NNPACK and always optional to the user.

nyanp:
@edgarriba 
This PR is still very useful for tiny-cnn, if it will provide a reasonable performance without external dependency. This version can be integrated to the kernel APIs as "default" backend in the future.

edgarriba:
sure. Waiting for this profiling!

bhack:
We are not in the list at https://github.com/DeepMark/deepmark/blob/master/README.md

edgarriba:
@bhack Does it makes sense before having GPU support?

bhack:
Not so much but producing compatible json output could prepare the road to DeepMark.

edgarriba:
yes, why not. I would go for it. Any json generation tool in cpp?

bhack:
https://github.com/nlohmann/json

edgarriba:
@bhack I opened another thread to discuss this https://github.com/nyanp/tiny-cnn/issues/197

beru:
[tiny-cnn-bench-beru-20160621.zip](https://github.com/nyanp/tiny-cnn/files/324379/tiny-cnn-bench-beru-20160621.zip)

I measured execution time of example_mnist_train program with latest revison in master branch and recent revisions of the branch (beru/speed_optimization). I manually took measuremens with my notebook and results should be very different based on PC performance and conditions. Also my tests are not very accurate... just an excuse.

It's a bit hard to briefly summarize all the changes I did in the past week with my poor language skills. so let me compare revisions (2cd7e4a88ec4f4511b01f307bec738fdbbae1a43 and d24c48abed02c4269311dfaaaa49937118c484a3) and describe changes.

- fmath folder
https://github.com/herumi/fmath
fmath is another person's library. I used the library to speed up activation functions (mostly tan_h class used in example_mnist_train).
To disable it, set USE_HERUMI_FMATH OFF in CMakeList.txt

- tiny_cnn/activations/activation_function.h	
For activation functions, I added implementations of new methods to enforce vectorization.
```
void f(fvec_t& dst, const fvec_t& v) const;
void f(dvec_t& dst, const dvec_t& v) const;
__m256 df(__m256 y) const;
__m256d df(__m256d y) const;
```
There is a downside with this approach.... the code is not so simple anymore...
so I will separate AVX version of activation functions into another file.
Users who want to try experimental optimized code may explicitly use them by template parameters.

- tiny_cnn/config.h	
Formerly, calculation data type was double, right now float is used and heavily optimized along the way.
Optimization of double code has not finished yet. (back_propagation is incomplete) 
```
#if 1
typedef float float_t;
#else
typedef double float_t;
#endif
```

- tiny_cnn/layers/average_pooling_layer.h	
I also optimized average_pooling_layer::forward_propagation and average_pooling_layer::back_propagation methods. They are not vectorized but still it makes a considerable difference. 

- tiny_cnn/layers/convolutional_layer_5x5.h
I mainly optimized 5x5 convolutional_layer class, I also wish to optimize 3x3 conv_layer in the future.
But to do so, I'd like to see NN example that uses 3x3 conv layer... Vectorized code is not well readable, maybe I should write a lot of comments.

- tiny_cnn/layers/feedforward_layer.h	
I vectorized float version and double version of feedforward_layer::backward_activation methods.
There is slight speed gain from it. I will try to seprate the vectorized implementation into new class in another file.

- tiny_cnn/layers/fully_connected_layer.h	
I stopped using for_i in fully_connected_layer::forward_propagation (but also in other places too)
In my case, it is simply faster in that way. I'd like to discuss about this.

I also stopped using activation function's f method inside lambda function (parameter of for_i)
Instead, I used activation function's (overloaded another) f method that takes container type arguments for input and output also.

I chose Data parallelism over Task parallelism. I wish I can take advantage of multithreading flexibly (whenever it's available) but it seems very difficult to do so.

- tiny_cnn/layers/layer.h	
I stopped using STL vector container in layer::forward and layer::backward methods.
At first I added static thread_local storage duration specifier to those local variables.
Later I changed types of those local variables to simple array of pointers.
But it is associated with prototype changes in forward_propagation and back_propagation methods,

I also changed layer::update_weight method implementation.
In the original implementation, it creates temporal diff local variable.
But it's heavy so I moved it to layer class member wdiff_.
And I replaced division in the loop to reciprocal multiplication.
and the operation is fused into implementation of edge::merge_grads to reduce load and store operations.

- tiny_cnn/layers/partial_connected_layer.h	
Stopped using for_i function since execution speed is faster in this way.
And changed the type of partial_connected_layer::in2wo_ member from std::vector<wo_connections> to wo_connections. It should be alright... please tell me if I'm wrong.

- tiny_cnn/node.h	
Optimized and vectorized edge::merge_grads method implementation.
But I need to move vectorized code into somewhere...

- tiny_cnn/optimizers/optimizer.h	
Vectorized adagrad::update method implementation. I need to quarantine those changes...

-----
I can't promise when I will post measurement of time difference with each of these changes.
I guess I'll use a lot of preprocessor to switch implementations to measure...

nyanp:
@beru 
Thank you for detail report :) In my PC(Core i7-6700K), your optimized code run 3.2sec/epoch! - much greater than I expected. But in my environment, training accuracy is far changed from original version : I got 19.0% accuracy after first epoch, and 98.94% after 15 epochs. I got yet another result if I switch-off the fmath library (58.94% accuracy after first epoch). Is it the intended behavior?

tiny_cnn/layers/arithmetic_layer.h
beru: @patrikhuber Thank you for the lucid indications and I'm so sorry for delayed response.

> I am not really in favour of these changes. Do we really have to go into "2 star programming"? One of the reasons I like and use tiny-cnn is because of it's modern, clear code, and if we start introducing these pointers-to-pointers, everything becomes much harder to reason about.

Generally speaking, modifications done by this PR extensively spoiles conciseness and esthetics of the library, so I appreciate your concern.

But I still prefer vec_t** type over std::vector<vec_t*>& type because vector container allows resizing although it isn't needed and shouldn't be allowed in this case. And checking size of container isn't performed either so I think pointers of pointers is sufficient.

There are cons and pros with vector and pointer, so opinions vary. First of all, using pointers is dangerous.

> 1) Since we're passing around references, I don't see why there would be dynamic allocations in the first place? If there are, can we not solve it in a different way? The memory has to be allocated somewhere, can we not use ::reserve() there, for example? 

If I understand correctly, dinamic memory allocations are in layer::forward and layer::backward methods. STL vector container type local variables are used in there so small sized containers are frequently constructed/destructed. C++ style solution for this is using stack storage allocator but it's not part of the standard yet. Another solution is using thread_local keyword for those local vector variables. But simple pointer array is very simple and it should meet the demand but changing the prototypes makes a lot of diff...

> 
2) If this is the only way, can you show a speed difference of these isolated changes, showing that the loss of clarity and ability to reason about the code would be worth it?

I've attached the measurement.
[tiny-cnn-bench-beru-20160625.zip](https://github.com/nyanp/tiny-cnn/files/333043/tiny-cnn-bench-beru-20160625.zip)
I used below branch.
https://github.com/beru/tiny-cnn/commits/2star_loathed

Right now, I can only use poor spec laptop, so it's hard to get constant result and it takes hours to measure and the results may be not so accurate but I guess marginal speedup is possible with the change. Other people can also check the difference with the branch. I'll measure the difference with desktop PC at the end of this month.

> 3) If we were to adopt these changes, please at least let us add a a clear documentation to each function and the parameters.

Sure, and there're a lof of things I need to do such as separating AVX optimized code, merging #181 , write tests and descriptions... so I will improve this PR slowly.

beru:
@nyanp Thank you for checking the code.

> In my PC(Core i7-6700K), your optimized code run 3.2sec/epoch! - much greater than I expected. 

If I remember correctly, the result with my desktop PC (Core i5-4679) was about 3.7sec/epoch. 

> But in my environment, training accuracy is far changed from original version : I got 19.0% accuracy after first epoch, and 98.94% after 15 epochs. I got yet another result if I switch-off the fmath library (58.94% accuracy after first epoch). Is it the intended behavior?

I guess mini-batch size is the cause. Please adjust the value in train.cpp.
I changed the value from original 10 to `std::thread::hardware_concurrency() * 3` (to distribute work evenly across threads), so in your environment, I suppose it is 16.

And to tell you the truth, I don't know the mechanisms/logics how accuracy changes with setup and code.

nyanp:
@beru 
> I guess mini-batch size is the cause. Please adjust the value in train.cpp.

Umm this can't explain why we get different results when we just switch ```USE_HERUMI_FMATH``` option. BTW, I want to integrate your great jobs with other conflicting changes (#181, #192 etc). Could you extract these optimized implementations as *non-member* functions with clear arguments ? For example:
```
// tiny_cnn/core/kernels/x86/conv2d.hpp

inline void conv2d_5x5(const shape3d&  in_shape,
                       const std::vector<float>&    in,
                       const std::vector<float>&    W,
                       const std::vector<float>&    bias,
                       std::vector<float>&          out)
{
  // your optimized code
}

template <typename T>
inline void conv2d_5x5(const shape3d&  in_shape,
                       const std::vector<T>&    in,
                       const std::vector<T>&    W,
                       const std::vector<T>&    bias,
                       std::vector<T>&          out)
{
  // "generic" version
  // If we configure tiny-cnn with double type, this version will be called
}
```

After these changes, we can easily merge to the new backend design, like a [tiny_conv2d_kernel.h](https://github.com/nyanp/tiny-cnn/blob/feat/generic-computational-graph/tiny_cnn/core/kernels/tiny_conv2d_kernel.h).

beru:
> BTW, I want to integrate your great jobs with other conflicting changes (#181, #192 etc). Could you extract these optimized implementations as non-member functions with clear arguments ?

Sure, I've made another branch to do so.
https://github.com/beru/tiny-cnn/tree/feat/generic-computational-graph-avx
Code of this branch is not as fast as below branch.
https://github.com/beru/tiny-cnn/tree/speed_optimization

Design of Backend API is really appealing, I'd like to see the same design applied to not only in
  tiny_cnn/core folder
but also
  tiny_cnn/activations
  tiny_cnn/layers
  tiny_cnn/lossfunctions
  tiny_cnn/optimizers
folders so that I can optimize thoroughly.

edgarriba:
@beru looking good! I'll comment later a couple of things in the PR.

 Regarding the backend API it's just an initial version, with some performance feedback we could optimise it with a small refactoring. Just quick comments:
-  Right now the API resides at layer level and as expected you found more use cases (activations, loss functions, etc), so maybe we can try move it to a node level instead.
- The design I think it can be improved since as you probably noticed each backend instance have empty variables depending on the used constructior.

Just thoughts in my mind for future work . /cc @nyanp

nyanp:
@beru 
Excellent! Thank you for your work :) The new branch has clean architectures, and is well integrated with the new backend design. Could you send a PR to computational-graph branch?

> Design of Backend API is really appealing, I'd like to see the same design applied to not only in
>
> - tiny_cnn/core
>
> folder, but also in
>
> - tiny_cnn/activations
> - tiny_cnn/layers
> - tiny_cnn/lossfunctions
> - tiny_cnn/optimizers
>
>folders too so that I can optimize thoroughly.

The backend designing is ongoing in #208, and we need to learn how to get a clean and fast architecture with GPU. Feel free to tell me your ideas :)

nyanp:
@edgarriba 
Thanks for the information! You can see that MxNet has similar auto registering mechanics:

https://github.com/dmlc/mxnet/blob/master/src/operator/crop.cc

Some people like these "magic" in C++, but this makes debugging and understanding source code more and more difficult. I like a straightforward approach without macro. It seems to be the magic vs boilerplate trade-off for me:

http://www.johndcook.com/blog/2016/03/07/the-magic-vs-boilerplate-tradeoff/

But it is true that adding a new layers/ops to tiny-cnn is harder task than TF, so we want to adopt a good part of TF.

edgarriba:
@nyanp I'll take a look into this magic stuff. Regarding the GPU implementation, we need to create a memory transfer modules. Caffe has [this](https://github.com/BVLC/caffe/blob/opencl/include/caffe/syncedmem.hpp) and TF [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/copy_tensor.cc)

edgarriba:
I think we can close this since it's merged in https://github.com/nyanp/tiny-cnn/pull/215.

*Note*: @beru @nyanp AVX support need revision since some tests are broken

beru:
@edgarriba  Thank you for the notification.

PR #195, opened 2016-06-19
Deconvolution modification for backend using

Wangyida:


tiny_cnn/core/conv_params.h
```diff
@@ -40,6 +40,13 @@ struct conv_layer_worker_specific_storage {
     vec_t prev_delta_padded_;
 };
 
+struct deconv_layer_worker_specific_storage {
+    const vec_t* prev_out_;
+    vec_t* curr_out_unpadded_;
+    vec_t curr_out_buf_;
+    vec_t curr_delta_padded;
+};
+
```
Wangyida: @edgarriba According to the backend you have contributed, maybe I should get some suggestion from you to better adapt my codes with the backend using. I want to use a standalone worker storage here for deconvolutional padded delta and unpadded output.

tiny_cnn/core/conv_params.h
```diff
@@ -90,6 +97,10 @@ struct conv_params {
     padding pad_type;
     size_t w_stride;
     size_t h_stride;
+    // deconvolution parameters
+    index3d<cnn_size_t> in_;
+    index3d<cnn_size_t> out_;
+    index3d<cnn_size_t> out_unpadded_;
```
Wangyida: @edgarriba And here, parameter for convolution and deconvolution could be in the same structure.

tiny_cnn/core/tiny_backend.h
```diff
@@ -35,13 +35,24 @@ class tiny_backend : public math_backend {
...
+        backward_activation(f3) {}  
+
+  // deconvolution      
+  tiny_backend(conv_params* params,
+               std::function<void(const vec_t&, int)> f1,
+               std::function<void(const vec_t&, vec_t&)> f2,
+               std::function<void(const vec_t&, const vec_t&, vec_t&)> f3,
+               std::vector<deconv_layer_worker_specific_storage>* ptr1)
+      : params_(params), deconv_layer_worker_storage_(ptr1),
+        copy_and_unpad_output(f1), copy_and_pad_delta(f2),
```
Wangyida: @edgarriba And here, the most important part. Is it proper to add another backend for deconvolution? I want to get access to different worker storage between convolution and deconvolution because the deconvolutional Class should have its own initiating function.

tiny_cnn/core/tiny_backend.h
```diff
@@ -214,6 +225,139 @@ class tiny_backend : public math_backend {
...
+       for (cnn_size_t outc = 0; outc < params_->out.depth_; outc++) {
+           cnn_size_t idx = params_->out.get_index(0, 0, outc);
+           const float_t * delta = &curr_delta[idx];
+           const float_t * deltaa = delta + params_->out.width_ *
+                                            params_->out.height_;
+           db[outc] += std::accumulate(delta, deltaa, float_t(0));
+       }
+    }
+    }
+
```
Wangyida: I think these two functions are defined properly here.

edgarriba:
@Wangyida Regarding to all of your questions check last commits. There, I split core functions into different kernels. Still I need to iterate for a generic design. If you have any concern just tell me.

Wangyida:
@edgarriba @nyanp OK, then the codes will be made more generic based on a worked one.

Wangyida:
@nyanp I rewrite the deconv codes for new backend, all test passed, ready for merge.

Wangyida:
@edgarriba @nyanp Oh, I just noticed that the code structure is merged with a new PR 192 by Edgar, Thanks. let me have a rebase again.

tiny_cnn/core/tiny_backend.h
edgarriba: @Wangyida I think that a good desgin could be having an instance of an abstract `class Workers` inside the `Backend`.

tiny_cnn/core/backend.h
```diff
@@ -60,6 +61,16 @@ class backend {
                         std::vector<vec_t*>&       out_grad,
                         std::vector<vec_t*>&       in_grad) = 0;
 
+    virtual void deconv2d(cnn_size_t                 index,
+                          const std::vector<vec_t*>& in_data,
+                          std::vector<vec_t*>&       out_data) = 0;
+
+    virtual void deconv2d_back(cnn_size_t                 index,
```
edgarriba: just overload `deconv2d()`

tiny_cnn/core/backend_tiny.h
```diff
@@ -213,18 +282,22 @@ class tiny_backend : public backend {
...
     std::vector<std::vector<cnn_size_t>>* out2in_;
     std::vector<cnn_size_t>* in2out_;
 
     /* Pointers to parent class functions */
     std::function<void(const vec_t&, int)> copy_and_pad_input;
+    std::function<void(const vec_t&, int)> copy_and_unpad_output;
     std::function<void(const vec_t&, vec_t&)> copy_and_unpad_delta;
+    std::function<void(const vec_t&, vec_t&)> copy_and_pad_delta;
     std::function<void(const vec_t&, const vec_t&, vec_t&)> backward_activation;
 };
```
edgarriba: @nyanp @Wangyida I think that we need to refactor the private attributes in a more generic way. Ideas?

tiny_cnn/core/kernels/tiny_deconv2d_back_kernel.h
```diff
@@ -0,0 +1,111 @@
...
+
+            for (cnn_size_t y = 0; y < params.in_.height_; y++) {
+                for (cnn_size_t x = 0; x < params.in_.width_; x++) {
+                    const float_t * ppw = pw;
+                    float_t * ppdelta_dst = pdelta_dst + y * params.in_.width_ + x;
+                    float_t sum = float_t(0);
+
+                    for (cnn_size_t wy = 0; wy < params.weight.height_; wy++) {
+                        for (cnn_size_t wx = 0; wx < params.weight.width_; wx++) {
+                            sum += ppw[wy * params.weight.width_ + wx] * pdelta_src[(y+wy) * params.h_stride * params.in_.width_ + (x+wx) * params.w_stride];
```
edgarriba: @Wangyida Please, follow the style. Lines lenght <= 80  
https://google.github.io/styleguide/cppguide.html

tiny_cnn/layers/deconvolutional_layer.h
```diff
@@ -369,6 +318,30 @@ class deconvolutional_layer : public feedforward_layer<Activation> {
...
 private:
+    void init_backend(backend_t backend_type) {
+        using namespace std::placeholders;  // for _1, _2, _3...
+        switch (backend_type) {
+            case backend_t::tiny_cnn:
+                Base::backend_ = std::make_shared<core::tiny_backend>(&params_,
+                    std::bind(&deconvolutional_layer::copy_and_unpad_output, this, _1, _2),
+                    std::bind(&deconvolutional_layer::copy_and_pad_delta, this, _1, _2),
+                    std::bind(&Base::backward_activation, this, _1, _2, _3),
+                    &deconv_layer_worker_storage_);
```
edgarriba: @Wangyida use lambdas here like in convolutional_layer

tiny_cnn/layers/deconvolutional_layer.h
```diff
@@ -377,113 +350,120 @@ class deconvolutional_layer : public feedforward_layer<Activation> {
...
+                                                cnn_size_t stride, padding pad_type) {
+        return pad_type == padding::same ? 
+                (cnn_size_t)ceil((float_t)in_length * stride) : 
+                (cnn_size_t)ceil((float_t)(in_length) * stride + window_size - 1);
     }
 
-    static cnn_size_t deconv_out_dim(cnn_size_t in_width, cnn_size_t in_height, cnn_size_t window_size, cnn_size_t w_stride, cnn_size_t h_stride, padding pad_type) {
-        return deconv_out_unpadded_length(in_width, window_size, w_stride, pad_type) * deconv_out_unpadded_length(in_height, window_size, h_stride, pad_type);
+    static cnn_size_t deconv_out_dim(cnn_size_t in_width, 
+                                    cnn_size_t in_height,
```
edgarriba: formatting: one more space here

tiny_cnn/core/nnp_backend.h
```diff
@@ -0,0 +1,240 @@
+/*
```
edgarriba: @Wangyida sorry, I renamed some files. Check to not add again those files.

tiny_cnn/layers/deconvolutional_layer_.h
```diff
@@ -0,0 +1,483 @@
+/*
```
edgarriba: I think this file is not needed anymore

Wangyida:
@edgarriba Thanks for your valuable advises for the coding details, I will adjust them for your demand.

edgarriba:
I'm not familiar with deconvolution details. Is forward the same as in convolution?
Because it's supposed that NNPACK only supports convolution operation.
I'll prepare some some skeleton tests for NNPACK so that we can test it.

@nyanp We can think on run tests using all available backend engines. Not sure how to handle that.

Wangyida:
@edgarriba The forward pass is not the same as convolution, I don't know whether nnpack support decovnolution operation, it's more likely as a upsampling process rather than linear operation using vector multiplication.

edgarriba:
in that case I think that deconvolution cannot be handled by NNPACK.

Wangyida:
@edgarriba But the forward pass process still could be implemented just using convolution, there are several different ways to implement deconvolution. For my case, I implemented it as an upsampling way for an intuitive operation for understanding based on the original definition of deconvolution.

I am sure it can be implemented using just convolution afterwards.

Wangyida:
@nyanp I have complete the adjustment for new backend data structure contributed by @edgarriba, I tested the deconv example on MNIST and the tests codes passed, so now it is able to be got merged.
Thanks for the efforts of Edgar!

bhack:
@edgarriba If you want to know a little bit more on this operator take a look at http://datascience.stackexchange.com/questions/6107/what-are-deconvolutional-layers

edgarriba:
Thanks @bhack!

@Wangyida
nice ! There are still some old files that should be removed.

tiny_cnn/core/backend.h
```diff
@@ -60,6 +61,16 @@ class backend {
                         std::vector<vec_t*>&       out_grad,
                         std::vector<vec_t*>&       in_grad) = 0;
 
+    virtual void deconv2d(cnn_size_t                 index,
+                          const std::vector<vec_t*>& in_data,
+                          std::vector<vec_t*>&       out_data) = 0;
+
+    virtual void deconv2d(cnn_size_t                 index,
+                               const std::vector<vec_t*>& in_data,
```
edgarriba: Space format

bhack:
@nyanp We want to rename this as transposed convolution 2d?

tiny_cnn/core/dnn_backend.h
```diff
@@ -0,0 +1,80 @@
+/*
+    Copyright (c) 2016, Taiga Nomi, Edgar Riba
```
edgarriba: This file should be removed

Wangyida:
@edgarriba @nyanp OK, spaces are removed and all lines in my updating codes are linted. index operation for the kernels are modified as the convolution style where key index are labeled as 'idx'.

nyanp:
@bhack 
Either is fine for me. Caffe calls it as ```deconvolution```, meanwhile TensorFlow calls it as ```conv2d_transpose``` (I think cv community tend to use deconvolution). In such case, I prefer to provide both names as alias ([example](https://github.com/nyanp/tiny-cnn/blob/master/tiny_cnn/tiny_cnn.h#L81-85) - "fc" is from caffe/tf, "dense" is from keras).

@Wangyida 
Good job, thank you! I'll add some alias names after merging.

edgarriba:
@nyanp Will you implement this PR in the new API ?

PR #198, opened 2016-06-21
add NNPACK check status and test

edgarriba:
@nyanp @Wangyida  @bhack @mtamburrano
Could you test NNPACK in your own devices? Getting a segfault here


bhack:
The Segfault was on the AVX2 flag?

edgarriba:
nope. Still not working

tiny_cnn/core/backend_nnp.h
```diff
@@ -87,29 +92,25 @@ class nnp_backend : public backend {
...
         if (params_d_->w_stride != 1 || params_d_->h_stride != 1) {
             throw nn_error("NNPACK Convolution requires stride 1.");
         }
 
-#ifdef CNN_USE_NNPACK
         deconv_layer_worker_specific_storage& cws =
             (*deconv_layer_worker_storage_)[index];
-        copy_and_pad_delta(cws.curr_delta_padded, *in_grad[0]);
+        copy_and_pad_delta(cws.curr_delta_padded, *in_grad[0]); //  *in_grad[0] does not exist here!
         const vec_t& W    = *in_data[1];
```
edgarriba: @Wangyida Are you sure that you compiled this? There are variables that are not declared

tiny_cnn/core/kernels/nnp_deconv2d_kernel.h
```diff
@@ -35,21 +35,21 @@ namespace tiny_cnn {
 namespace core {
 namespace kernels {
 
-nnp_deconvolution_algorithm nnp_algorithm() {
```
edgarriba: @Wangyida this doesn't exist. You should check that deconv works with NNPACK

tiny_cnn/core/kernels/nnp_maxpool_kernel.h
```diff
@@ -0,0 +1,97 @@
...
+        static_cast<size_t>(params.in_.height_)
+    };
+
+    const nnp_padding input_padding = {
+        /*static_cast<size_t>(this->pad_h_),
+        static_cast<size_t>(this->pad_h_),
+        static_cast<size_t>(this->pad_h_),
+        static_cast<size_t>(this->pad_h_)*/
+    };
+
```
edgarriba: @nyanp Any tip about how can I get these parameters?

tiny_cnn/core/kernels/nnp_deconv2d_kernel.h
bhack: @Wangyida you can open an upstream issue in NNPACK

bhack:
@edgarriba How the segfault is reproducible?

bhack:
@edgarriba I've executed the test and i got an error. Have you executed init_nnp_engine()?

bhack:
```
terminate called after throwing an instance of 'tiny_cnn::nn_error'
  what():  ��m[ERROR] NNPACK Max-Pool requires a stride == 2.
```

edgarriba:
good point ! Unfortunately my laptop doesn't support AVX2. 

If I compile with -mavx2 flag I got the following error: `Illegal instruction (core dumped)`. In the other case, NNPACK raises the following error: `NNPACK does not implement this function for the host CPU`.

bhack:
Arg! We have avx2 at work. But I've only avx on laptop.
Avx2 require a quite new CPU. How many kernels are avx2 only?

bhack:
P.s. Avx2 was introduced with Haswell arch.

tiny_cnn/core/backend_nnp.h
Wangyida: Here is a mistake, line 101:
### 1
```cpp
copy_and_pad_delta(cws.curr_delta_padded, *in_grad[0]);
```
should be replaced as
```cpp
(*deconv_layer_worker_storage_)[index].prev_out_ = in_data[0];
```
### 2
and line 166:
```cpp
std::function<void(const vec_t&, vec_t&)> copy_and_pad_delta;
```
should be replaced as:
```cpp
std::function<void(const vec_t&, int)> copy_and_unpad_output;
```

You can modify them directly, output of deconvolution might be unpadded and the delta in backforward might be padded, it's just the opposite operation of convolution in padding strategy.
I haven't compile with nnp backend as nnPack is not installed yet. These codes are modified logically in nnp.

tiny_cnn/core/kernels/nnp_deconv2d_kernel.h
Wangyida: @bhack Yeah I noticed this just when @edgarriba informed that NNPack just support for typical convolution, I have another method to implement deconvolution using convolution. That will make tiny_cnn able to run deconv based on NNPack.

The method detail is roughly about:
1. padding input with zeros of **2(weight.width-1)** rather than **weight.width-1** in convolution with padding::same strategy.
2. reverse the convolution kernel in 2D space and remain the channel sequence.

bhack:
@Maratyszcza Needs all the kernels in NNPACK avx2?

bhack:
@edgarriba You can execute tests with (Intel SDE)[https://software.intel.com/en-us/articles/intel-software-development-emulator/] or Qemu to emulate avx2

edgarriba:
@bhack oki,. I'll try

edgarriba:
@bhack geetting the same. I updated the test, the parameters were wrong.

ajtulloch:
Make sure all the unit tests for the Caffe integration pass on your machine?

edgarriba:
@ajtulloch thx for your answer.
I've tried nnpack-pr branch and all tests passes except InnerProductLayerTest/0.TestGradientNNPACK. Check [log.txt](https://github.com/nyanp/tiny-cnn/files/328361/log.txt)

@bhack @nyanp If tests for convolution and pooling works in Caffe's version it means that we are doing something wrong.

bhack:
@edgarriba  Are you running with Intel SDE?

edgarriba:
@ajtulloch @Maratyszcza I noticed that `nnp_max_pooling_output` is running, however it deallocates the content of my output_ptr. Any idea?

Maratyszcza:
@edgarriba An issue in convolution backpropagation came up during Caffe integration (see Maratyszcza/NNPACK#15), but I believe it is fixed now. AFAIK, there aren't any unsolved crash or correctness bugs.

As for `nnp_max_pooling_output`, you may want to check how you compute the output size (and compare it with NNPACK). I remember that there are some variations in output size between packages, and NNPACK follows Caffe & cuDNN convention.

Maratyszcza:
@bhack I'd expect that all kernels require AVX2 or at least FMA3 (which means Haswell+ among Intel CPUs). However, you may build NNPACK with clang with `--enable-psimd` flag. Performance wouldn't be as good, but it will work everywhere, and probably suit your development needs.

bhack:
@Maratyszcza Thank you. I've already executed with Intel sde on a cpu without avx2 but your option it is useful cause doen't rely on closed source tools. There will be a common maintainer ship of NNPACK with the Facebook group or will remain a personal project?

Maratyszcza:
@bhack I heard you can emulate new instruction sets with qemu if Intel SDE doesn't suit your needs.

For the next year I am still a Ph.D. student at Georgia Tech, but NNPACK is developed in collaboration with F.A.I.R., and it is production-quality code.

Wangyida:
This is a major PR, so I think that some bug fixes in my local machine is better not being pushed as a new PR that will make it a additional task for @edgarriba to rebase again. I think I could push them after this PR got merged, am I right?
BTW, Line 141 of CMakeLists.txt might be better replaced as:
```
LINK_LIBRARIES(${REQUIRED_LIBRARIES})
```
Some OpenCV related libs would be regarded placed in the current directory if it is written as:
```
link_directories(${REQUIRED_LIBRARIES})
```

edgarriba:
@bhack @nyanp NNPACK working. We're good to merge. Next LibDNN.

@Maratyszcza @ajtulloch thx for your support !
@Maratyszcza Can we link tiny-cnn to NNPACK project readme?

nyanp:
@edgarriba 
OK, great!! (I'm sorry but I wasn't enough time for the past few days to support your debbuging...) Let's go next :)

Maratyszcza:
@edgarriba Referenced tiny-cnn in NNPACK readme. Or did you mean something else?

edgarriba:
@Maratyszcza That's fine. Thanks a lot !

Issue #197, opened 2016-06-20
Add Tiny-cnn to DeepMark


edgarriba:
I open this thread to discuss about to add tiny-cnn into [DeepMark](https://github.com/DeepMark/deepmark) benchmarks.

Currently, tiny-cnn only supports CPU forward and backward.
While GPU support is not added, we can start preparing scripts to generate the JSON files.

- Output must in the following format: https://github.com/autumnai/deep-learning-benchmarks
- Proposed tool for JSON generator: https://github.com/nlohmann/json

patrikhuber:
:+1: for `nlohmann/json`

PR #207, opened 2016-06-27
Add Benchmarks series

edgarriba:


edgarriba:
/cc @Wangyida @bhack feel free to comment, add new models, etc.

In this PR I just took Alexnet implementation from DeepMark
https://github.com/DeepMark/deepmark/blob/master/torch/image%2Bvideo/alexnet.lua

nyanp:
@edgarriba 
Should we conform to deepmark format? #197

edgarriba:
@nyanp I would go for splitting benchmark models itself and serialization in DeepMark format. Just to not have many third-party depencies in CMake. However, nlohmann/json is a header lib. As you wish.

Merge it only to have a basic example. Probably need to be modified

examples/benchmarks/alexnet.h
```diff
@@ -0,0 +1,48 @@
...
+*/
+
+#include <iostream>
+#include "tiny_cnn/tiny_cnn.h"
+
+using namespace tiny_cnn;
+using namespace tiny_cnn::activation;
+using namespace tiny_cnn::layers;
+
+class AlexNet : public network<sequential> {
```
nyanp: @edgarriba 
It would be interesting if we move these model class into tiny-cnn's core (under ```tiny_cnn::models``` namespace, for example). It's convenient for users who just want to use off-the-shelf network model.

nyanp:
@edgarriba 
Ok :) We can add a deepmark style output after the libdnn integration.

examples/benchmarks/alexnet.h
edgarriba: oki. Seems fair to me

edgarriba:
@nyanp From my side it can be merged. We should check the Alexnet model since I got a segfault during inference.

nyanp:
@edgarriba Great! I couldn't reproduce the error, could you specify where you got a segfault?

edgarriba:
oh, really? have you tried to launch `./benchmarks ` ??

nyanp:
@edgarriba
Yes, with win10/msvc2015.

examples/benchmarks/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+int  main(int argc, char** argv) {
+    
+    models::alexnet nn;
+
+    // change all layers at once
+    nn.weight_init(weight_init::lecun());
+    nn.bias_init(weight_init::xavier(2.0));
+
+    vec_t in(224*224*3, bernoulli(0.5));
```
nyanp: ```bernoulli``` returns a single boolean value. If you want to get a random sequence, you can use ```uniform_rand(in.begin(), in.end(), 0, 1)``` for real values, and ```std::generate(in.begin(), in.end(), [](){ return bernoulli(0.5); })``` for boolean values.

Wangyida:
@edgarriba Just a remind for you: if you build tiny_cnn in ```tiny_cnn/build``` folder rather than ```tiny_cnn``` folder, when you ```cmake ..``` and ```make all -j4``` for the first time, it might alert you with:
```
fatal error: 'caffe.pb.h' file not found
```
**Solution** is ```cmake ..``` and ```make all -j4``` again. I don't know how to fix this now and it is caused by the compiled caffe.pb.h file is not put in a suitable place such as rootpath.

edgarriba:
@nyanp updated with your suggestion but still getting a segfault. Check the [log](http://pastebin.com/NhQDG4ff)

@Wangyida noticed that before. What's the proper solution then?

Wangyida:
@edgarriba I haven't found a good solution I used to copy caffe.pb.h from the building directory to the right path, but this only works in Ubuntu 14.04, it still needs to be ```cmake ..``` again on Mac.

nyanp:
@edgarriba 
We should add the following line into max_pooling_layer's move constructor:
```cpp
init_backend(std::move(Base::get_backend_type()));
```

edgarriba:
@nyanp awesome, fixed!

nyanp:
@edgarriba 
Good :) I'll merge it.

PR #208, opened 2016-06-29
LibDNN integratio

edgarriba:
This PR introduces LibDNN into tiny-cnn as detailed below:

- Forward propagation for convolutional layer


tiny_cnn/core/kernels/libdnn_conv2d_kernel.h
```diff
@@ -0,0 +1,98 @@
...
+    const float_t* weights_ptr = reinterpret_cast<const float_t*>(&W[0]);
+    const float_t* bias_ptr    = reinterpret_cast<const float_t*>(&bias[0]);
+
+    float_t* output_ptr = reinterpret_cast<float_t*>(&a[0]);
+    
+    const int batch_sz = 1;
+    
+    // call libdnn kernel
+    kernel.Forward(input_ptr, weights_ptr, bias_ptr, output_ptr, batch_sz);
+}
```
edgarriba: @naibaf7 Nothing seems to happen when running test. Do you feel that I miss something?

naibaf7:
@edgarriba 
Oh, some hints:
- A valid device* must be passed to LibDNNConv (in the config, as `dev_ptr`), there must be a viennacl-context set up for it as well. To do that, look at this function: `device::setupViennaCLContext` and at the `device` class. 
- The memory type is not `float_t*` but `cl_mem_*` aliased as `float_t*`. This is to ensure CUDA compability (although it is not very pretty). Do a pointer cast from `cl_mem` to `float_t*`. CPU pointers are not supported here.

naibaf7:
Some example code to make this more clear:
```    LibDNNConfig config;
    config.dev_ptr = this->device_;
    config.in_shape = bottom[0]->shape();
    config.out_shape = top[0]->shape();
    config.kernel = kernel_vec;
    config.pad = pad_vec;
    config.stride = stride_vec;
    config.dilation = dilation_vec;
    config.group = this->group_;
    config.bias_term = this->bias_term_;
    config.fast_unsafe_math = true;
    config.weights_backward = this->param_propagate_down_[0];
    config.bias_backward = this->param_propagate_down_[1];

    if (std::is_same<Dtype, float>::value ||
        this->device_->CheckCapability("cl_khr_int64_base_atomics")) {
      config.wgalgo = LIBDNN_CONVOLUTION_WG_ALGO_ATOMIC;
      config.bwalgo = LIBDNN_CONVOLUTION_BW_ALGO_COL2IM_ATOMIC;
    } else {
      config.wgalgo = LIBDNN_CONVOLUTION_WG_ALGO_DIRECT;
      config.bwalgo = LIBDNN_CONVOLUTION_BW_ALGO_IM2COL;
    }

    LibDNNConv<Dtype>* libdnn = new LibDNNConv<Dtype>(config);```

and:
```
  const Dtype* weight = this->blobs_[0]->gpu_data();
  const Dtype* bias = nullptr;
  if (this->bias_term_) {
     bias = this->blobs_[1]->gpu_data();
  }

  for (int_tp i = 0; i < bottom.size(); ++i) {
    const Dtype* bottom_data = bottom[i]->gpu_data();
    Dtype* top_data = top[i]->mutable_gpu_data();
    libdnn_.get()->Forward(bottom_data, weight, bias,
                           top_data, bottom[i]->shape()[0]);
  }
```
(note here, gpu_data() in Caffe returns either `cl_mem_*` or CUDA-`Dtype*`, never CPU-`Dtype*`).

**Comment by naibaf7** [comment]
You should also take care to store the `LibDNNConv` instance somewhere. It is expensive to recreate it for every forward/backward step, as this triggers kernel compilation.

**Comment by bhack** [comment]
@naibaf7 Are you parially trying to solve similar problems using viennacl and @hughperkins [easycl](https://github.com/hughperkins/EasyCL/blob/master/README.md)?

**Comment by naibaf7** [comment]
@bhack
Yes that seems similar. ViennaCL is used for device management, error handling and kernel compile & launch in both Caffe and libDNN.
I found that ViennaCL can properly handle device initialization and error handling (also during kernel launch) where other implementations fail (for example AMD's OpenCL Caffe will fail on many mobile platforms such as Optimus and AMD's dGPU+iGPU solutions, while ViennaCL based Caffe works flawlessly). And I also just like the abstraction. Having it available is also very handy when a simple BLAS call is needed and then often compiling clBLAS or CLBlast becomes unnecessary/optional.

**Comment by edgarriba** [comment]
@naibaf7 thx for your support! A quick question: is there any reason for not having `void device::setupViennaCLContext()` inside of `void device::Init()` ? Since `setup-context` needs `cl_device` and `cl_command_queue` I think it will be easier to use OpenCL as much as posible only inside libdnn.

> (note here, gpu_data() in Caffe returns either cl_mem_* or CUDA-Dtype*, never CPU-Dtype*).

@nyanp @bhack Does it mean that we need to keep tracked gpu data inside `class edge` ?

**Comment by naibaf7** [comment]
@edgarriba 
I could add a simplified interface where everything OpenCL (memory transfers, device initialization) is kept inside libDNN, but I am currently in holidays, so I only have time to do small fixes for the next 2 weeks.
Would you want something like that? For the device initialization and handling I can agree, there is already the necessary logic for that inside Caffe and I could move these parts to libDNN.
For memory handling, I STRONGLY recommend keeping track of GPU data inside your implementation, because all the host-to-device transfers will degrade your performance massively.

**Comment by edgarriba** [comment]
@naibaf7 I would appreciate that interface. From our side we need to revise how to handle GPU data.

**Comment by edgarriba** [comment]
@nyanp 
I've some ideas for sesion/device api that can be helpful for this PR and future backend refactoring 

/cc @bhack 

    template<typename T>
    class device {
      device(const int id) {}
 
      void register_op(const vector<layer>);
      void register_op(const vector<node>);
      void register_op(const network<sequential>);
      void register_op(const network<graph>);
 
      void register_op(...) {
          ptr_list.push_back(obj_ptr);
          obj_ptr.set_device(this);
      }
 
    private:
       T type_;
    #ifdef HAVE_LIBDNN
        greentea::device dev_ptr_;
    #endif
    };
 
    // USAGE 
    auto conv1 = make_shared<conv>(...);
    auto conv2 = make_shared<conv>(...);
    auto conv3 = make_shared<conv>(...);
 
    // similar to TF: '/gpu:0'
    device my_device<GPU>(0).register_op({ conv1, conv2, conv3 });
 
    session my_session("my_session").register_device({ my_device });

**Comment by nyanp** [comment]
@edgarriba 
Oh it's really cool :)

File: CMakeLists.txt [hunk 2]
```diff
@@ -212,6 +212,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
     set(EXTRA_C_FLAGS "${EXTRA_C_FLAGS} -Wall -Wpedantic")
     set(EXTRA_C_FLAGS_RELEASE "${EXTRA_C_FLAGS_RELEASE} -O3")
     set(EXTRA_C_FLAGS_DEBUG   "${EXTRA_C_FLAGS_DEBUG} -g3")
+
+    # TODO(nyanp): could you check why do we need this even when
+    # USE_TBB is enabled? I got the following errors during linkage:
+    # /usr/bin/ld: CMakeFiles/tiny_cnn_test.dir/test.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
+    # //lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line 
+    set(EXTRA_C_FLAGS_DEBUG   "${EXTRA_C_FLAGS_DEBUG} -pthread")
+
```
edgarriba: @nyanp any idea?

nyanp: @edgarriba 
It is because NNPack and gemmlowp requires pthread to utilize their own threadpool. I think replacing them with std::thread is the best way for tiny-dnn (as @back [commented](https://github.com/nyanp/tiny-cnn/pull/206#issuecomment-233435944)).

patrikhuber: If I remember correctly, on Linux (or only on gcc?), even the use of `std::thread` requires `-pthread`.

bhack: http://stackoverflow.com/questions/11784153/in-g-is-c-11-thread-model-using-pthreads-in-the-background

tiny_cnn/core/framework/device_base.h
```diff
@@ -47,10 +47,10 @@ class device_base {
         : id_(id), type_(type) {}
 
     // Register an ops to the current device
-    void register_ops(const std::vector<layer*>& ops) {
+    void register_op(const std::vector<layer*>& ops) {
         for (auto o: ops) {
             ops_.push_back(o);
-            // TODO: o.set_device(this);
+            o->set_device(this);
```
edgarriba: @nyanp We have a cyclic dependency problem here
Check the compilation error here: http://pastebin.com/rVE45i0q

edgarriba:
Hi! so the current situation is the following:

For GPU implementation we need somehow to redesign low level API. I've introduced the TF `session` concept since for launching layers/ops kernels we need to retrieve and share devices context. My first approach was to directly link with pointers layers/ops and devices. With this I entered in a circular dependency class issue due to our header-project architecture.

Another possibility could be @bhack proposals which `session` will work as main were we will register each layer/op, devices and they relationships. From `session.run()` method we could everything, even launching layers kernels and keep aside current layers API.

So, big low level decisions that will be the keystone of our framework. Let's discuss about that.

/cc @bhack @nyanp @Wangyida @mtamburrano @naibaf7

PD: we coud find another cool name to represent `session` just to not be equal to TF. :)

tiny_cnn/layers/layer.fwd.h
```diff
@@ -0,0 +1,336 @@
+/*
```
edgarriba: @bhack @nyanp 
I had to adapt layer class into different files in order to avoid errors during forward declaration.

It's the same behaviour as in normal projects where you have class definition in *.hpp its implementation in *.cpp.
Here definition is in `layer.fwd.h` and implementation in `layer.h`

nyanp:
@edgarriba @bhack 
Using ```session.run``` as main is an interesting approach, but I would like to keep current API. In some user scenarios, especially in embedded systems, they don't want tiny-cnn to invoke background thread implicitly.

Splitting header into 2 files is good idea :)

bhack:
Why session need to be background? Could not be sync?

nyanp:
@bhack 
Oh I probably mistook the meanings of "session will work as main". Do you mean just letting user call ```session.run()``` to execute forward/backward calculation?

bhack:
Yes.. Probably there will be room there to plug any "smart" algorithm that could optimize dag, device placements and its scheduling in the future.

nyanp:
Thanks, I got it. We already have ```graph``` class which manage DAG and [execute ops](https://github.com/nyanp/tiny-cnn/blob/master/tiny_cnn/nodes.h#L324-L343), so it can be the room for these kind of optimizations.

PR #225, opened 2016-07-09
minor fix, mainly indentation

beru:
added missing inline keyword, fixed indentations, etc...


nyanp:
@beru 
Nice, thanks! :+1:

PR #206, opened 2016-06-27
quantization, bug fix in deconv and graph enet

Wangyida:
## Task list:
- [x] Basic functions and pass the tests on core quantization utilities.
- [x] Quantized convolution layer and pass the tests for q_conv.
- [x] Quantized deconvolution layer and pass the tests for q_deconv.
- [x] Quantized fully connected layer and pass the tests for q_fully_connected.
- [x] Quantized bias inside other kernels.
- [x] Ensure acceptable accuracy on typical examples.
- [x] Add low precision gemm from Google TF as a kernel for matmul.
- [x] Resolve unnecessary quantize-dequantize procedure.
- [x] Add backward propagation for quantization.
 
## Method:
The method is totally the same as described in [Pete's blog for quantization](https://petewarden.com/2016/05/03/how-to-quantize-neural-networks-with-tensorflow/) and my basic codes are modified from [TensorFlow quantization module](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/quantization).

Issue #186, opened 2016-06-15
Branching policy in tiny-cnn


nyanp:
Increasing number of PR's in tiny-cnn (#166, #181, #182) requires some branching policy. I think the following policy will work, but I want feedback:

- feat/generic-computational-graph should focus on opencv gsoc project (device abstraction, NNPack/libdnn integration and so on by @edgarriba  )
- another PR's should be sent to master branch, basically
- master branch should be merged into feat/generic-computational-graph when "heavy" PR is accepted to master (#181, for example)
    - but too heavy PR should be reviewed by opencv team before accepting, to avoid painful confliction

@bhack @edgarriba @reunanen @Wangyida If you have any thoughts on that please share it with me :)

edgarriba:
seems okay to me

Wangyida:
OK well, some functions merged in graph branch still have bugs in some condition (such as the unpool layer from me, sorry for hash push). Tested version could be merged into master.

edgarriba:
There are some functionalities that are under development like NNPACK and still not working.
A good strategy will be merging after we solve that issue. Besides, there other features in master branch that must be adapted to computational-graph (recent batch processing)

nyanp:
@edgarriba /cc @Wangyida  
Nice idea. I want to solve problems in the following order (all works are on computational-graph branch):

1. Fix issues around NNPACK (#198) by @edgarriba 
1. Merge other features in master branch into computational-graph (#181 etc) by @nyanp
1. Fix deconvolution crashes (#199) by @Wangyida

edgarriba:
good plan!

nyanp:
@edgarriba 
> I think that after this next big merge we should rename or treat the computational-graph branch as dev and force all pull requests there.

Great idea. I've updated the task list.

bhack:
Probably we can start to use labels.

Randl:
@nyanp Should we close the issue and move some generic version of this to [CONTRIBUTING.md](https://github.com/tiny-dnn/tiny-dnn/blob/master/CONTRIBUTING.md)?

PR #226, opened 2016-07-10
Add batch normalization

nyanp:
TODO:
- [x] Implement backprop
- [x] Add unit tests
- [x] Add caffe importer


bhack:
We need also quantized BatchNorm when this is ready /cc @wangyida

Wangyida:
@bhack OK of course.

edgarriba:
@nyanp why committing to master instead of feat/computational-graph? I see here a lot of interesting things for a further refactoring for backend and operation kernels

https://github.com/edgarriba/tiny-cnn/blob/backend/tiny_cnn/core/kernels/identity_op.h#L38

IMO, we should converge all current developments in a `dev` branch as we previously disscussed.

nyanp:
@edgarriba
This is a successor of the #181, which is based on master branch. I agree that we should go on in a dev branch, but I want to confirm that we can implement BN on the #181 style API before merging into graph branch.

bhack:
Better to use tag on master and use master branch directly. So we could maintain only long feature branches and PR.

nyanp:
@edgarriba @Wangyida
After this PR is completed, I'll merge a master branch into a graph branch (as discussed in #186). Since this merging contains some changes around layer API, you'll need to adopt your ongoing works with this. When is the best time for you to merging?

nyanp:
@bhack 
Thanks for your advice, do you think the current GSoC project should go on a master branch?

bhack:
When you are fine with the branch status it is ok for me. What matters it is that the master branch is always  in shape to be run (always test coverage before merging and review a little bit severe without a dev branch). The advantage is to have master always on play by users and other developers.

nyanp:
@bhack 
It's ok for me about the current branches, so I'll keep going on this way. Thx :)

@edgarriba @Wangyida 
I think it's better to merge #181 and this PR into ```graph``` branch at some time in the near future (as discussed in #186), because these PRs affect a policy of parallelism in tiny-cnn. When is the best timing for you to merge these changes into ```graph``` branch?

edgarriba:
from my side still fixing some issues with new CMake and planning some improvements to have a proper development toolkit:
 - Add dockefile
 - improve TravisCI with Docker and test each compile flag, NNPACK, etc.
 - Switch to GTest (as was discussed) and integrate it to CMake format.
 - Add custom git hooks in order to have a consistent code format (open to other solutions). 

After that it will be okay for me.
 
The use of tags I think it's the correct way. We could have a single master branch for current users and keep updating in each big release. Meanwhile, the current development I would go for a `dev` branch (just a name to not confuse the users). From an outsider POV it seems a bit confusing the project infrastructure.

@nyanp Regarding further community bounding, I understand that it's your personal project, but what about to create an independent repository for tiny-nn or as dicussed tiny-nn, tiny-dnn or whatever evolution name? Another task is that we should put a LICENSE file and discuss its type. Thinking about a good logo could be nice too.

Wangyida:
I am OK with this, my current PR is more likely an external function for quantization with basic quantization utilities. I have finished core algorithms about this and the accuracy is guaranteed. Now I need to do optimization on data flow in network for a real quantization function. I think these details could also come with a new PR rather than my current one because it is related to the graph nodes design.

nyanp:
Thx! I'll merge it after completing works @edgarriba listed.

@Wangyida
Creating a new branch for further optimization seems a good idea for me. Please call me when you think #206 is ready to merge.

@edgarriba
Nice plan :)
I'll use a tag for each major/minor version-up, as I did in v0.1.0. I agree that we use graph(will be renamed to dev) branch for the current GSoC working. Do you think its better for everyone to use dev branch, even if we use tags for each release version of master? It might be a bit confusing for newcomers from other dnn frameworks, because other dnn frameworks directly use a master branch for their development.

bhack:
I vote for master with tag instead of dev.

edgarriba:
@nyanp @bhack whatever coherent option is okay for me. Master branch for development and tags for releases seems a good practice instead of having extra branches. For this reason I proposed tiny in a standalone repo since there your own personal feature branches. You are the project owner so as you wish, however working as "organization" I think that could be more productive since we could assign different roles such code revision (+1), tiquets revision, handle plugin projects or just like TF does with their [models zoo](https://github.com/tensorflow/models).

bhack:
I'am :+1: also for migrating to an organizzation account.

nyanp:
@edgarriba @bhack 
I started this project as my personal one, but now about 400 people folk (it's more than [chainer](https://github.com/pfnet/chainer)!) and more than 1k people star tiny-cnn - happily for me. It seems good to transfer this repository to an organization account for further development, and I've made [the new account](https://github.com/tiny-nn).
As far as I read [a github help](https://help.github.com/articles/about-repository-transfers/) our ongoing PRs and folks will be preserved and automatically redirected to the new url, so we can transfer tiny-cnn into the org account. Do you have any problem with transferring?

bhack:
:+1:

nyanp:
> @nyanp @bhack whatever coherent option is okay for me. Master branch for development and tags for releases seems a good practice instead of having extra branches. 

Sure, I'll use master branch with tagging, instead of dev branch :) Putting a LICENSE file is important too, I'll submit this as a new issue :+1:

edgarriba:
@nyanp awesome! Are we officially tiny-nn ?

bhack:
Probably we need an improved logo.. I try to ask to a designer if she can contribute one.

patrikhuber:
One thing to consider is the results that pop up if you google for "tiny-nn". Apparently, the abbreviation mainly stands for something else. In this case I am not sure if it could even cause problems with "safe search" and parents filters.
But then again, it is said computer scientists are not good at naming things ;-)

edgarriba:
more facts: https://github.com/uyu-kickstart/tiny-nn

nyanp:
The org/project name and a logo are both my drafts. I've noticed that it's result of google for "tiny nn" is bad enough to change the name (thanks for the important fact, @patrikhuber !). It would be very nice if a designer can contribute :)

edgarriba:
tiny-dnn ?

nyanp:
@edgarriba 
Good idea :) When I consider that we can construct non-deep architectures on this framework, tiny-dnn looks similar to tiny-cnn - keep its name is one idea.

edgarriba:
actually, it's funny to have a kind of opposite meanings in one xD

edgarriba:
@nyanp awesome!

I have a couple of questions around this refactorization: as I've seen you removed the worker indexes, so what's happening with the data buffers?  Do you solve with this the mini batch thing?

edgarriba:
@bhack How's the logo thing going?
@nyanp When do you plan to move the code to the organization account?

nyanp:
In the previous version of tiny-cnn, we used a mixed strategy of __data parallelism__ and __layer-wise parallelsim__ . We've splitted each mini-batch into multiple chunks (see [these lines](https://github.com/nyanp/tiny-cnn/blob/feat/generic-computational-graph/tiny_cnn/network.h#L670-L680)), and simuletanousely execute forward/backward pass (data parallelism). Moreover, we've used parallel-for loop for each layers (layer-wise parallelism).
In this PR, I removed data parallelism in tiny-cnn because batch-normalize operation requires to get all chunks of mini-batch data at once. This change slow down execution time ~20%, but this change makes codes much simpler and introducing BN and recent optimization efforts of you and @beru is much greater than this slight degradation.

nyanp:
@edgarriba 
I haven’t made up my mind which name is better:

- tiny-cnn
- hornet (loose acronyms of Header Only Network)

Which one do you like? I'll move to org account right after I decide new name. Of course we should wait to make a logo until the naming problem is solved :)

edgarriba:
@nyanp definitely removing data parallelism was needed.
It will be easier for the future to add new layers and backend :)

Regarding names, if we are staying in tiny line, I'd prefer `tiny-dnn`, however other pokemon style name like hornet sounds interesting. The only drawback for radical names changing could be a bit confusing for the current user. In case of hornet we could relate it with a similar [logo](https://s-media-cache-ak0.pinimg.com/236x/de/83/9b/de839b6228ac7cbca5b8b06685c98742.jpg).

Let's keep this discussion along this entire week and decide by the end of it.
Organization bases have to be robust :)

nyanp:
@edgarriba 
> It will be easier for the future to add new layers and backend :)

Yes, that is one of the biggest reason to remove the data parallelism. 

> Let's keep this discussion along this entire week and decide by the end of it.
> Organization bases have to be robust :)

Sure! This is off-topic from this PR, so I'll submit a new issue for the discussion.

nyanp:
@edgarriba 

> Switch to GTest (as was discussed) and integrate it to CMake format.
> Add custom git hooks in order to have a consistent code format (open to other solutions).

I think these tasks aren't conflict with merging ```master``` branch. Can I start to merge? I don't want to keep @Wangyida  waiting so long for the task completed...

edgarriba:
sure! I'll PR this features later.

Proceed with merging. Are you gonna merge everything in master as we discussed and use tags?

nyanp:
@egarriba
Thx :) Sure I'll do, are you worried about something?

edgarriba:
nope, it's fine! :)

PR #217, opened 2016-07-06
Generalization of loss functions + Correcting MSE 

H4kor:
Solves issue #159

PR #232, opened 2016-07-17
Generalization of loss functions + Correcting MSE

H4kor:
New pull request for #217 (changed branch)

Loss function compute `f` and `df` for vectors instead of floats.

### DIFF
--- a/tiny_cnn/lossfunctions/loss_function.h
+++ b/tiny_cnn/lossfunctions/loss_function.h
@@ -32,7 +32,7 @@ namespace tiny_cnn {
 // mean-squared-error loss function for regression
 class mse {
 public:
-    static float_t f(vec_t y, vec_t t) {
+    static float_t f(const vec_t& y, const vec_t& t) {
         assert(y.size() == t.size());
         float_t d = 0.0;
         for(unsigned int i = 0; i < y.size(); ++i)
@@ -40,15 +40,19 @@ public:
         return d/y.size();
     }
 
-    static float_t df(float_t y, float_t t) {
-        return y - t;
+    static vec_t df(const vec_t& y, const vec_t& t) {
+        assert(y.size() == t.size());
+        vec_t d; d.resize(t.size());
+        for(unsigned int i = 0; i < y.size(); ++i)
+            d[i] = y[i] - t[i];
+        return d;
     }
 };
 
 // cross-entropy loss function for (multiple independent) binary classifications
 class cross_entropy {
 public:
-    static float_t f(vec_t y, vec_t t) {
+    static float_t f(const vec_t& y, const vec_t& t) {
         assert(y.size() == t.size());
         float_t d = 0.0;
         for(unsigned int i = 0; i < y.size(); ++i)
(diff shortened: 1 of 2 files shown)


tiny_cnn/lossfunctions/loss_function.h
```diff
@@ -32,48 +32,67 @@ namespace tiny_cnn {
...
 public:
-    static float_t f(float_t y, float_t t) {
-        return (y - t) * (y - t) / 2;
+    static float_t f(const vec_t& y, const vec_t& t) {
+        assert(y.size() == t.size());
+        float_t d = 0.0;
+        for(unsigned int i = 0; i < y.size(); ++i)
+            d += (y[i] - t[i]) * (y[i] - t[i]);
+        return d/y.size();
     }
```
nyanp: @H4kor 
Could you tell me the reason why you change the API from ```float f(float,float)``` to ```float f(vec_t,vec_t)```, and ```float df(float,float)``` to ```vec_t df(vec_t,vec_t)```?
If we choose this API, using ```std::accumulate``` to avoid boilerplate for-loop seems better.

H4kor: The change from `float f(float,float)` to `float f(vec_t,vec_t)` was necessary to support loss functions computed over a vector instead of accumulating the element wise loss (i.e. MSE where the loss is divided by the number of elements).

I've changed the df functions in respond to your comment on #217. I'm not aware of functions requiring the whole vector to compute the derivative. This change might not be necessary but might be more convenient as it doesn't require a extra loop where the derivative is used.

`std::accumulate` is only suitable for operations on one vector, to my knowledge.

nyanp: @H4kor 
Thanks, it's ok to extend API and you are right regarding to ```std::accumulate```.

nyanp: In #217 I mean that we should keep a coherence between definitions of ```f``` and ```df```. If you change the definition of ```f``` to ```1/N * sum(y[i] - t[i])^2```,  ```df``` must be ```2/N * sum(y[i] - t[i])```.

H4kor: @nyanp Ok, misunderstood your comment. Tests are still failing. Don't know what exactly happens in `generate_gradient_check_data` and `gradient_check`

tiny_cnn/lossfunctions/loss_function.h
```diff
@@ -43,8 +43,9 @@ class mse {
     static vec_t df(const vec_t& y, const vec_t& t) {
         assert(y.size() == t.size());
         vec_t d; d.resize(t.size());
+        float_t factor = 2.f/static_cast<float_t>(t.size());
         for(unsigned int i = 0; i < y.size(); ++i)
-            d[i] = y[i] - t[i];
+            d[i] = factor*y[i] - t[i];
```
nyanp: @H4kor 
Probably what we need is ```d[i] = factor*(y[i] - t[i])```.

nyanp:
@H4kor 
Ok, everything is fine. Thank you very much!

PR #240, opened 2016-07-18
New loss functions: absolute and absolute with epsilon range

H4kor:


tiny_cnn/lossfunctions/loss_function.h
```diff
@@ -50,6 +50,72 @@ class mse {
...
+
+    static vec_t df(const vec_t& y, const vec_t& t) {
+        assert(y.size() == t.size());
+        vec_t d; d.resize(t.size());
+        float_t factor = 1/static_cast<float_t>(t.size());
+        for(unsigned int i = 0; i < y.size(); ++i)
+        {
+            float_t sign = y[i] - t[i];
+            if(sign < 0.f)
+                d[i] = -1.f * factor;
```
edgarriba: use float_t(0)

nyanp:
@H4kor  /cc @edgarriba 
Thanks for nice contribution! :+1: 
Using ```float_t(0)``` may be better but it's working. Since we have a lot of compiler warnings in other places, I'll modify it as well :)

PR #246, opened 2016-07-23
Fix AVX compilation error on gcc/clang

nyanp:


CMakeLists.txt
```diff
@@ -163,12 +163,12 @@ if(CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
     # set Advanced Vector Extensions (AVX)
     if(USE_AVX AND COMPILER_HAS_AVX_FLAG)
     	add_definitions(-DCNN_USE_AVX)
-        set(EXTRA_C_FLAGS "${EXTRA_C_FLAGS} -mavx")
+        set(EXTRA_C_FLAGS "${EXTRA_C_FLAGS} -mavx -mfma")
```
beru: @nyanp 

FMA3 instruction set are not supported by Intel Sandy Bridge cpus and Ivy Bridge cpus (though I read that it's supported by some of AMD nondominant processors...), so always coupling -mavx compiler flag with -mfma compiler flag may cause troubles.

I still haven't used the intrinsics for FMA3 instruction set in AVX backend code at ( https://github.com/nyanp/tiny-cnn/blob/feat/generic-computational-graph/tiny_cnn/core/kernels/avx_kernel_common.h ) yet. In the AVX backend, I've tried to use only AVX intrinsic functions and avoided AVX2 and FMA3 intrinsic functions. 

I thought I could achieve noticeable performance gains with AVX2/FMA3 intrinsinc functions but I have failed so far.

It would be better if the backend checks the availabilities of instructions with cpuid on x86/x64 and can automatically select the most performant routines. Maybe in the distant future.

nyanp:
@beru 
It seems that you use FMA intrinsics [here](https://github.com/nyanp/tiny-cnn/blob/feat/generic-computational-graph/tiny_cnn/core/kernels/avx_conv2d_kernel.h#L389). If we don't use mfma, gcc complains that we can't use _mm_fmadd_sd instrinsics. 
Should we replace this line with mul+add and remove mfma option? Automatic selection seems great, [this sample](https://msdn.microsoft.com/ja-jp/library/hskdteyh.aspx) would be helpful.

beru:
@nyanp It was my mistake that I used _mm_fmadd_sd intrinsinc there. Sorry about that. Actually, I haven't focused on optimizing double floating point code... I'm not so sure about the relationship between error rate and numerical accuracy of computation. I need to research and experiment.

PR #249, opened 2016-07-25
bugfix for deconv and unpool

Wangyida:
1 segmentation and 1 corner case bug in unpool and deconv bugs are solved.

bugs in forward codes for padding::same of deconvlolution and deconvolution is solved.
bugs in unpoolign layer is solved in the bias term.

### DIFF
--- a/tiny_cnn/core/kernels/tiny_quantized_conv2d_kernel.h
+++ b/tiny_cnn/core/kernels/tiny_quantized_conv2d_kernel.h
@@ -374,13 +374,13 @@ void tiny_quantized_conv2d_kernel(const conv_params& params,
         &max_output_value);
     // data type restore
     std::vector<uint8_t> in_quantized, W_quantized, bias_quantized;
-    for (int i = 0; i < in.size(); i++) {
+    for (uint i = 0; i < in.size(); i++) {
        in_quantized.push_back(static_cast<uint8_t>(in[i]));
     }
-    for (int i = 0; i < W.size(); i++) {
+    for (uint i = 0; i < W.size(); i++) {
         W_quantized.push_back(static_cast<uint8_t>(W[i]));
     }
-    for (int i = 0; i < bias.size(); i++) {
+    for (uint i = 0; i < bias.size(); i++) {
         bias_quantized.push_back(static_cast<uint8_t>(bias[i]));
     }
 
@@ -446,7 +446,7 @@ void tiny_quantized_conv2d_kernel(const conv_params& params,
     quantize_down_and_shrink_range<int32_t, uint8_t>(a_quantized, min_output_value, max_output_value,
         &min_output_requantized, &max_output_requantized, &a_requantized);
     // store directly in float datatype
-    for (int i = 0; i < a_requantized.size(); i++) {
+    for (uint i = 0; i < a_requantized.size(); i++) {
         a[i] = static_cast<float>(a_requantized[i]);
     }
     a_r[0] = min_output_requantized;
--- a/tiny_cnn/core/kernels/tiny_quantized_conv2d_kernel.h
+++ b/tiny_cnn/core/kernels/tiny_quantized_conv2d_kernel.h
@@ -100,14 +100,6 @@ void tiny_quantized_conv2d_kernel(const conv_params& params,
     const int32_t zero_in_total_space =
         float_to_quantized<int32_t>(0.0f, min_output_value, max_output_value);
 
-    const int32_t offset_output = 0;
-    const int32_t mult_output = 1;
-    const int32_t shift_output = 0;
-
-    const int32_t rounding = (shift_output < 1) ? 0 : (1 << (shift_output - 1));
-    const int32_t highest_ = static_cast<int32_t>(highest<uint8_t>());
-    const int32_t lowest_ = static_cast<int32_t>(lowest<uint8_t>());
-
     for_i(layer_parallelize, params.out.depth_, [&](int o) {
         for (cnn_size_t inc = 0; inc < params.in.depth_; inc++) {
             if (!params.tbl.is_connected(o, inc)) continue;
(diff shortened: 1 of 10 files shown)


Wangyida:
@nyanp  This is a minor bug fix for all my previous forward pass codes for new codes structure. There are still bugs in backward codes in deconv, unpool and quantized layers because all gradient checking in deconv and unpool failed as previous codes won't. I think they might be also related to new codes structure, I'll try to find them out.

edgarriba:
@Wangyida Could you revise for this PR the warnings in quantization module?
http://pastebin.com/1dWb6ibW

Wangyida:
@edgarriba My local server blocks [pastebin](http://pastebin.com/), could you please provide another copy of script such as [pastebin.ca](http://pastebin.ca/)? Thanks a lot.

edgarriba:
@Wangyida http://pastebin.ca/3665550

Wangyida:
@edgarriba Thanks, I can see it now.

nyanp:
@Wangyida 
Quick work, nice ;) As @edgarriba said it's better to cleanup warnings in this PR.

Wangyida:
Agree, I think current efficient quantization without unnecessary quan/dequan is also not proper, this should be a backend based solution instead.

nyanp:
@Wangyida 
Do you have any progress on fixing remaining bugs? Is there anything I can help?

Wangyida:
I solved some warning in quantization, I will push again after the majority is solved. This isn't hard work, Thanks. 
@nyanp @bhack I had a try on based on a TensorFlow contribution which is based on makefile building process and place it under thrid_party folder for TF implementation [TensorFlow-minimal](https://github.com/Wangyida/tiny-cnn/tree/tf/thrid_party/tensorflow). These codes solve the problem of dependency on bazel build system carried out by Google by using makefile.
This is also an implementation based on Session concept which is not so ideal for serialization between TF and tiny-cnn, but this might be interesting though.

nyanp:
@Wangyida 
Great, I'll wait for fixing other tests ;) I want to merge graph into master when all CI tests are passed.

Will you include TF binaries into tiny-cnn? Is there problem to parse protobuf (documented [here](https://www.tensorflow.org/versions/r0.10/how_tos/tool_developers/index.html)) and convert to tiny-cnn layers?

Wangyida:
@nyanp No I won't include TF binaries into tiny-cnn, this makes tiny-cnn not so tiny as a standalone toolbox, Here in my [tf branch](https://github.com/Wangyida/tiny-cnn/tree/tf/thrid_party/tensorflow) give an example for automatic finding core source files in TF and compile a standalone command line tool for parsering weights and graph. So even if TF is updated in proto definition, we can still generating tools automatically without rewritting a new parser API.

bhack:
What is the target of conversione? A protobuff version of tinycnn graph and weights? The goal of this last week is to have Enet  trained with TF and converted  in a tinycnn protobuff serialized format (a quantized one?). And then inference with tinycnn gemmlowp kenels (the are updated, we need to find a way to stay in sync with upstream) on mobile.

bhack:
There is also need of an PR against OpenCV to demonstrate integration. We have less then 3 weeks to the end of GSoC.

Wangyida:
@bhack As for C++ API for transforming TF trained weights and graph, Protobuf 3.0 based parser after compiled with protoc with [PR 254](https://github.com/nyanp/tiny-cnn/pull/254) is enough. This is different from [minimal-TF branch] (https://github.com/Wangyida/tiny-cnn/tree/tf) which is aimed at auto-embedding tensorflow.

bhack:
Yes but I'm talking about to not always handling the coversion each time. Are you think on an tinycnn own protobuff?

Wangyida:
Em I see, ENet, but the quantized ENet is not tried in python API. For tinycnn's own protobuf implementation, I haven't got a good idea yet. Maybe a registration structure should be 2 level (kernel Message under Layer Message) if weights of kernels from particular layers are stored similar to TF graph.pb.

bhack:
We are really shorting our Gsoc time.

Wangyida:
@bhack Got it, the trained graph could be loaded in the updated TF parser PR, so it seems that the tensorflow minimal-embed branch is not so useful mainly caused by introduce unnecessary methods. On this point, applying trained graph from TF using this [parser](https://github.com/nyanp/tiny-cnn/pull/254) is the key task.

Wangyida:
@nyanp I can see that there are problem in back propagation kernels of both deconvolution and quantized deconvolution. The forwards pass codes are modified well with new codes structure, and the backward codes fails my previous test for both layers. Is there good method to solve such bugs? Because I find no different between linear algebra operation between the [latest codes](https://github.com/Wangyida/tiny-cnn) and my [quantization branch](https://github.com/Wangyida/tiny-cnn/tree/quantization).

nyanp:
@Wangyida 
No silver bullet for such bugs... if all tests are passed in your branch, feed the same vector into deconv layers of latest code and your branch, and check the difference of those outputs (especially  we should focus on "prev_delta" and "dW" in bprop, because only gradient check is failed) by your debugger. Comment out all other tests and create minimal test case that can cause the difference.

Wangyida:
@nyanp  Good advice, I am monitoring the expected delta by bp and the approximate delta.

tiny_cnn/core/kernels/tiny_quantized_conv2d_kernel.h
```diff
@@ -388,13 +374,13 @@ void tiny_quantized_conv2d_kernel(const conv_params& params,
...
-    for (int i = 0; i < in.size(); i++) {
+    for (uint i = 0; i < in.size(); i++) {
        in_quantized.push_back(static_cast<uint8_t>(in[i]));
     }
-    for (int i = 0; i < W.size(); i++) {
+    for (uint i = 0; i < W.size(); i++) {
         W_quantized.push_back(static_cast<uint8_t>(W[i]));
     }
-    for (int i = 0; i < bias.size(); i++) {
+    for (uint i = 0; i < bias.size(); i++) {
```
nyanp: @Wangyida 
```uint``` is non-standard name. Use ```size_t``` instead.

nyanp:
@Wangyida 
Test of deconv/unpool is still broken on my local environment... :( Any idea?

tiny_cnn/core/kernels/tiny_quantized_conv2d_kernel.h
Wangyida: @nyanp OK, fixed

Wangyida:
@nyanp These failed back propagation tests are previously written, seems problem lies in parameter passing like pointer and reference misusing in new architecture as I look into the gradient check function where the computed gradient is in abnormal range.

PR #258, opened 2016-08-03
add base OpKernel and Conv2dKernel skeletons

edgarriba:
Quick PR to introduce OpKernel and Conv2dCustomOp

### DIFF
--- a/tiny_cnn/core/framework/op_kernel.h
+++ b/tiny_cnn/core/framework/op_kernel.h
@@ -1,0 +1,90 @@
+/*
+    COPYRIGHT
+
+    All contributions by Taiga Nomi
+    Copyright (c) 2013, Taiga Nomi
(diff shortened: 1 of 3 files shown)


Wangyida:
There improper implementation on efficient quantized kernels with additional range representation [conv2d_eq etc.](https://github.com/nyanp/tiny-cnn/blob/feat/generic-computational-graph/tiny_cnn/core/backend_tiny.h). These 3 method could be reconsidered in data flow.

edgarriba:
@nyanp @Wangyida If we all agree with this design, can we go ahead merging to not block other tasks?

Wangyida:
@edgarriba Seems OK for me as I'm on serialise branch now.

nyanp:
@edgarriba 
Sure! It's ok for me ;)

PR #260, opened 2016-08-03
remove get_* from getters

edgarriba:


nyanp:
@edgarriba 
I think it's a good suggestion :) Since Layer APIs are public, it would be better we keep old names but marked them as deprecated. (See [caffe's deprecated API](https://github.com/BVLC/caffe/blob/master/include/caffe/blob.hpp#L131-L138) as an example)

patrikhuber:
Or just increase the major version, to comply with semver ;-) I think it's not worth leaving "old cruft" hanging around at this stage of the library.

In any case I'm curious about why you two like it without the `get_*`. I'm a bit undecided of which one I find better, but I'm leaning towards using `get_*`, so I'm very curious to hear some arguments about it.

nyanp:
@patrikhuber 
Thanks for your advice ;) Indeed many user may not expect this library to such a strict backward compatibility.

> In any case I'm curious about why you both like it without the ```get_*```. 

It's the matter of consistency for me. Currently some getter method in ```layer``` have the prefix ```get_*``` and others don't have. Either style is fine for me, but I don't want to mix them.

edgarriba:
yes, it's a matter of consistency and from my point of view it's more clean omitting get.

Issue #122, opened 2016-04-07
How to recognize the images using the weights of cifar-10?


18569035534:
I have trained the cifar-10 dataset and get a network,but how to recognize the images ?

pansk:
Once you load your network, weights and load the image into a vec_t variable, you can use the predict_label method on your network object to receive the class, or you can use the predict method on the same object, receiving the output vector, and find the index containing the maximum value yourself (the former is slightly more efficient).

18569035534:
Thank you ! After your advice, the program has been run successfully, but the accuracy is a bit low.
And how to improve the accuracy of recognition?

stereomatchingkiss:
Are you using the codes from the example?If so, you could try

1 : augment your training data, like horizontal flip, vertical flip, pertube image color, add noise etc, figure out which one works for you
2 : use deeper network, you could reference the architecture of vggnet, googlenet and so on
3 : study videos of cs231n, these videos will help you a lot
4 : use pretrained model of caffe to extract features

Last thing, please share your training results if you are able to gain high accuracy, thanks

edgarriba:
@18569035534 Solved that?

Randl:
@18569035534 Check the now existing example of CIFAR-10 if you're still interested. Feel free to ask questions, if you have.

PR #263, opened 2016-08-05
add cifar10's test. (isse #122)

MasazI:
I added cifar10's test code to recognize own image. (isse #122)


nyanp:
@MasazI 
Great!! Thank you very much :)

PR #266, opened 2016-08-08
Refactor core ops

edgarriba:


tiny_cnn/core/kernels/conv2d_tiny_op.h
```diff
@@ -44,36 +44,152 @@
...
-        // const vec_t&          b = context->input(2);
-        // tensor_t&      out_data = context->output(0);
-
+        // incomimg/outcoming data 
+        const tensor_t& in_data = context->input(0);
+        const vec_t&          W = context->input(1)[0];
+        const vec_t&       bias = context->input(2)[0];
+        tensor_t&      out_data = context->output(1);
+
+        // pad input data
```
nyanp: @edgarriba 
Outputs aren't initialized in functor, so we need to insert ```fill_tensor(out_data, 0);``` here. It will fix the test :)

tiny_cnn/core/kernels/conv2d_tiny_op.h
```diff
@@ -44,36 +44,152 @@
...
+        tensor_t in_data_padded;
+        Conv2d::setParams(context->params());
+        Conv2d::copy_and_pad_input(in_data, in_data_padded);
+
+        // convolution algorithm
+        Conv2dCustomForwardFunctor()(in_data_padded, W, bias, out_data,
+                                     Conv2d::params(), context->parallelize());
```
nyanp: We can use normal function instead of functor. Do we need to treat this as functor?

tiny_cnn/core/params/conv_params.h
```diff
@@ -90,6 +95,18 @@ struct conv_params {
...
+    friend std::ostream& operator<<(std::ostream &o,
+                                    const core::conv_params& param) {
+        o << "in:        " << param.in        << "\n";
+        o << "out:       " << param.out       << "\n";
+        o << "in_padded: " << param.in_padded << "\n";
+        o << "weight:    " << param.weight    << "\n";
+        o << "has_bias:  " << param.has_bias  << "\n";
+        o << "w_stride:  " << param.w_stride  << "\n";
+        o << "h_stride:  " << param.h_stride  << "\n";
+        return o;
```
nyanp: This function is ok. We need to return ostream at backend.h#L52, as with this.

tiny_cnn/layers/convolutional_layer.h
```diff
@@ -233,12 +238,26 @@ class convolutional_layer : public feedforward_layer<Activation> {
      * @param out_data     output vectors
      **/
     void forward_propagation(const std::vector<tensor_t*>& in_data,
-                             std::vector<tensor_t*>&       out_data) {
+                             std::vector<tensor_t*>&       out_data) { 
+        // forward convolutional op context
+        auto ctx = new OpKernelContext(in_data, out_data);
+             ctx->setParams(&params_);
+             ctx->setParallelize(layer::parallelize());
+
```
nyanp: Memory leak occurs on this line. If ```ctx``` can be destructed at the end of this scope, we can simply use stack.
```
OpKernelContext ctx(in_data, out_data);
```

tiny_cnn/layers/convolutional_layer.h
```diff
@@ -252,16 +271,28 @@ class convolutional_layer : public feedforward_layer<Activation> {
...
-        Base::backend_->conv2d(in_data, out_data, out_grad, in_grad);
+        // activations
+        // TODO(edgar/nyanp): refactor and move activations out
+        this->backward_activation(*out_grad[0], *out_data[0], *out_grad[1]);
+
+        // forward convolutional op context
+        auto ctx = new OpKernelContext(in_data, out_data, out_grad, in_grad);
+             ctx->setParams(&params_);
+             ctx->setParallelize(layer::parallelize());
+
```
nyanp: memory leak.

tiny_cnn/layers/convolutional_layer.h
```diff
@@ -445,7 +476,16 @@ class convolutional_layer : public feedforward_layer<Activation> {
     }
 
     void init_backend(const backend_t backend_type) {
-        std::shared_ptr<core::backend> backend = nullptr;
+        // TODO(edgar): add device?
+        auto ctx = new core::OpKernelConstruction(nullptr);
+
```
nyanp: memory leak.

tiny_cnn/core/kernels/conv2d_tiny_op.h
```diff
@@ -44,36 +44,152 @@
...
+#include "tiny_cnn/core/kernels/conv2d.h"
+#include "tiny_cnn/core/kernels/conv2d_tiny_op_functor.h"
+
 #include "tiny_cnn/core/framework/op_kernel.h"
 
 namespace tiny_cnn {
 
-class Conv2dCustomForwardOp : public core::OpKernel {
+class Conv2dCustomForwardOp : public core::OpKernel, Conv2d {
  public:
```
nyanp: ```Conv2d``` has no polymorphic functions. Let's use private inheritance or containment instead.

tiny_cnn/core/kernels/conv2d_tiny_op.h
edgarriba: do you think we could initialize outputs during allocation? I think it will prevent that others will face the same issue.

tiny_cnn/core/kernels/conv2d_tiny_op.h
edgarriba: well in this case we don't really need to hold the function state but I think that having it, fits more in OOP style. I know it's not a strong argument but is there any performance consideration in using functors over normal functions?

nyanp: Well C++ is not an OOP language, it's a multi-paradigm language. We can choose the simplest way to handle the problem. If we'll have to hold the state, we can hold them in ```conv2d_tiny_op``` class. (And actually I don't agree that adding ```XXXFunctor``` for handling the state is good OOP style...)
Performance would be differ if we use them as predicate. In this case there is no difference between functor and normal function :)

tiny_cnn/core/kernels/conv2d_tiny_op.h
nyanp: @edgarriba 
Since allocated memory will be reused through multiple data, we should re-initialize per forward op.

tiny_cnn/core/kernels/conv2d_tiny_op.h
edgarriba: @nyanp 
another benefit I see is that we could simplify the batch loop with `std::foreach`.
I've been checking some info about functors and they claim that in some cases it's better than normal functions for inline.
http://stackoverflow.com/questions/6451866/why-use-functors-over-functions

nyanp: @edgarriba 
I agree that some cases performance differs between functor and normal function, but this case isn't (we don't use functor/function as an argument, so compiler can generate the same binary). Batch loop is executed inside the functor, so we don't need to use ```std::foreach```. 
In this case, normal function is simpler choice.

nyanp: And when we want to use functor, we can consider using lambda (it's more c++11-way). Lambda is syntax sugar of functor. http://stackoverflow.com/questions/7627098/what-is-a-lambda-expression-in-c11

PR #272, opened 2016-08-11
Refactor core ops

edgarriba:
same as https://github.com/tiny-dnn/tiny-dnn/pull/266

- Partial migration of old backend concept to pure Ops style.
- Disabled deconvolution tests since they are broken: @Wangyida enable when you fix them
- Fixed warning


edgarriba:
@nyanp I opened this PR since the previous branch falled into a detached state.
Tests and warning are fixed (not deconv).

nyanp:
@edgarriba 
Great :) How do you think about [this comment](https://github.com/tiny-dnn/tiny-dnn/pull/266#discussion_r74674811)?

edgarriba:
hi! Sorry for late reply. Yes, so let's get rid of functors and use pure normal functions. Regarding batch loop, I'm just trying to find a solution in order to place it at layer level but in case we want run networks with pure Ops not sure if it's the proper design.

edgarriba:
@nyanp if you agree I'll merge

nyanp:
@edgarriba
Ok, plz merge it :)

PR #294, opened 2016-08-24
Allocate storage for temporary buffer.

syoyo:
Memory is not allocated for `buf[sample]`, thus it causes segmentation fault. This patch fix it.

### DIFF
--- a/tiny_dnn/core/kernels/conv2d.h
+++ b/tiny_dnn/core/kernels/conv2d.h
@@ -75,6 +75,8 @@ class Conv2d {
                 out[sample] = in[sample];
             }
             else {
+                // alloc temporary buffer.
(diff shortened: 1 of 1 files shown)


edgarriba:
@syoyo 
thx to fix the bug!

It could be interesting to have a coupe of tests for this `core::padding::same ` case.
Could you also check if same happens to `copy_and_unpad_delta` ?

/cc @nyanp

nyanp:
@syoyo 
Thank you very much, great! I'll merge it at once.

@edgarriba 
> Could you also check if same happens to copy_and_unpad_delta ?

Probably no. But it's a good idea to add tests for this case to check our new backends :)
I think we should merge this PR as soon as possible, before further checking.

syoyo:
Thanks!

> Could you also check if same happens to copy_and_unpad_delta ?

I'll investigate other code part including `copy_and_unpad_delta` when I have spare time.

Issue #245, opened 2016-07-22
when i debug examples\main.cpp ,it will 0x002AA9DC 处有未经处理的异常(在 tiny_cnn.exe 中):  0xC000001D: Illegal Instruction。


angelanf5:
my vs is VS2013 express for windows desktop.
tiny_cnn is tiny-cnn-master  download at 19th july.
when i compile it, it will warning this:
![up](https://cloud.githubusercontent.com/assets/17898886/17062573/04326b74-5066-11e6-8be9-bae4338df285.JPG)

waiting for help :)

xylcbd:
your application heap is crashed,may be caused by heap overflow.

nyanp:
@angelanf5 
Whats kind of CPU do you use? If your CPU doesn't support AVX instructions, please remove ```CNN_USE_AVX``` definition from project property (project -> property -> c/c++ -> Preprocessor)

angelanf5:
thanks u for your help,nyanp.  it runs normally now.

angelanf5:
@nyanp  my cpu is AMD Athlon(tm) II X2 215 Processor. when i remove CNN_USE_AVX definition from project property, it can run normally.
thank u.

PR #296, opened 2016-08-26
fixed bugs in caffe converter example

Jiaolong:
The current caffe converter example is not fully tested. Mainly fixed bugs in:
- CMakefiles: enable compile caffe_converter example
- caffe_converter.cpp: updated the code with latest tiny-dnn

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -44,7 +44,7 @@ target_link_libraries(example_cifar_train
 
 find_package(Protobuf QUIET)
 if(PROTOBUF_FOUND)
-    set(proto_file "${CMAKE_CURRENT_SOURCE_DIR}/tiny_dnn/io/caffe/caffe.pb.cc")
+    set(proto_file "${CMAKE_SOURCE_DIR}/tiny_dnn/io/caffe/caffe.pb.cc")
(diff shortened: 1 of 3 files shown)


edgarriba:
@Jiaolong 
thx for the contribution !

Jiaolong:
@edgarriba
I just tested it with caffe bvlc reference model, however, an error occured when running in release mode. It shows `Illegal instruction     (core dumped)`.

In debug mode, everything is fine. I tried to locat the error and it turns out that the crash came from this line:
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/random.h#L68

I haven't solved this problem. Could you please help me on this? I have compared the cmake c++ flags of release and debug, but got no hint.

nyanp:
@Jiaolong 
Removing ```CNN_USE_AVX``` flag would solve the problem if you use AMD processor. See:  https://github.com/tiny-dnn/tiny-dnn/issues/245

Jiaolong:
Thanks a lot! Setting USE_AVX=OFF, it works perfectly! I use Intel CPU though, `Intel(R) Core(TM) i5-2410M CPU @ 2.30GHz`.

PR #297, opened 2016-08-26
Suppress compiler warnings on `override` keyword.

syoyo:
This simple patch suppress missing override keyword warning, e.g.

```
In file included from train.cpp:29:
In file included from ../../tiny_dnn/tiny_dnn.h:45:
../../tiny_dnn/layers/max_pooling_layer.h:138:10: warning: 'back_propagation' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    void back_propagation(const std::vector<tensor_t*>& in_data,
         ^
```


edgarriba:
@syoyo 
thx for your contribution :+1:

PR #298, opened 2016-08-28
fix padding::same issue

edgarriba:


edgarriba:
@nyanp 
Cannot find the error. This is the log with valgrind: http://pastebin.com/wJmLa9Fc

nyanp:
@edgarriba 
By [dr.memory](http://www.drmemory.org/) (windows equivalent to valgrind) and IDE. I've noticed that [current usage of unpad](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_grad_op.h#L114) is incorrect, because we are copying data into its gradient.
We need to allocate temporary buffer which have enough room for padded operation, and then we need to shrink it to original ```prev_delta```. Please check implementation of the [latest release version](https://github.com/tiny-dnn/tiny-dnn/blob/v0.1.1/tiny_cnn/layers/convolutional_layer.h#L359).

edgarriba:
@nyanp 
I haven't tried upadding yet, but with `copy_and_pad_input` there's also a memory error.

tiny_dnn/core/kernels/conv2d_grad_op.h
```diff
@@ -72,8 +72,15 @@ class Conv2dGradOp : private Conv2d, public core::OpKernel {
         // initalize outputs
         fill_tensor(prev_delta, float_t(0));
         
+        // in case we want zero-pading we need to
+        // reshape the data container
+
+        if (Conv2d::params().pad_type == core::padding::same) {
+            Conv2d::copy_and_pad_input(prev_delta, prev_delta);
+        }
+
```
edgarriba: @nyanp I added this fix in order to resize data container, but not sure if is okay since not pass tests

tiny_dnn/core/kernels/conv2d_grad_op.h
```diff
@@ -106,7 +113,7 @@ class Conv2dGradOp : private Conv2d, public core::OpKernel {
 
         // apply unpadding
         if (Conv2d::params().pad_type == core::padding::same) {
-            Conv2d::copy_and_unpad_delta(prev_out, prev_delta);
+            Conv2d::copy_and_unpad_delta(prev_delta, prev_delta);
         }
```
edgarriba: @nyanp I think this is like the original version

nyanp:
@edgarriba 
In forward-pass, ```in_data_padded``` is [allocated inside op](https://github.com/edgarriba/tiny-cnn/blob/0764a8340f94cffdabdc2dbdf3b3e0d48166cef6/tiny_dnn/core/kernels/conv2d_op.h#L69), and destructed for each loop. We need to reuse this in backward-pass when we calculate dW, because dot-product requires padding (see [backprop  of the latest release](https://github.com/tiny-dnn/tiny-dnn/blob/v0.1.1/tiny_cnn/layers/convolutional_layer.h#L292)).
One idea to solve this is moving padding/unpadding operation to convolutional-layer.

p.s.
It's hard to detect this kind of error when we access vector element via raw pointer. Using ```assert``` or ```vector::operator[]``` would be helpful (remember you should build tiny-dnn with debug mode when you want to enable them)

edgarriba:
@nyanp check last commit. I've added the buffers again to the conv layer but it doesn't solve the issue. At some point the padding routine crashes due to a corrupted double-linked list error.

tiny_dnn/layers/convolutional_layer.h
```diff
@@ -447,12 +467,97 @@ class convolutional_layer : public feedforward_layer<Activation> {
...
 
     /* Forward and backward ops */
     std::shared_ptr<core::OpKernel> kernel_fwd_;
     std::shared_ptr<core::OpKernel> kernel_back_;
+
+    /* Buffer to store padded data */
+    struct conv_layer_worker_specific_storage {
+        tensor_t prev_out_padded_;
+        tensor_t prev_delta_padded_;
+    } cws_;
```
nyanp: @edgarriba 
We need to edit move constructor (L218) to move cws_ correctly.

tiny_dnn/layers/convolutional_layer.h
```diff
@@ -369,6 +384,11 @@ class convolutional_layer : public feedforward_layer<Activation> {
         params_.w_stride = w_stride;
         params_.h_stride = h_stride;
         params_.tbl      = tbl;
+
+        // init padding buffer
+        if (params_.pad_type == padding::same) {
+			cws_.prev_delta_padded_.resize(1, vec_t(params_.in_padded.size(), float_t(0)));
```
nyanp: @edgarriba 
We need NxKx(W-padded)x(H-padded) storage for ```prev_delta_padded_``` but this line reserve 1xKx(W-padded)x(H-padded). We can override ```set_sample_count``` and reserve appropriate buffer there. see example in [max pooling](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/layers/max_pooling_layer.h#L162_L165).

tiny_dnn/layers/convolutional_layer.h
```diff
@@ -231,7 +231,12 @@ class convolutional_layer : public feedforward_layer<Activation> {
      * @param out_data     output vectors
      **/
     void forward_propagation(const std::vector<tensor_t*>& in_data,
-                             std::vector<tensor_t*>&       out_data) override {
+                             std::vector<tensor_t*>&       out_data) override { 
+        std::cout << "BEFORE PADDED" << std::endl;
+        copy_and_pad_input(*in_data[0], cws_.prev_out_padded_);
+        *in_data[0] = cws_.prev_out_padded_;
```
nyanp: @edgarriba 
```in_data``` and ```out_data``` is shared among layers. Changing these size of vectors will affect neighbor layers which depends on their size. We should encapsulate the effect of padding/unpadding inside convolution layer.

File: tiny_dnn/layers/convolutional_layer.h (hunk 3 above)
Comment (edgarriba) [inline tiny_dnn/layers/convolutional_layer.h]: @nyanp this seems to fix the corrupted list issue

File: tiny_dnn/layers/convolutional_layer.h [hunk 6]
```diff
@@ -232,13 +233,16 @@ class convolutional_layer : public feedforward_layer<Activation> {
...
-        std::cout << "BEFORE PADDED" << std::endl;
         copy_and_pad_input(*in_data[0], cws_.prev_out_padded_);
-        *in_data[0] = cws_.prev_out_padded_;
-        std::cout << "AFTER PADDED" << std::endl;
+
+        std::vector<tensor_t*> in_data_;
+        in_data_.push_back(&cws_.prev_out_padded_);
+        for (cnn_size_t i = 1; i < in_data.size(); ++i) {
+            in_data_.push_back(in_data[i]);
+        }
```
edgarriba: @nyanp I've tried this workaround solution but I think we are doing sth wrong since all tests for conv are broken in this way :disappointed:

nyanp:
@edgarriba 
No worries :smile: Maybe you missed  ```Conv2d::setParams(OpKernel::params_);``` in ```conv2d_op```.

tiny_dnn/layers/convolutional_layer.h
```diff
@@ -231,9 +232,17 @@ class convolutional_layer : public feedforward_layer<Activation> {
...
-                             std::vector<tensor_t*>&       out_data) override {
+                             std::vector<tensor_t*>&       out_data) override { 
+        copy_and_pad_input(*in_data[0], cws_.prev_out_padded_);
+
+        std::vector<tensor_t*> in_data_;
+        in_data_.push_back(&cws_.prev_out_padded_);
+        for (cnn_size_t i = 1; i < in_data.size(); ++i) {
+            in_data_.push_back(in_data[i]);
+        }
+
```
edgarriba: @nyanp
convolution tests are fixed with https://github.com/tiny-dnn/tiny-dnn/pull/298/commits/ca862cef3eb49a3487f743ea79cb0e864636d603. 
However, network tests which have convolution operations are broken. So, I suspect that the codes above are not a proper solution to share data between neighbor layers.

nyanp:
@edgarriba 
Sorry for late, we need to remove assignment at [convolutional_layer.h#L534](https://github.com/edgarriba/tiny-cnn/blob/ca862cef3eb49a3487f743ea79cb0e864636d603/tiny_dnn/layers/convolutional_layer.h#L534), because we don't use padded buffer in ```padding::valid``` mode. This is probably the last one bug in conv-layer.

PR #301, opened 2016-09-02
fixed bugs in concat_layer; added test_concat_layer

Jiaolong:
The current version of concat_layer does not work correctly and is lacks of testing cases.

There is another issue in the `tiny_dnn_test`. After passing data through slice_layer or concat_layer, the value of the data will be changed, and the testing will be failed. It reported the testing error like:
`actual_out2[0] == in[0] ^ in[1] failed for: 0.877698 == 1`

I think this issue should be address in another pull request.


nyanp:
@Jiaolong 
Thanks a lot, nice :+1: I'll check failed tests later.

PR #305, opened 2016-09-14
Fix/duplicate symbols 2

rolandpersson:
The duplicate symbols test originally added in #171 has become disabled (not sure why, couldn't find the commit that removed it in the history).

This re-enables the test and also fixes a number of new duplicate symbols. There may still be others in configurations that I'm not building but the test should find them hopefully.

### DIFF
--- a/tiny_dnn/core/framework/device.fwd.h
+++ b/tiny_dnn/core/framework/device.fwd.h
@@ -75,7 +75,7 @@ class Device {
      *
      * @param type The device type. Can be only CPU.
      */
-    explicit Device(device_t type);
+    inline explicit Device(device_t type);
 
     /* CPU/GPU OpenCL constructor.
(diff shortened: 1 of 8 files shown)


tiny_dnn/core/framework/device.fwd.h
```diff
@@ -84,7 +84,7 @@ class Device {
      * @param platform_id The platform identification number.
      * @param device_id The device identification number.
      */
-    explicit Device(device_t type,
+    inline explicit Device(device_t type,
                     const int platform_id,
```
edgarriba: @rolandpersson preserve parameter alignment

edgarriba:
@rolandpersson thx to find this issue! 
As you probably noticed we are in the middle of a huge refactor and about to release alpha v.1.0. If you consider to fix more things now is the moment! :smile:

PR #308, opened 2016-09-15
Use logo with Alpha

bhack:


bhack:
Seems to me that also https://github.com/tiny-dnn/tiny-dnn/blob/master/docs/logo/TinyDNN-logo-letters.png has an alpha channel. But it is not interpreted correctly on the github mobile view.

bhack:
Nothing. Also tthis image on github mobile view doesn't render alpha.

PR #319, opened 2016-09-24
add Tensor class

edgarriba:


edgarriba:
@nyanp @bhack first basic tensor class prototype. Any consideration or feature to consider?
Next I'm gonna implement CPU/GPU data transfer

nyanp:
@edgarriba 
Nice, I'll check it.
BTW, do you think the ```Tensor``` class should be used only for implementation detail? or will you replace all public interfaces with this class? We need to have a policy to using this class.

edgarriba:
By now, I was thinking only for internal usage. However, could be interesting to have public interfaces in order to be able to manipulate tensors with further maths operations and maybe plug auto differentiation. This last will require more maintenance though. Let's discuss more.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,156 @@
...
+    /* Vector with the size of the tensor
+     * shape_[0]: batch
+     * shape_[1]: width
+     * shape_[2]: height
+     * shape_[3]: depth
+     */
+    std::vector<cnn_size_t> shape_;
+
+    /* Pointer to the Tensor data in CPU mode */
+    std::shared_ptr<std::vector<float_t> > data_cpu_;
```
nyanp: @edgarriba 
Should we need to use ```shared_ptr``` here? If we don't provide interfaces to share data, we can remove the overhead caused by shared_ptr.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,156 @@
...
+    float_t at(const cnn_size_t batch,  const cnn_size_t width,
+               const cnn_size_t height, const cnn_size_t depth) const {
+        return *access_data(batch, width, height, depth);
+    }
+
+    // Returns the pointer to a requested position in the tensor
+    float_t* ptr(const cnn_size_t batch,  const cnn_size_t width,
+                 const cnn_size_t height, const cnn_size_t depth) const {
+        return access_data(batch, width, height, depth);
+    }
```
nyanp: @edgarriba 
const method should not return non-const pointer. It is problematic if we can change the internal data via const method (it will break the coherency between cpu/gpu data). I want to have const/non-const overloading, and zero-overhead access like ```std::vector```.

```cpp
template <class T>
class vector {
public:
    // checked version (throw exceptions for out-of-range error)
    T& at(size_type index);
    const T& at(size_type index) const;

    // zero-overhead version (same performance to raw pointer access. have an assertion for out-of-range error)
    T& operator [] (size_type index);
    const T& operator [] (size_type index) const;
}; 
```

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,156 @@
...
+    // Method to access to the tensor data.
+    // It checks if the requested position is feasible or not.
+    float_t* access_data(const cnn_size_t batch,
+                         const cnn_size_t width,
+                         const cnn_size_t height,
+                         const cnn_size_t depth) const {
+        if ((batch  < 0 || batch  > shape_[0]) ||
+            (width  < 0 || width  > shape_[1]) ||
+            (height < 0 || height > shape_[2]) ||
+            (depth  < 0 || depth  > shape_[3])) {
```
nyanp: @edgarriba 
size_t is always non-negative. We can remove first condition for each argument.

tiny_dnn/core/framework/tensor.h
edgarriba: Are not tensors shared between layers?

bhack:
Do we want to take a look also on how OpenVX [data objects](https://www.khronos.org/registry/vx/specs/1.0/html/d0/d84/page_design.html#sec_data_objects) and virtual data objects are organized?

tiny_dnn/core/framework/tensor.h
nyanp: What we want to share is probably ```Tensor``` itself, instead of internal representation of ```Tensor```. We can share tensors between layers by ```std::shared_ptr<Tensor>```.

nyanp:
@bhack 
Nice pointer :)
I don't think we need to think seriously about this specification. Since OpenVX standard requires pure C interface and pure C struct decleration, we need a wrapper layer to adopt OpenCV specs no matter what we design this Tensor class.

bhack:
@nyanp Like what AMD has done https://github.com/GPUOpen-ProfessionalCompute-Libraries/amdovx-core/tree/master/openvx/api ;)

bhack:
And this is the OpenVX module that implemented a mechanism to access OpenCV functionality as OpenVX kernels https://github.com/GPUOpen-ProfessionalCompute-Libraries/amdovx-modules

tiny_dnn/core/framework/tensor.h
```diff
@@ -108,7 +110,17 @@ class Tensor {
...
+    void toDevice(const Device& device) {
+#if defined(USE_OPENCL) || defined(USE_CUDA)
+        CLCudaAPI::Context ctx = device.context();
+        CLCudaAPI::Queue queue = device.queue();
+
+        data_gpu_ = std::make_shared<CLCudaAPI::Buffer<float_t> >(
+            ctx, queue, data_cpu_->begin(), data_cpu_->end());
+#endif
+    }
+
```
edgarriba: @nyanp  @bhack 
this is what I think that should be a minimal API to transfer data between devices.
It still needs more logic to handle corner cases. Do you have any consideration?

bhack: It could be interested to know if @CNugteren will support [RDMA](https://streamcomputing.eu/blog/2015-04-18/gpudirect-and-directgma-direct-gpu-gpu-communication/) (probably we need a more update link) between devices in clcudaapi.

CNugteren: @bhack I have no current plans to add RDMA  to CLCudaAPI. But you can always submit a pull request of course!

nyanp: It's a nice starting point. We need to decide what kind of parallelism we will support for multi-gpu., because this choice would affect the API design of tensor class. I think data parallelism is a simple & reasonable choice, but I may be wrong..

patrikhuber:
I have no idea if this is helpful or even related, but there seems to be some ongoing effort to work on Eigen's tensor class and to make it work with OpenCL and suitable for deep-learning frameworks. I believe the discussion here https://github.com/tensorflow/tensorflow/issues/22 might be helpful. (at least it may give some inspiration with regards to a tensor class API)

Sorry for the noise if it's not helpful!

Randl:
Hi!

I'm willing to contribute to this fuctional. Can you provide some kind of roadmap - what should be done and in which order so it'll be possible to join development?

bhack:
@Randl Join the chat: https://gitter.im/tiny-dnn/developers

hcedwar:
Still targeting next library TS.  The ISO/C++ standard committee meeting this week is completely consumed with C++17, so progress on array_ref will resume at the next meeting.

PR #343, opened 2016-10-18
ios library compiling

wangyida:
This based on hunter and polly and also with the help of gates, so we can have iOS or andriod libs


nyanp:
👍 
@wangyida ready to merge?

wangyida:
@nyanp Yeah I tested it for the video I sent you previously~

PR #348, opened 2016-10-19
GPU tests

Randl:
GPU tests work and are passed on my PC. 

Also added an example of how devices can be checked. We may use it to check if there are compatible devices to run GPU tests (i.e. Travis)

@edgarriba @nyanp @bhack

### DIFF
--- a/test/test_core.h
+++ b/test/test_core.h
@@ -53,8 +53,19 @@ TEST(core, add_bad_device) {
 
     convolutional_layer<sigmoid> l(5, 5, 3, 1, 2,
         padding::valid, true, 1, 1, backend_t::opencl);
-
-    my_gpu_device.registerOp(l);
+    try {
+        my_gpu_device.registerOp(l);
+        EXPECT_TRUE(false);
+    }
+    catch (const nn_error& e) {
+        std::string err_mess = "Cannot register layer: " + l.layer_type() +
+                               ". Device has disabled OpenCL support. Please specify platform and device "
+                                       "in Device constructor";
+        EXPECT_STREQ(err_mess.c_str(), e.what());
+    }
+    catch(...) {
+        EXPECT_TRUE(false);
+    }
 }
 
 TEST(core, add_bad_layer) {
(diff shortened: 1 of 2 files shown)


bhack:
We can evaluate if some test could be covered with https://github.com/jrprice/Oclgrind in travis

Randl:
@bhack yet checking for compatible devices before running tests might be useful.

Randl:
Another option is to check if `USE_OPENCL` before running any GPU test

bhack:
Yes it is useful. I was only talking about the possibility to run some of the CL test also on Travis.

edgarriba:
@Randl rebase your codes

test/test_core.h
```diff
@@ -28,16 +28,80 @@
...
+    for (auto p = platforms.begin(); p != platforms.end(); ++p)
+        for (size_t d = 0; d < p->NumDevices(); ++d) {
+            auto dev = CLCudaAPI::Device(*p, d);
+            if (dev.Type()=="CPU") {
+                platform = p - platforms.begin();
+                device = d;
+                return device_t::CPU;
+            }
+        }
+
```
edgarriba: @Randl what's the difference between having this two loops here instead of one?

edgarriba:
@Randl the git history looks good now.

Just last comment: notice that I opened a PR to test code style and might happen in a future that you need do small modifications in your code.
https://github.com/edgarriba/tiny-cnn/blob/cpplint/.travis.yml#L92-L95

we are following this guide: https://google.github.io/styleguide/cppguide.html
Please, take a look at it.

test/test_core.h
Randl: @edgarriba Well I was prototyping, and two loops was most straightforward approach which I considered readable enough.

It can be refactored like that:
```
	size_t cpu_platform, cpu_device;
	bool has_CPU=false;
    for (auto p = platforms.begin(); p != platforms.end(); ++p)
        for (size_t d = 0; d < p->NumDevices(); ++d) {
            auto dev = CLCudaAPI::Device(*p, d);
            if (dev.Type()=="GPU") {
                platform = p - platforms.begin();
                device = d;
                return device_t::GPU;
            }
            if (!has_CPU && dev.Type()=="CPU") {
                cpu_platform = p - platforms.begin();
                cpu_device = d;
                has_CPU = true;
            }
        }

	//no GPU found
    if (has_CPU) {
		platform = cpu_platform;
		device = cpu_device;
		return device_t::CPU;
	}
	
	//no CPUs or GPUs
	return device_t::NONE;
```

Do you think it's more readable and intuitive?

edgarriba: I think it will be easier a set of if/elseif/else rules

```
    size_t cpu_platform, cpu_device;
    for (auto p = platforms.begin(); p != platforms.end(); ++p)
        for (size_t d = 0; d < p->NumDevices(); ++d) {
            auto dev = CLCudaAPI::Device(*p, d);
            platform = p - platforms.begin();
            device = d;
            if (dev.Type()=="CPU") {
                return device_t::CPU;
            } else if (dev.Type()=="CPU") {
                return device_t::CPU;
            } 
    return device_t::NONE;
```

Randl: Your code is different from mine. If CPU comes earlier than GPU, mine will return GPU and yours CPU. I believe we should prefer GPU if available

edgarriba: oh, I see. Mmm, but then what about future additional device types? Not sure if this codes will be generic. For this reason I proposed to have separate binaries by device type

Randl: Another option is:
```cpp
	auto devices_order = {"FPGA" /* future */, "GPU", "CPU"}
	auto devices_t_order = {device_t::FPGA /* future */, device_t::GPU, device_t::CPU}
	for (size_t d_type = 0; d_type < devices_order.size(); ++d_type)
		for (auto p = platforms.begin(); p != platforms.end(); ++p)
			for (size_t d = 0; d < p->NumDevices(); ++d) {
				auto dev = CLCudaAPI::Device(*p, d);
				if (dev.Type() == devices_order[d_type]) {
					platform = p - platforms.begin();
					device = d;
					return devices_t_order[d_type];
				}
			}
	return device_t::NONE;
```
Also we can give different priority to devices with `define`s.

edgarriba: I like more this one

Issue #336, opened 2016-10-10
Compilation errors when tiny_dnn::float_t type is defined as double


pi-null-mezon:
In the last commit of the version 1.0.0 (master branch) if you define float_t as double a lot of errors occur when 'tiny_quantized_(de)conv2d_kernel.h' is compiled. It seems, that in most cases this errors are produced because type float is used when float_t should be.

edgarriba:
@pi-null-mezon unfortunately we don't test against double type in TravisCI, so I'm not surprised that you find some errors. However, could paste the log? /cc @wangyida

pi-null-mezon:
Issue log is too big (contains 505 strings), but most of the errors point to this:
````
C:\Programming\3rdParties\tiny-dnn-1.0.0\tiny_dnn\core\kernels\tiny_quantized_conv2d_kernel.h:49: error: C2782: 'const _Ty &std::max(const _Ty &,const _Ty &) noexcept(<expr>)': template parameter '_Ty' is ambiguous
```` 
and they happens because function std::max/min is called with arguments pair (float, double).

PR #353, opened 2016-10-20
Double precision support #336

nyanp:
This is mainly for providing backward compatibility to v0.1.1.


nyanp:
@edgarriba 
Not required I think, but it's valuable to keep consistent coding style. I've added now.

PR #354, opened 2016-10-21
Auto code style

Randl:
All code base was processed with CLion's 'Reformat code' tool with 'google' style.


edgarriba:
@Randl I think that we have to do this by file in order to assure consistency. Notice that cpplint gives a lot of warnings that probably have to be fixed by hand

Randl:
Closed in favor of https://github.com/tiny-dnn/tiny-dnn/pull/363

Issue #357, opened 2016-10-21
Caffe Converter Needs to Be Able to Import Power Layer


goranrauker:
When importing some caffe models (see below) a power_layer is required by the converter.  The current layer_factory_impl does not provide it as an option.  In addtion, the power_layer needs to be updated to allow for the optional scale parameter.

`input: "data"
input_shape {
  dim: 1
  dim: 3
  dim: 28
  dim: 28
}
layer {
  name: "scale"
  type: "Power"
  bottom: "data"
  top: "scale"
  power_param {
    scale: 0.0125
  }
}
layer {
  name: "conv1"
  type: "Convolution"
  bottom: "scale"
  top: "conv1"
  param {
    lr_mult: 1.0
  }
  param {
    lr_mult: 2.0
  }
  convolution_param {
    num_output: 20
    kernel_size: 5
    stride: 1
    weight_filler {
      type: "xavier"
    }
    bias_filler {
      type: "constant"
    }
  }
}
layer {
  name: "pool1"
  type: "Pooling"
  bottom: "conv1"
  top: "pool1"
  pooling_param {
    pool: MAX
    kernel_size: 2
    stride: 2
  }
}
layer {
  name: "conv2"
  type: "Convolution"
  bottom: "pool1"
  top: "conv2"
  param {
    lr_mult: 1.0
  }
  param {
    lr_mult: 2.0
  }
  convolution_param {
    num_output: 50
    kernel_size: 5
    stride: 1
    weight_filler {
      type: "xavier"
    }
    bias_filler {
      type: "constant"
    }
  }
}
layer {
  name: "pool2"
  type: "Pooling"
  bottom: "conv2"
  top: "pool2"
  pooling_param {
    pool: MAX
    kernel_size: 2
    stride: 2
  }
}
layer {
  name: "ip1"
  type: "InnerProduct"
  bottom: "pool2"
  top: "ip1"
  param {
    lr_mult: 1.0
  }
  param {
    lr_mult: 2.0
  }
  inner_product_param {
    num_output: 500
    weight_filler {
      type: "xavier"
    }
    bias_filler {
      type: "constant"
    }
  }
}
layer {
  name: "relu1"
  type: "ReLU"
  bottom: "ip1"
  top: "ip1"
}
layer {
  name: "ip2"
  type: "InnerProduct"
  bottom: "ip1"
  top: "ip2"
  param {
    lr_mult: 1.0
  }
  param {
    lr_mult: 2.0
  }
[...]

PR #358, opened 2016-10-21
Feature/adds power layer to caffe converter

goranrauker:
This pull request should resolve issue #357.  I'd appreciate someone reviewing the back propagation.

### DIFF
--- a/tiny_dnn/layers/power_layer.h
+++ b/tiny_dnn/layers/power_layer.h
@@ -39,14 +39,14 @@ class power_layer : public layer {
 public:
     typedef layer Base;
 
-    power_layer(const shape3d& in_shape, float_t factor)
+    power_layer(const shape3d& in_shape, float_t factor, float_t scale=1.0f)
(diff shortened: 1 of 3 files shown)


tiny_dnn/layers/power_layer.h
```diff
@@ -84,11 +84,11 @@ class power_layer : public layer {
 
         for (cnn_size_t i = 0; i < x.size(); i++) {
             for (cnn_size_t j = 0; j < x[i].size(); j++) {
-                // f(x) = x^factor
+                // f(x) = (scale*x)^factor
                 // ->
                 //   dx = dy * df(x)
-                //      = dy * factor * x^(factor-1)
-                //      = dy * factor * f(x) / x
+                //      = dy * scale * factor * (scale * x)^(factor - 1)
```
edgarriba: @goranrauker you missed to multiply by the scale value un line 91

goranrauker: Thanks - good catch!

PR #359, opened 2016-10-22
cpplint version bump

Randl:


edgarriba:
@Randl Is this an updated version?

Randl:
Yeah this is latest version from [pip](https://pypi.python.org/pypi/cpplint)

edgarriba:
Nice!

Issue #242, opened 2016-07-20
Switch test framework


edgarriba:
I open this ticket to discuss about switching the test framework. My idea is the following:
- Migrate to [GTest](https://cmake.org/cmake/help/v3.0/module/FindGTest.html). Force user to install it and use command `findPackage(GTest)`.
- Include tests in CMake format using command  `add_test`.

patrikhuber:
When I last checked 1 year ago, it was recommended by GTest authors to
include GTest into the project and not use and external, separately built
one. I believe one reason was so that compile flags match.

Have you looked at Catch? It's header-only and quite appealing.

edgarriba:
@patrikhuber whatever option that's easy to integrate with CMake is okay.
However, Gtest you can easily get it via `apt-get install libgtest-dev` if I'm not wrong.

patrikhuber:
I think the reason is in their FAQ (and it was a pretty good one), if you
can't find it I'll check tomorrow. I think I went with add_subdirectory
(you can copy from my superviseddescent project on GitHub) but better first
check if this info is still up to date :)

edgarriba:
If we agree to include third-party libraries at least for testing, your solution looks good to me.

patrikhuber:
Found it, it's in the [GTest FAQ](https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-is-it-not-recommended-to-install-a-pre-compiled-copy-of-google-test-for-example-into-usrlocal) :-) Apparently it's because of ODR and `#ifdef`'s in GTest, but I don't fully understand without thinking more about it.

Anyway I recommend giving [Catch](https://github.com/philsquared/Catch) a thought, I think if I were to redo my Unit tests I might switch (not that setup/integration of GTest is a huge hassle, but there can be some issues and a header-only file is certainly easier).

nyanp:
I think it's no problem to include third-party libraries under specific (e.g. ```thrid_party```) directory for testing, if its license has no problem to include. Since picotest has compatibility with gtest and I don't think gtest has critical problems, I vote for gtest.
I like Catch as a header-only-believer, and it's a very good alternative if we have problem with gtest :)

PR #361, opened 2016-10-23
replacing picotests with gtests 

Randl:
Added gtest. https://github.com/tiny-dnn/tiny-dnn/issues/242


patrikhuber:
@Randl Did you consider my comment in #242? If you want to switch to GTest, I think its sources should be included into tiny-dnn and it should be built with the tests, for a variety of reasons. Let me know in case I'm wrong.

Randl:
@patrikhuber My code checkouts gtest from github

edgarriba:
@Randl Could you move Download project into tiny-dnn/cmake?

patrikhuber:
@Randl Ok, I see what DownloadProject is doing now. Can we really not get by with just `ExternalProject_Add`? Every additional script adds a complexity overhead to the build. Adding a whole new 3rd party tool just to use a unit test framework is a bit... yea... unfortunate. At this point, if it was my own project, I would really strongly consider Catch - header-only and no other overhead like building or other build scripts necessary.

edgarriba:
@patrikhuber To use Catch we need to rewrite test codes. Maybe in a future we can give it a shot. Other priorities now.

bhack:
Catch it is interesting but actually we don't have the bandwidth to rewrite tests cause the todo list it is quite busy. Better Gtest that picotest to have now.

patrikhuber:
Sure. I'm just advocating for less complexity, whatever you decide to do.

bhack:
@patrikhuber /off topic/ Do you want to take any bullet on the to do list? ;)

patrikhuber:
@bhack Hehehe! I've been considering for quite a while, but never got around to it - I'm just too busy with my own PhD and 3D face fitting software (particularly now that I'm in my last 6 months). Actually the feature I'd like to see most in tiny-dnn is a sparse AutoEncoder :-) I think a sparsity constraint/layer is still missing. In any way, I'll let you guys know if I can take up on any tasks - sorry that it's not now.

Randl:
@edgarriba yeah sure, I was considering it, but then thought that it'll be better to keep licence and readme and so put it into third party. 
@patrikhuber  well as for me simple script to checkout dependency isn't that critical, given we can keep all the tests

Randl:
@nyanp @bhack  @edgarriba Any problem to merge?

edgarriba:
one test do not pass
https://ci.appveyor.com/project/tinydnn/tiny-dnn/build/master-328/job/1ipwdyssouo6xlmv#L757

edgarriba:
@Randl check my review comment.
You should consider to add this [lines](https://github.com/dmonopoly/gtest-cmake-example/blob/master/CMakeLists.txt#L53) in order to be able to call the `make test` command

edgarriba:
@Randl 
Please also rebase from master to check that tests pass with double precision 10a02710a929986fcdc36ac14d080eaf4cd3839a

edgarriba:
@Randl  thx. Very preciate contribution!

PR #362, opened 2016-10-24
quick conv_params refactor

edgarriba:


Randl:
Should probably remove include [here](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_op_opencl.h#L47) and [here](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_op_libdnn.h#L47)

edgarriba:
@nyanp 
I found something that I don't understand in the padding codes and probably makes the test crash

Why in the version inside the `conv_laye`r we return void when `pad_type` is equal to valid instead of return the input tensor?

https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/layers/convolutional_layer.h#L527-528
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d.h#L61-L63

nyanp:
@edgarriba 
Because of performance issue : We use ```in_data``` directly when ```pad_type``` is equal to ```valid```, to avoid unnecessary memory allocation.

https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/layers/convolutional_layer.h#L414

PR #364, opened 2016-10-26
Allows for an Archive mode in Caffe Converter Example

goranrauker:
The idea here is to create a single "archive" tiny-dnn model that contains the converted caffe network, mean file and labels so that it can be used freely without caffe or protobuf dependencies.   I think this would get new users up and running quicker.

I stopped short on the implementation as I wanted to start a discussion on this topic and gauge interest.  But, I can envision a tiny_dnn/models/archive.hpp that would be able to consume the converted network and predict on an image<>.  

Thoughts?

As for the converter, a more complete example would include a "test_archive" mode so the user could run a test image against it.  This mode would obviously use the archive.hpp to predict.


edgarriba:
@goranrauker your idea sounds good.
Please join us in the gitter developers room to discuss your proposal:  https://gitter.im/tiny-dnn/developers

nyanp:
@goranrauker 
Nice!
Did you confirm that your code works fine with AlexNet?

PR #366, opened 2016-10-28
Add include path to protobuf when building caffe_converter.

syoyo:
This PR simply adds include path to protobuf when building caffe_converter example to fix the build on some platform(e.g. In macOS + homebrew installed protobuf, we need to add `/usr/local/include` to the include path list)


edgarriba:
LGTM

nyanp:
👍

PR #370, opened 2016-10-29
sync with latest nnpack

azsane:
- convolution layer always do padding,so ignore padding in nnpack
  or convolution_layer must judge by engine value whether it shoud do padding  
- nnpack convolution-inference.c will throw an error,when stride is more than 1
  maybe a bug? so use auto algorithm stage may encounter this error.

### DIFF
--- a/tiny_dnn/core/kernels/conv2d_op_nnpack.h
+++ b/tiny_dnn/core/kernels/conv2d_op_nnpack.h
@@ -35,8 +35,8 @@ inline nnp_convolution_algorithm nnp_algorithm() {
     return nnp_convolution_algorithm_auto;
 }
 
-inline nnp_convolution_kernel_transform_strategy nnp_kts() {
-    return nnp_convolution_kernel_transform_strategy_reuse;
+inline nnp_convolution_transform_strategy nnp_kts() {
+    return nnp_convolution_transform_strategy_tuple_based;//some algorithm accept tuple based only
 }
 #endif
 
@@ -62,9 +62,11 @@ conv2d_op_nnpack(const tensor_t&         in_data,
     const cnn_size_t input_channels = params.in.depth_;
     const cnn_size_t output_channels = params.out.depth_;
 
+    //input data passed by convolution layer has been padded already
+    //set input_size to padded size
     const nnp_size input_size = {
-        static_cast<size_t>(params.in.width_),
-        static_cast<size_t>(params.in.height_)
+        static_cast<size_t>(params.in_padded.width_),
+        static_cast<size_t>(params.in_padded.height_)
     };
 
     const nnp_size kernel_size = {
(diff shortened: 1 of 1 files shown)


edgarriba:
@azsane just make sure that codes pass local tests compiled with `USE_NNPACK=ON`. It would very appreciate if you can provide additional tests for NNPACK to cover corner cases.

Regarding the possible bugs: @Maratyszcza did you notice in your tests?

azsane:
@edgarriba  @Maratyszcza

When stride is more than 1,the algorithm only can be implict GEMM.
But the if below has overidden the algorithm

```
if (algorithm == nnp_convolution_algorithm_auto) {
		if ((max(kernel_size.width, kernel_size.height) > 16) ||
			(max(output_subsampling.width, output_subsampling.height) > 1))
		{
			algorithm = nnp_convolution_algorithm_implicit_gemm;
		} **missing else here ? ** if (max(kernel_size.width, kernel_size.height) > 8) {
			algorithm = nnp_convolution_algorithm_ft16x16;
		} 
```

Maratyszcza:
@azsane Right, `else` was missing. Thanks for finding this, I patched the bug upstream.

As for padding, you can pass `(struct nnp_padding) { 0, 0, 0, 0 }` if you don't need it.

bhack:
@azsane If we can have any addition NNPACK test coverage with this PR we try to merge before a2 release. In this case has a special relevance cause we still don't have an NNPACK upstream versioning.

edgarriba:
@azsane thx!
Could you also check and update the following codes
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/fully_connected_op_nnpack.h#L42
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/nnp_maxpool_kernel.h#L67

azsane:
@edgarriba 

nnpack only optimize pool size 2 stride 2 maxpool,so I didn't use it.

And after a quick check ,I found some problem in nnp_maxpool_kernel.h

code after modified
```
const float* input_ptr = in[0].data();
                float*      output_ptr =a[0].data();

                // TODO: embed it into a class
                const size_t num_mkl_threads = 1;
                pthreadpool_t threadpool = pthreadpool_create(num_mkl_threads);

                const auto status =
                    nnp_max_pooling_output(
                        1, // < - this parameter requires batch_size not input_channels
                        input_channels, // <-hence no outputchannels here ,input_channels instead
                        input_size,
                        input_padding,
                        pooling_size,
                        pooling_stride,
                        input_ptr,
                        output_ptr,
                        threadpool);
```

And another problem in backend_nnp.h

```
 void maxpool(const std::vector<tensor_t*>& in_data,
                 std::vector<tensor_t*>&       out_data) override {
       // nnp support pool size !=2 or stride !=2 no need to check the constraint below
        if (params_m_->stride_ != 2) {
            throw nn_error("NNPACK Max-Pool requires a stride == 2.");
        }

        if (params_m_->pool_size_ != 2) {
            throw nn_error("NNPACK Max-Pool requires a pool size == 2.");
        }

        const tensor_t& in = *in_data[0];
        tensor_t&       a = *out_data[1];

        kernels::nnp_maxpool_kernel(*params_m_, in, a);
    }
}
```

edgarriba:
@azsane oh, good finds! Please consider to check recent to submit a fix to this
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/maxpool_op.h
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/maxpool_op_nnpack.h

azsane:
@edgarriba a PR is submitted to fix the problem in maxpool & fc layer

PR #374, opened 2016-10-30
Add pad_type and non-square shape into pooling layers

nyanp:


nyanp:
@edgarriba 
oops, thanks :)

PR #376, opened 2016-10-31
refactor max_pool layer to op

edgarriba:


edgarriba:
@nyanp Could you review this PR? No sure if I missed sth during refactoring since tests do not pass. thx!

tiny_dnn/core/params/maxpool_params.h
```diff
@@ -30,20 +30,33 @@
...
     size_t              stride_y;
     padding             pad_type;
+    
+    /* mapping out => max_index(in) (1:1) */
+    std::vector<std::vector<cnn_size_t>> out2inmax;
+    /* mapping out => in (1:N) */
+    std::vector<std::vector<cnn_size_t> > out2in;
+    /* mapping in => out (N:1) */
+    std::vector<cnn_size_t> in2out;
 };
```
edgarriba: @nyanp Do you think that we need an explicit move ctor here?

tiny_dnn/core/kernels/maxpool_grad_op.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace tiny_dnn {
+
+class MaxPoolGradOp : public core::OpKernel {
+ public:
+    explicit MaxPoolGradOp(const core::OpKernelConstruction& context)
+        : core::OpKernel(context) {}
+
+    void compute(const core::OpKernelContext& context) override {
+        auto params = OpKernel::params_->maxpool();
```
edgarriba: @nyanp problem fixed. Thx!

Regarding the `any` concept, mmm what's the main advantage of introducing it? In my opinion having such structure might lead to bad designs since it makes the things too generic, no? Just a personal opinion.

PR #378, opened 2016-11-03
Type changes

pansk:
* conv2d_op_opencl was not compiling due to non-existent type ushort, changed to cl_ushort (I assumed that was the intent)
* removed a couple of useless floor(size_t), removed static_cast<size_t>, changed type of two variables wieght_w_half and wieght_h_half in order to avoid cnn_size_t<->size_t warnings.
* removed another cnn_size_t-related warning

### DIFF
--- a/examples/cifar10/train.cpp
+++ b/examples/cifar10/train.cpp
@@ -76,7 +76,7 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream& log) {
 
     cout << "start learning" << endl;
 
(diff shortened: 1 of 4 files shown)


edgarriba:
@pansk conv2d_op_opencl is completely experimental. To use opencl based convolutional check this https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_op_libdnn.h

edgarriba:
thx!

PR #379, opened 2016-11-03
Fix modification of CMAKE_CXX_FLAGS_RELEASE and CMAKE_CXX_FLAGS_DEBUG in CMakeLists.txt

beru:
In original CMakeLists.txt,
CMAKE_CXX_FLAGS_RELEASE is replaced by EXTRA_C_FLAGS_RELEASE
and CMAKE_CXX_FLAGS_DEBUG is replaced by EXTRA_C_FLAGS_DEBUG.

Replacing those flags results in loss of many default settings, so this PR fixes the problem by
appending the extra flags.


edgarriba:
LGTM

@beru just realized that there are some style errors near your patch codes.
Could you fix the extra tabs in the following lines?
https://github.com/tiny-dnn/tiny-dnn/blob/master/CMakeLists.txt#L201
https://github.com/tiny-dnn/tiny-dnn/blob/master/CMakeLists.txt#L228-L236
https://github.com/tiny-dnn/tiny-dnn/blob/master/CMakeLists.txt#L222-L223

If you see something more, go ahead.

beru:
@edgarriba 
Thank you for checking this PR.
I changed all tabs into 4spaces in CMakeLists.txt

Somehow this TravisCI job did't work out well...
https://travis-ci.org/tiny-dnn/tiny-dnn/jobs/173437129
and Coverage % decreased a little bit (-0.07%).

edgarriba:
@beru
I restarted TravisCI jobs and now tests are okay. thx for the contribution!

BTW, have you noticed any performance improvement with this patch?

beru:
@edgarriba 
I haven't checked performance differences with this patch.

To measure performance, what about switching backend type with command line argument and using shell script to run example programs?

Issue #356, opened 2016-10-21
Serialization status


Kamairo:
Hi all,
I am trying to train a network on an i7 Linux Mint 64-Bit and load and run it on an Intel Atom (custom) Gentoo 32-Bit.
However, I get a Floating Point Exception when trying to load the trained network on the Atom.
So:
1.  Is this supposed to be finished and working? #276 Makes me think it might be
2. (Can't find the issue) stated that Conv Layer serialization is not finished, but it seems to work from the pc to the same pc
3. #316 gets me confused about cereal, which version should I choose?
4. If serialization is not supposed to work cross-platform right now, can I somehow implement it myself? For example construct the network via the json files and serialize the weights in and out for each layer or do I need to serialize more parameters?

patrikhuber:
Just a wild guess, is the endianness different? If so, cereal has [PortableBinaryArchive](http://uscilab.github.io/cereal/serialization_archives.html#binary_archive) to be fully cross-platform compatible.

Kamairo:
Thanks for the quick reply. I also had that feeling about the endianess, might explain why a Floating Point Exception is thrown, e.g. division by zero. Unfortunately I do not have the atom at home, but I will test it ASAP. And thx for the cereal function tip, I will check that out if the Endianess is different.

nyanp:
@Kamairo
tiny-dnn often use size_t for serialization type, but it is problematic as you pointed out. I will change them to ```cnn_size_t``` (now it is a synonym of size_t, but will be changed to a synonym of uint64_t) in the next (maybe 1.0.0a2) release.


 > 2. JSON and XML are not working at all (import), not even on the same machine and not even only the network model (excluding the weights).


It is apparently a bug in tiny-dnn. Could you tell me your network model?

Kamairo:
Sure, it is enough to add one convolutional layer. For example:
```
net << conv<activation::sigmoid>(20, 20, 5, 10,  5, padding::same, true, 2, 2);
```
Then `net.to_json()` produces:
```
{
    "nodes": [
        {
            "type": "conv<sigmoid>",
            "in_size": {
                "width": 20,
                "height": 20,
                "depth": 10
            },
            "window_width": 5,
            "window_height": 5,
            "out_channels": 5,
            "connection_table": {
                "rows": 0,
                "cols": 0,
                "connection": "all"
            },
            "pad_type": 1,
            "has_bias": true,
            "w_stride": 2,
            "h_stride": 2
        }
[...]

PR #381, opened 2016-11-06
Add some tests & update docs

nyanp:
- Add tests for serializers to confirm #356
- Update docs
- Fix serialization for concat-layer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ addons:
       - lcov
       # caffe-importer dependencies
       - libprotobuf-dev
+      - protobuf-compiler
 
 branches:
   only:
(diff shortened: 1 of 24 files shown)


edgarriba:
@nyanp why is autodoc.py needed?

nyanp:
@edgarriba 
It's painful to write "layer catalogue" by hand... so I decided to use simple python script to generate markdown docs from source code. You can see similar idea at:
https://github.com/fchollet/keras/blob/master/docs/autogen.py

edgarriba:
looks good overall, just check typo error in constructor arguments

nyanp:
@edgarriba 
Thanks :) could you point out where I have a typo?

edgarriba:
for typo I mean that arguments from the generated docs are not alligned

PR #382, opened 2016-11-06
Fix an unnecessary compiler warning

reunanen:


nyanp:
@reunanen 
Nice fix 👍 Thanks!

edgarriba:
@nyanp why not `cnn_size_t` here?

PR #383, opened 2016-11-06
Make result::print_detail a const function

reunanen:
Problem: `result::print_detail` is non-`const`, meaning that the details cannot be printed on `const result` objects.

Solution: Make the method `const`. Adds some code, but won't ever change the state of the object.


nyanp:
@reunanen 
👍

PR #384, opened 2016-11-06
Fix another compiler warning

reunanen:
Fix another compiler warning, this time when building 64-bit (so `sizeof size_t > sizeof cnn_size_t`).

The fixed warning (again with MS Visual C++):
`tiny_dnn/util/util.h(128): warning C4267: 'return' : conversion from 'size_t' to 'tiny_dnn::cnn_size_t', possible loss of data`


nyanp:
@reunanen
Thanks again ;)

PR #385, opened 2016-11-07
rename tiny_dnn backend to custom

edgarriba:


beru:
In this case, custom means custom-made or customizable?
I've thought ex-tiny_dnn (now custom) backend as a reference backend.

edgarriba:
@beru custom stands for custom-made. Do you think is not clear the rename?

bhack:
Is it considered a fallback engine? Internal? I don't think that custom is one of the best naming ;)

nyanp:
"internal" or "native" seems good for me :)

edgarriba:
oki, native sounds good. I'll rename again

bhack:
native could be mislead to the native code concept of not compiled langauges

edgarriba:
generic?

bhack:
I think that it is internal.. if we consider that a fallback (I.e. In the case I don't want any external library) we could call fallback

patrikhuber:
Fallback also has a slightly negative conotation (i.e. "If all else fails, use this, but it will not be optimal"). Is this the case? If not, "internal" might be the best and most neutral name!

edgarriba:
internal then, because the backends do not follow any rule to consider a fallback case

edgarriba:
And talking about names convention: backends or engines ?

bhack:
This is more diffucult :)

patrikhuber:
"backend" sounds more natural to me - it's a backend (mostly linear algebra I suppose), that does computation. I don't think I'd really call it an "engine".

bhack:
Others have called engines https://github.com/caffe2/caffe2/blob/master/caffe2/contrib/nnpack/nnpack_ops.cc#L260

PR #395, opened 2016-11-10
Minor optimization in load_weights_fullyconnected

reunanen:
Problem: Caffe-trained AlexNet loads forever, at least in debugger (Visual Studio).

Solution: Get dst in and out sizes only once, and not separately for every iteration.


edgarriba:
@reunanen when you say optimization it's because you noticed an improvement in terms of performance?

reunanen:
@edgarriba Yes, correct.

edgarriba:
@reunanen nice! thx for the PR!

PR #398, opened 2016-11-12
fix max-pool & full-connected layer problem when using nnpack

azsane:
* make max-pool layer work using  nnpack and test
* make  full-connected layer work using nnpack and test
* add a stride size !=2 pool size !=2 test for max-pool

code modified pass the test for max-pool & full-connected layer


Kamairo:
Unfortunately I get totally different results from FC layer with NNPack than with tinydnn backend.

edgarriba:
@Kamairo could you open a ticket for this specific issue? Report backtrace and so on. Thx!

azsane:
@edgarriba the conflict in maxpool_op_nnpack.h is solved

edgarriba:
@azsane cool! just fix style typo that I pointed out in my review and we'll be good to merge. Also notice https://github.com/tiny-dnn/tiny-dnn/pull/416/files#diff-26d451c6b2b93a9e871ebcc919d1488bR98

azsane:
@edgarriba I didn't find review in 398...so I just modified some variables like #416 

and the thing should be noticed is that conv has changed to nnp_convolution_output?

edgarriba:
@azsane it's fine. I'll fix typos myself. Please could you check why [nnp_convolution_inference](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_op_nnpack.h#L107) breaks the tests?

azsane:
@edgarriba I just ran the conv test with latest master branch,and all passed.
Do you mean the newly modified code in #416 ?

Issue #41, opened 2015-09-16
read the trained model and predict error


kingloo2014:
I have a trained model and saved it. But a mistake happened when I predict a test sample's class attribute. Of course, I have load the model before the prediction correctly. The debugging information gave the tips "the input layer's size is empty", I donnot understand the information, because the entire proccess is right!

kingloo2014:
I found the error. when the model is a member varable of a class, the model's parameters will be modified at the different member function. I donnot know why!

nyanp:
@kingloo2014 
It would be helpful if you show me example code and tell me your compiler version. Thanks!

edgarriba:
@kingloo2014 Still having this issue?

Randl:
@kingloo2014 Do you still have this issue?

Randl:
No reply, closing

PR #400, opened 2016-11-12
add Tensor class

edgarriba:
Add the Tensor structure:
- data hold by `std::vector<>` with 64bytes alignment
- three different data accessors: `t.ptr<float_t>()`, `t.at<float_t>()` and `t[idx]`
- basic `reshape()` and `resize()` routines
- basic `toDevice()` and `fromDevice()` routines
- implement element-wise add, sub, mul, div

### DIFF
--- a/test/test_tensor.h
+++ b/test/test_tensor.h
@@ -392,6 +392,27 @@ TEST(tensor, fill) {
     }
 }
 
+TEST(tensor, linspace) {
+    Tensor<float_t> tensor(2,2,2,2);
+
+    // fill all tensor values with values from 1 to 16
+
+    tensor.linspace(float_t(1.0), float_t(16.0));
+
+    for (size_t i = 0; i < tensor.size(); ++i) {
+        EXPECT_EQ(tensor[i], float_t(1.0+i));
+    }
+
+    Tensor<float_t> tensor2(101,1,1,1);
+    // fill all tensor values with from 0 to 1
+
+    tensor2.linspace(float_t(0.0), float_t(1.0));
+
+    for (size_t i = 0; i < tensor2.size(); ++i) {
+        EXPECT_NEAR(tensor2[i], float_t(0.01*i), 0.001);
+    }
+}
+
 TEST(tensor, add1) {
     Tensor<float_t> t1(2,2,2,2);
     Tensor<float_t> t2(2,2,2,2);
@@ -541,6 +562,40 @@ TEST(tensor, div2) {
 
     for (size_t i = 0; i < t2.size(); ++i) {
         EXPECT_EQ(t2[i], float_t(0.5));
+    }
+}
+
+TEST(tensor, sqrt) {
+    Tensor<float_t> t(2, 2, 2, 2);
+
+    // fill tensor with initial values
+    t.fill(float_t(4.0));
+
+    // calculate square root
+
+    Tensor<float_t> t2 = t.sqrt();
+
+    // check that root is okay
+
+    for (size_t i = 0; i < t2.size(); ++i) {
+        EXPECT_EQ(t2[i], float_t(2.0));
+    }
+}
+
+TEST(tensor, exp) {
+    Tensor<float_t> t(2, 2, 2, 2);
+
+    // fill tensor with initial values
+    t.linspace(float_t(1.0), float_t(16.0));
+
+    // calculate exp
+
+    Tensor<float_t> t2 = t.exp();
+
+    // check that exponent is okay
+
+    for (size_t i = 0; i < t2.size(); ++i) {
+        EXPECT_NEAR(t2[i], float_t(exp(float_t(i+1))), 1e-5);
     }
 }
 
--- a/test/test_tensor.h
+++ b/test/test_tensor.h
@@ -34,7 +34,7 @@ using namespace tiny_dnn;
 namespace tiny_dnn {
 
 TEST(tensor, shape) {
-    Tensor tensor(1,2,2,2);
+    Tensor<float_t,1,2,2,2> tensor;
 
     EXPECT_EQ(tensor.shape()[0], cnn_size_t(1));
     EXPECT_EQ(tensor.shape()[1], cnn_size_t(2));
@@ -43,11 +43,11 @@ TEST(tensor, shape) {
 }
 
 TEST(tensor, access_data) {
-    Tensor tensor(1,2,2,2);
+    Tensor<float_t,1,2,2,2> tensor;
 
-    float_t* begin_ptr = tensor.ptr(0,0,0,0);
-    float_t* end1_ptr  = tensor.ptr(0,1,1,0);
-    float_t* end2_ptr  = tensor.ptr(0,1,1,1);
+    float_t* begin_ptr = tensor.ptr<float_t>(0,0,0,0);
+    float_t* end1_ptr  = tensor.ptr<float_t>(0,1,1,0);
+    float_t* end2_ptr  = tensor.ptr<float_t>(0,1,1,1);
 
     // set tensor data
     
@@ -67,9 +67,9 @@ TEST(tensor, access_data) {
         for (cnn_size_t j = 0; j < 2; ++j) {
             for (cnn_size_t k = 0; k < 1; ++k) {
                 if (k == 0) {
-                    EXPECT_EQ(tensor.at(0,i,j,k), cnn_size_t(1));
+                    EXPECT_EQ(tensor.at<float_t>(0,i,j,k), cnn_size_t(1));
                 } else {
-                    EXPECT_EQ(tensor.at(0,i,j,k), cnn_size_t(2));
+                    EXPECT_EQ(tensor.at<float_t>(0,i,j,k), cnn_size_t(2));
                 }
             }
         }
--- a/test/test_tensor.h
+++ b/test/test_tensor.h
@@ -42,37 +42,291 @@ TEST(tensor, shape) {
     EXPECT_EQ(tensor.shape()[3], cnn_size_t(2));
 }
 
-TEST(tensor, access_data) {
+TEST(tensor, access_data1) {
+    Tensor<float_t> tensor(1,2,2,1);
+
+    EXPECT_EQ(tensor.at<float_t>(0,0,0,0), float_t(0.0));
+    EXPECT_EQ(tensor.at<float_t>(0,0,1,0), float_t(0.0));
+    EXPECT_EQ(tensor.at<float_t>(0,1,0,0), float_t(0.0));
+    EXPECT_EQ(tensor.at<float_t>(0,1,1,0), float_t(0.0));
+
+    EXPECT_THROW(tensor.at<float_t>(0,0,0,1), nn_error);
+    EXPECT_THROW(tensor.at<float_t>(1,0,0,0), nn_error);
+    EXPECT_THROW(tensor.at<float_t>(1,0,0,1), nn_error);
+}
+
+TEST(tensor, access_data2) {
+    Tensor<float_t> tensor(1,2,2,1);
+
+    const float* ptr = tensor.ptr<float_t>(0,0,0,0);
+
+    size_t i = 0;
+    EXPECT_EQ(ptr[i++], float_t(0.0));
+    EXPECT_EQ(ptr[i++], float_t(0.0));
+    EXPECT_EQ(ptr[i++], float_t(0.0));
+    EXPECT_EQ(ptr[i++], float_t(0.0));
+}
+
+TEST(tensor, access_data3) {
+    Tensor<float_t> tensor(1,2,2,1);
+
+    size_t i = 0;
+    EXPECT_EQ(tensor[i++], float_t(0.0));
+    EXPECT_EQ(tensor[i++], float_t(0.0));
+    EXPECT_EQ(tensor[i++], float_t(0.0));
+    EXPECT_EQ(tensor[i++], float_t(0.0));
+}
+
+TEST(tensor, access_data4) {
     Tensor<float_t> tensor(1,2,2,2);
 
-    float_t* begin_ptr = tensor.ptr<float_t>(0,0,0,0);
-    float_t* end1_ptr  = tensor.ptr<float_t>(0,1,1,0);
-    float_t* end2_ptr  = tensor.ptr<float_t>(0,1,1,1);
+    // modify data using .ptr() accessor
 
-    // set tensor data
-    
-    // channel #1
-    for (float_t* i = begin_ptr; i != end1_ptr + 1; i++) {
-        *i = float_t(1);
+    float_t* ptr1 = tensor.ptr<float_t>(0,0,0,0);
+    float_t* ptr2 = tensor.ptr<float_t>(0,0,0,1);
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        ptr1[i] = float_t(1.0);
     }
 
-    // channel #2
-    for (float_t* i = end1_ptr + 1; i != end2_ptr + 1; i++) {
-        *i = float_t(2);
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        ptr2[i] = float_t(2.0);
     }
 
-    // check data
+    // check data using .ptr() accessor
+
+    const float_t* ptr11 = tensor.ptr<float_t>(0,0,0,0);
+    const float_t* ptr22 = tensor.ptr<float_t>(0,0,0,1);
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        EXPECT_EQ(ptr11[i], float_t(1.0));
+    }
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        EXPECT_EQ(ptr22[i], float_t(2.0));
+    }
+}
+
+TEST(tensor, access_data5) {
+    Tensor<float_t> tensor(1,2,2,2);
+
+    // modify data using .ptr() accessor
+
+    float_t* ptr1 = tensor.ptr<float_t>(0,0,0,0);
+    float_t* ptr2 = tensor.ptr<float_t>(0,0,0,1);
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        ptr1[i] = float_t(1.0);
+    }
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        ptr2[i] = float_t(2.0);
+    }
+
+    // check data using .at() accessor
+
+    for (cnn_size_t i = 0; i < 2; ++i) {
+        for (cnn_size_t j = 0; j < 2; ++j) {
+            EXPECT_EQ(tensor.at<float_t>(0,i,j,0), float_t(1.0));
+        }
+    }
+
+    for (cnn_size_t i = 0; i < 2; ++i) {
+        for (cnn_size_t j = 0; j < 2; ++j) {
+            EXPECT_EQ(tensor.at<float_t>(0,i,j,1), float_t(2.0));
+        }
+    }
+}
+
+
+TEST(tensor, access_data6) {
+    Tensor<float_t> tensor(1,2,2,2);
+
+    // modify data using .ptr() accessor
+
+    float_t* ptr1 = tensor.ptr<float_t>(0,0,0,0);
+    float_t* ptr2 = tensor.ptr<float_t>(0,0,0,1);
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        ptr1[i] = float_t(1.0);
+    }
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        ptr2[i] = float_t(2.0);
+    }
+
+    // check data using operator[] accessor
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        EXPECT_EQ(tensor[i], float_t(1.0));
+    }
+
+    for (cnn_size_t i = 0; i < 4; ++i) {
+        EXPECT_EQ(tensor[4 + i], float_t(2.0));
+    }
+}
+
+TEST(tensor, access_data7) {
+    Tensor<float_t> tensor(1,2,2,2);
+
+    // modify data using .at() accessor
+
+    tensor.at<float_t>(0,0,0,0) = float_t(1.0);
+    tensor.at<float_t>(0,0,1,0) = float_t(1.0);
+    tensor.at<float_t>(0,1,0,0) = float_t(1.0);
+    tensor.at<float_t>(0,1,1,0) = float_t(1.0);
+
+    tensor.at<float_t>(0,0,0,1) = float_t(2.0);
+    tensor.at<float_t>(0,0,1,1) = float_t(2.0);
+    tensor.at<float_t>(0,1,0,1) = float_t(2.0);
+    tensor.at<float_t>(0,1,1,1) = float_t(2.0);
+
+    // check data using .at() accessor
+
+    for (cnn_size_t i = 0; i < 2; ++i) {
+        for (cnn_size_t j = 0; j < 2; ++j) {
+            EXPECT_EQ(tensor.at<float_t>(0,i,j,0), float_t(1.0));
+        }
+    }
     
     for (cnn_size_t i = 0; i < 2; ++i) {
         for (cnn_size_t j = 0; j < 2; ++j) {
-            for (cnn_size_t k = 0; k < 1; ++k) {
-                if (k == 0) {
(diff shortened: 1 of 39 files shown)


beru:
@edgarriba 
I checked the Tensor class.
I'm not sure why Tensor class has b,w,h,d template parameters since they should be dynamic and not compile-time fixed-size. Can you please tell me the reason?
What about adding getter methods to return batch, width, height, depth?

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+    typedef std::unique_ptr<std::vector<U> > data_ptr;
+    
+    // Returns the tensor shape
+    std::vector<cnn_size_t> shape() const { return shape_; }
+    
+    // Returns the value of a specified index in the tensor.
+    // Checked version (throw exceptions for out-of-range error)
+    template<typename T>
+    T& at(const cnn_size_t batch,
+          const cnn_size_t width,
```
beru: I think `width` parameter should be renamed to `x`.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+          const cnn_size_t height,
```
beru: I think `height` argument should be renamed to `y`.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+          const cnn_size_t depth) {
+        return *access_data<T>(batch, width, height, depth);
+    }
+
+    template<typename T>
+    const T& at(const cnn_size_t batch,
+                const cnn_size_t width,
```
beru: In my opinion, `width` parameter should be renamed to `x`.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+                const cnn_size_t height,
```
beru: From my perspective, `height` parameter should be renamed to `y` in here.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+        return *access_data<T>(batch, width, height, depth);
+    }
+
+    // Returns the pointer to a specified index in the tensor
+    // Checked version (throw exceptions for out-of-range error)
+    template<typename T>
+    T* ptr(const cnn_size_t batch,
+           const cnn_size_t width,
+           const cnn_size_t height,
+           const cnn_size_t depth) {
```
beru: I propose to rename argument names `width` to `x` and `height` to `y` in here.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+        return access_data<T>(batch, width, height, depth);
+    }
+    
+    template<typename T>
+    const T* ptr(const cnn_size_t batch,
+                 const cnn_size_t width,
+                 const cnn_size_t height,
```
beru: If you rename `width` to `x` and `height` to `y`, all the problems in the world will not be solved for sure.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+                 const cnn_size_t depth) const {
+        return access_data<T>(batch, width, height, depth);
+    }
+
+    // zero-overhead version (same performance to raw pointer access.
+    // have an assertion for out-of-range error)
+    template<typename T>
+    T& operator[] (cnn_size_t index) {
```
beru: Please add assert statement to check index parameter.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+        return *access_data<T>(index);
+    }
+
+    template<typename T>
+    const T& operator[] (cnn_size_t index) const {
```
beru: Please add assert statement to check `index` parameter.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+       shape_[2] = height; shape_[3] = depth;
+    }
+
+    // Method to access to the tensor data.
+    // It checks if the requested position is feasible or not.
+    template<typename T>
+    T* access_data(const cnn_size_t batch,
+                   const cnn_size_t width,
+                   const cnn_size_t height,
+                   const cnn_size_t depth) const {
```
beru: Renaming `width` to `x` and `height` to `y` would be good.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+        if (batch  > shape_[0] || width > shape_[1] ||
+            height > shape_[2] || depth > shape_[3]) {
```
beru: instead of using `shape_[magic_number_index]`, what about using getter methods such as `batch()`, `width()`, `height()`, `depth()`.

```
if (batch >= batch()
    || x >= width()
    || y >= height()
    || depth >= depth()
) {
    nn_error("Fool! You have just ensured the doom of this world.");
}
```

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+	fromDevice(device_);
+        return static_cast<T>(*access_data(0));
+    }*/
+
+ private:
+    // Initializes the data buffer with zeroes
+    void init_data(const cnn_size_t batch,  const cnn_size_t width,
+                   const cnn_size_t height, const cnn_size_t depth) {
+        host_data_ = data_ptr(new std::vector<U>(
+            batch * width * height * depth, float_t(0.0)));
```
edgarriba: what do you mean here? probably this method could be public and  implemented a resize routine

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    // Initializes the shape vector
+    void init_shape(const cnn_size_t batch,  const cnn_size_t width,
+                    const cnn_size_t height, const cnn_size_t depth) {
+       shape_[0] = batch;  shape_[1] = width;
+       shape_[2] = height; shape_[3] = depth;
+    }
```
edgarriba: how would you do it?

tiny_dnn/core/framework/tensor.h
edgarriba: batch or N? width or w or x? ...

edgarriba:
@beru 
regarding template parameters you are right, I was just playing with templates :-)

Besides, what do you think about the [data container] (https://github.com/edgarriba/tiny-cnn/blob/tensor/tiny_dnn/core/framework/tensor.h#L236) ? E.g. in Caffe2 they use `void*` to store it maybe to have a more general case? https://github.com/caffe2/caffe2/blob/master/caffe2/core/tensor.h#L538

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,281 @@
...
+
+    // Method to access to the tensor data.
+    // It checks if the requested position is feasible or not.
+    template<typename T>
+    T* access_data(const cnn_size_t d0,
+                   const cnn_size_t d1,
+                   const cnn_size_t d2,
+                   const cnn_size_t d3) const {
+        if (d0 > shape_[0] || d1 > shape_[1] ||
+            d2 > shape_[2] || d3 > shape_[3]) {
```
beru: I think comparison operators should be `>=` instead of `>`.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,281 @@
...
+        for (auto d : shape_) { new_size *= d; }
+        return new_size;
+    }
+
+ private:
+    // Initializes the data buffer with zeroes
+    void resize(const U value = 0) {
+        if (!host_data_) {
+            host_data_ = std::unique_ptr<std::vector<U> >(
+                new std::vector<U>(size(), value));
```
beru: Probablly, you can use `std::make_unique` in here.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,281 @@
...
+    explicit Tensor(const cnn_size_t d0,
+                    const cnn_size_t d1,
+                    const cnn_size_t d2,
+                    const cnn_size_t d3) {
+        reshape(d0, d1, d2, d3);
+        resize();
+    }
+
+    // Returns the tensor shape
+    std::vector<cnn_size_t> shape() const { return shape_; }
```
beru: I guess `const std::vector<cnn_size_t>& shape() const { return shape_; }` would be better.
But I'm wondering type of `shape_` needs to be std::vector container..
An alternative type is `std::array<cnn_size_t, 4>`.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,281 @@
...
+    
+    // Returns the value of a specified index in the tensor.
+    // Checked version (throw exceptions for out-of-range error)
+    template<typename T>
+    T& at(const cnn_size_t d0,
+          const cnn_size_t d1,
+          const cnn_size_t d2,
+          const cnn_size_t d3) {
```
beru: Are const type qualifiers for arguments really absolutely necessary?

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,281 @@
...
+        // the specified during the tensor initilization
+        // we cast the type.
+        if (!std::is_same<T,U>::value) {
+            return reinterpret_cast<T*>(value);
+        }
+        return value;
+    }
+
+    template<typename T>
+    T* access_data(const size_t index) const {
```
beru: Can this method be const method if it returns no-const pointer?

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,281 @@
...
+    template<typename T>
+    const T* ptr(const cnn_size_t d0,
+                 const cnn_size_t d1,
+                 const cnn_size_t d2,
+                 const cnn_size_t d3) const {
+        return access_data<T>(d0, d1, d2, d3);
+    }
+
+    // zero-overhead version (same performance to raw pointer access.
+    // have an assertion for out-of-range error)
```
beru: Here it says this method is zero-overhead version but this method uses `T* access_data(const size_t index) const` method, and in there it uses if statement to check index position and uses vector::at method.

beru:
> edgarriba wrote:
> Besides, what do you think about the data container ? E.g. in Caffe2 they use void* to store it maybe to have a more general case? https://github.com/caffe2/caffe2/blob/master/caffe2/core/tensor.h#L538

@edgarriba 
Right now, tiny_dnn library's calculation data type (`float_t`) is uniform.
I wish it will be able to handle arbtrary types of data.
But I wonder... Which came first, the chicken or the egg?

beru:
@edgarriba 
It would be better if this multidimentional array container would allocate memory aligned to cache lines.
The size of the cache lines on x86/x64 is 64 bytes. So the starting address should be multiple of 64. {{Citation needed}}

And you should probablly add allocator template parameter just like any other STL container classes.

bhack:
@beru See 3.4 section at https://github.com/kokkos/array_ref/blob/4098367b2f666fbcab4df0114450c8e93977a103/proposals/P0331.rst

edgarriba:
@beru right, I missed [this](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/util.h#L69) thx!

@bhack what do you mean with that? no alignment needed or what?

tiny_dnn/core/framework/tensor.h
edgarriba: if we return here a reference it means that a user could "reshape" the tensor just modifying this vector but we also need to resize and copy data in case that we want a real reshape. Do we want this feature now?

tiny_dnn/core/framework/tensor.h
edgarriba: @beru not really needed since we are passing arguments by value. However, for sanity and readability/maintenance I think it's a good practice to mark non-mutable arguments.

tiny_dnn/core/framework/tensor.h
edgarriba: @beru you are right. thx!

tiny_dnn/core/framework/tensor.h
beru: If you return **const** reference, caller cannot modify the returned vector without using const_cast.
Besides, if you return vector container instance by value, another heap allocation and copy occurs inside (without return value optimization).

tiny_dnn/core/framework/tensor.h
beru: Maybe I'm missing something but it seems to me that type of return value of this method is `T*`, I simply thought it should be `const T*`. And you also need to make non-const version of this method.
```
template<typename T>
T* access_data(const size_t index) {
    // almost same impl..
}
```

beru:
@bhack Thank you for introducing the document. I suppose the array_ref library is the exact kind of library we'd like to use in future. It has a lot of features that mekes it very flexible.

I really wish the Tensor class will have similar features such as extracting subarrays of an array.

One thing I noticed about current implementation of the Tensor class is it holds data using vector container that means this class has ownership of array data. But then, I suppose the class cannot have array reference feature. Is it intentionally designed?

bhack:
@beru As you see [R3](https://github.com/kokkos/array_ref/blob/master/proposals/P0009.rst) in the dscription section point to P0367 implemented at https://github.com/keryell/accessors

edgarriba:
@beru regarding the vector container ownership I assume (and as I remember from an old discussion with @nyanp) that container itself will not be shared, only tensor instances.

@bhack @nyanp @pansk @beru  IMO, the basics are there. Probably we need to rethink about `reshape()` and `resize()` routines, also if we want any `fill()` method, more constructors, etc. But I would say that during the integration to the pipeline we'll what are the real needs fir this Tensor class.

Give me some feedback in order to merge.

tiny_dnn/core/framework/tensor.h
beru: In access_data method, I've confirmed vector::at is replaced with operator [ ] but it still uses if statement. If you use non-compile-time constant value in if statement, compiler has to emit branch instruction so it has a little overhead to execute (even in release build). So you should use assert statement (this will be removed in release build) to check index.

Before
```
        if (index >= host_data_->size()) {
            throw nn_error("Access tensor out of range.");
        }
```

After
```
        assert(index < host_data_->size());
```

And there is another if statement.
```
        if (!std::is_same<T,U>::value) {
            return reinterpret_cast<T*>(value);
        }
```

This if statement will probablly be removed in release build. If not, you may need to use C++ technique like template meta programming to optimize it away. Then code will start to get messy like bxxst.

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,466 @@
...
+     */
+    Tensor<U> div(const Tensor<>& src) const {
+        if (this->shape() != src.shape()) {
+            throw nn_error("Tensor must have same shape");
+        }
+
+        Tensor<U> res(src.shape());
+
+        for_i(true, res.size(), [&](size_t i) {
+            res[i] = this->operator[](i) / (src[i] + 1e-10);
```
Randl: Are you trying to avoid division by 0 via adding 1e-10? But division by 1e-10 will result in same zero division. Also for very small numbers (rare case, but still) 1e10 might be significant. You might want to check for zero and add `std::numeric_limits<float_t>::min()`

edgarriba: oh, thx!

tiny_dnn/core/framework/tensor.h
edgarriba: @beru what about using `std::enable_if` ?

test/test_tensor.h
```diff
@@ -0,0 +1,547 @@
...
+    t2.fill(float_t(2.0));
+
+    // sum tensor along axis 0
+
+    Tensor<float_t> t3 = t1.div(t2);
+
+    // check that sum is okay
+
+    for (size_t i = 0; i < t3.size(); ++i) {
+        EXPECT_EQ(t3[i], float_t(0.5));
```
Randl: You should use `EXPECT_NEAR` hear to avoid fp related problems

edgarriba: ok thx. Do you think that all assertions could be affected by this?

Randl: Theoretically, any check where numbers are acquired in two different ways could be affected (e.g. 0.5==0.5 should be OK, while 0.5==1.0/2.0 might be a problem). IMHO, just replaced everything instead of thinking of each case.

tiny_dnn/core/framework/tensor.h
Randl: @edgarriba you fixed one problem out of two. To avoid division by zero you should explicitly check if a number isn't zero. I think you should add assert on equality to zero to not hurt performance

bhack: See alternative subsection  in Es. 105 of https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md

bhack: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-zero

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,506 @@
...
+        resize();
+    }
+
+    explicit Tensor(const std::vector<cnn_size_t>& shape) {
+        reshape(shape[0], shape[1], shape[2], shape[3]);
+        resize();
+    }
+
+    // Move constructor
+    Tensor(Tensor<U>&& other) = default;
```
Randl: VS2013 compiler doesn't support defaulted move constructors. https://stackoverflow.com/questions/24573963/move-constructor-invalid-type-for-defaulted-constructor-vs-2013

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,506 @@
...
+     * It is assumed that a tensor will hold data in NxWxHxD order,
+     * where:
+     *  N the batch axis
+     *  W the width axis
+     *  H the heigth axis
+     *  D the depth axis
+     *
+     *  Data will be hold by a std::vector with 64bytes alignment.
+     */
+    explicit Tensor(const cnn_size_t d0,
```
edgarriba: ok, but before that: are we going to support tensors serialization? (could be cool this feature :D)  If so, I think that `cnn_size_t` a.k.a `std::uint32_t` should be okay. Otherwise, I'll move to standard `size_t`.

Randl:
Looks like VS somewhere tries to copy `unique_ptr`:

```
C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xstddef(316): error C2280: 'std::unique_ptr<std::vector<tiny_dnn::float_t,tiny_dnn::aligned_allocator<tiny_dnn::float_t,0x040>>,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)' : attempting to reference a deleted function [C:\projects\build\test\tiny_dnn_test.vcxproj]          with          [              _Ty=std::vector<tiny_dnn::float_t,tiny_dnn::aligned_allocator<tiny_dnn::float_t,0x040>>          ] (C:\projects\tiny-dnn\test\test.cpp)          C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\memory(1486) : see declaration of 'std::unique_ptr<std::vector<tiny_dnn::float_t,tiny_dnn::aligned_allocator<tiny_dnn::float_t,0x040>>,std::default_delete<_Ty>>::unique_ptr'          with          [              _Ty=std::vector<tiny_dnn::float_t,tiny_dnn::aligned_allocator<tiny_dnn::float_t,0x040>>          ]          This diagnostic occurred in the compiler generated function 'tiny_dnn::Tensor<tiny_dnn::float_t>::Tensor(const tiny_dnn::Tensor<tiny_dnn::float_t> &)'
```

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,516 @@
...
+        if (this->shape() != src.shape()) {
+            throw nn_error("Tensor must have same shape");
+        }
+
+        Tensor<U> res(src.shape());
+
+        for_i(true, res.size(), [&](size_t i) {
+            const U tmp = src[i];
+            res[i] = tmp == U(0.0) ? std::numeric_limits<U>::quiet_NaN() :
+                this->operator[](i) / (tmp + std::numeric_limits<U>::min());
```
Randl: Now you don't need addition

edgarriba: > Looks like VS somewhere tries to copy unique_ptr:

maybe because the destructor is missing ? I don't have a MS environment to test :(

> Now you don't need addition

thx!

Randl:
on first glance the problem is that VS tries to copy unique pointer in copy constructor

tiny_dnn/core/framework/tensor.h
pansk: Adding a destructor might not suffice. If you add a destructor, copy constructor and copy assignment are still generated (for backward compatibility with pre-C++11).

See here: http://stackoverflow.com/questions/3734247/what-are-all-the-member-functions-created-by-compiler-for-a-class-does-that-hap

In your case, if unique_ptr is a member, when a copy (constructor|assignment) is generated, it will invoke the copy (constructor|assignment) from unique_ptr, resulting in an error (since unique_optr is uncopiable).

The only way is to implement your custom copy constructor and copy assignment, or delete them if you want to make your class uncopiable.

bhack: Take a look athttps://katyscode.wordpress.com/2012/10/04/c11-using-stdunique_ptr-as-a-class-member-initialization-move-semantics-and-custom-deleters/

Randl:
Custom copy constructor seems necessary anyway. @edgarriba @bhack @nyanp

bhack:
@Randl see https://accu.org/index.php/journals/1896

Randl:
@bhack Correct me if I'm wrong, but enforcing rule of zero together with `unique_ptr` results in  non-copyable class, and that's not what we want

bhack:
Yes exactly: "By using a unique_ptr we make our class non-copyable/assignable and identical in behavior to the previous examples."
And we need to avoid this here.

Randl:
You can still make a custom copy constructor which will work the `unique_ptr`. That's what I suggested

bhack:
Yes was a documentation enforcing this. And also http://stackoverflow.com/questions/29124592/copy-constructor-to-transfer-ownership-of-a-unique-ptr

bhack:
/off topic/ @keryell Any news from the United Nations conference on "Bringing About HPC Open-Standards World Peace" ;) ?

bhack:
/off topic/ @keryell Any news from the United Nations conference on "Bringing About HPC Open-Standards World Peace" ;)

bhack:
/off topic/ @keryell Any news from the United Nations conference on "Bringing About HPC Open-Standards World Peace" ;)

edgarriba:
I close this. You can find the valid pull request in https://github.com/tiny-dnn/tiny-dnn/issues/41

bhack:
The new tensor PR is https://github.com/tiny-dnn/tiny-dnn/pull/411

beru:
> bhack wrote:
> @beru As you see R3 in the description section point to P0367 implemented at https://github.com/keryell/accessors

@bhack Thank you for specifying the proposal for C++ standard.
Even if it can survive under a fluid situation and will be adopted into the standard, it will may take a few years so I guess we need to build or find different one.

tiny_dnn/core/framework/tensor.h
beru: @edgarriba I guess your are talking about using function/method tempaltes overload with SFINAE rule. I didn't know about `std::enable_if`. Thanks!

I wish future C++ will have compile-time `static_if` feature though. Metaprogramming should be more easier to be written and read.

keryell:
@beru: array_ref was well received. Probably for C++20, but anyway there is a special department in the C++ committee dedicated to bike-sheding, so it will be probably slightly different... :-)

@bhack : stop asking off-topic questions all over GitHub threads :-) But yes, everybody survived to this SC16 panel...

bhack:
@keryell We are living in a too covered tecnological roadmaps era that create fragmentation. HPC standard domain it is only part of this trend. Probably after C++20 all this discussions will be so useless like discussing an if statement. :)

PR #401, opened 2016-11-13
add a workaround to mitigate memory consumption problem

beru:
This is a temporary workaround to mitigate memory footprint problem.
By skipping calling wrapper function `tiny_dnn::network::fprop(const std::vector<vec_t>& in)`, a number of copies of STL vector container is reduced.

To confirm the effect from this change, I used `examples/benchmarks/main.cpp` program and compare memory usages at head of `tiny_dnn::sequential::forward(const std::vector<tensor_t>& first)`.

I expect ongoing Tensor class should greatly reduce the problem.


edgarriba:
@beru nice! thx!!

Issue #399, opened 2016-11-12
New name for cnn_size_t


Randl:
As explained [here](https://github.com/tiny-dnn/tiny-dnn/pull/397#discussion_r87693123), `cnn_size_t` is a type used by serializer, and it should be reasonable size to keep model file small. If we need a big enough type compared to the memory space, we still need to use `size_t`, not `cnn_size_t`.

The naming is quite misleading, so I suppose to rename `cnn_size_t`. The name should

- not be confusing and mixed up with `size_t`
- be expressive, i.e. be connected to serialization in some way.

Options that were supposed are

1. `serial_size_t`.  Somewhat long, but seems OK for me
2.  `ssize_t`. Not good imho. Too similar with `size_t` and not expressive.
3. `serial_t`. Nice. Short, expressive and easy distinguished
4. `cnn_serial_t`. Longer, but mentions `cnn`.
5. Anything else?

edgarriba:
I vote for 3

beru:
I vote for 1.
3 and 4 can be misleading to serialized_type (data type? size type?).

bhack:
I don't mention cnn prefix if we are in progress to bet not only a cnn. We could also use cereal_type it is a little bit less future prof but it is clear

nyanp:
@Randl 
Nice suggestions :)

I vote for 1. A bit long, but it's clear. ```ssize_t``` is already used in some systems (ex. [sys/types.h](http://pubs.opengroup.org/onlinepubs/7908799/xsh/systypes.h.html)), and I don't think it's good to expose the name of 3rd-party library.

edgarriba:
`cnn_size_t` is a typdef of [`uint32_t`](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/config.h#L99) , why not just renamed to `uint_t`?

PR #407, opened 2016-11-15
Rename cnn_size_t to serial_size_t

Randl:
As was discussed in https://github.com/tiny-dnn/tiny-dnn/issues/399 `cnn_size_t` should be renamed.

### DIFF
--- a/docs/developer_guides/Adding-a-new-layer.md
+++ b/docs/developer_guides/Adding-a-new-layer.md
@@ -83,7 +83,7 @@ std::vector<shape3d> out_shape() const override {
 Execute forward calculation in this method.
 
 ```cpp
-void forward_propagation(cnn_size_t worker_index,
+void forward_propagation(serial_size_t worker_index,
                          const std::vector<vec_t*>& in_data,
                          std::vector<vec_t*>& out_data) override {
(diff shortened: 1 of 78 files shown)


nyanp:
@Randl 
LGTM 👍 

@beru and @pansk pointed out on gitter that there's an excessive use of cnn_size_t. Since we can just use ```size_t``` instead of ```sereal_size_t``` where serialization isn't the matter, I'll do further improvement after this PR is merged. Thanks!

pansk:
Is the serialization going to be an external project (dnn-serialize)?
If so, I think we can safely change all occurrences of serial_size_t back to size_t, and let the external lib to do any conversion.

edgarriba:
@pansk maybe it's a good idea to have it as an external project since after introducing it to the pipeline the compilation time increased a lot and it's really annoying.

pansk:
I have two reasons why an external serialization project would be a good idea:
- If we can do serialization outside the library it means that the library is fully configurable from outside (and, so, is complete)
- if we need to introduce additional dependencies just for serialization (e.g. protobuf, some json serialization library, boost::ptree or anything else), we can keep the main library "clean".

PR #408, opened 2016-11-15
Replace catches with EXPECT_THROW

Randl:


edgarriba:
@Randl good catch! thx!

PR #411, opened 2016-11-17
Tensor (reprise)

pansk:
Well, as usual, I made a mess with GIT, and apparently I can't easily push to @edgarriba's original PR.

Today I reworked a bit @edgarriba's #400, I changed a bit the interface, added the lazy-allocation and lazy-movement of memory around, renamed accessor to host_ptr and host_at (to clarify they work on host memory), and implemented the generic functions for binary and unary host operations, element-wise and scalar, so it's now easier to implement functions such as add, sub, mul, div, exp, sqrt, ...

I commented out linspace, if someone has a strong opinion about that, we can still get it back in.

### DIFF
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -82,6 +82,8 @@ class Tensor {
         resize();
     }
 
+    // Move constructor
+    Tensor(Tensor<U>&& other) = default;
 
     // Returns the tensor shape
     const std::vector<cnn_size_t>& shape() const { return shape_; }
@@ -193,98 +195,98 @@ class Tensor {
 
     /* @brief Element-wise addition
      */
-    Tensor<> add(const Tensor<>& src) const {
+    Tensor<U> add(const Tensor<U>& src) const {
         Tensor<U> res(src.shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) + src[i];
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise addition
      */
-    Tensor<> add(const float_t scalar) const {
+    Tensor<U> add(const float_t scalar) const {
         Tensor<U> res(this->shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) + scalar;
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise subtraction
      */
-    Tensor<> sub(const Tensor<>& src) const {
+    Tensor<U> sub(const Tensor<U>& src) const {
         Tensor<U> res(src.shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) - src[i];
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise subtraction
      */
-    Tensor<> sub(const float_t scalar) const {
+    Tensor<U> sub(const float_t scalar) const {
         Tensor<U> res(this->shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) - scalar;
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise multiplication
      */
-    Tensor<> mul(const Tensor<>& src) const {
+    Tensor<U> mul(const Tensor<U>& src) const {
         Tensor<U> res(src.shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) * src[i];
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise multiplication
      */
-    Tensor<> mul(const float_t scalar) const {
+    Tensor<U> mul(const float_t scalar) const {
         Tensor<U> res(this->shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) * scalar;
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise division
      */
-    Tensor<> div(const Tensor<>& src) const {
+    Tensor<U> div(const Tensor<>& src) const {
         Tensor<U> res(src.shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) / (src[i] + 1e-10);
         });
 
-        return res;
+        return std::move(res);
     }
 
     /* @brief Element-wise division
      */
-    Tensor<> div(const float_t scalar) const {
+    Tensor<U> div(const float_t scalar) const {
         Tensor<U> res(this->shape());
 
         for_i(true, res.size(), [&](size_t i) {
             res[i] = this->operator[](i) / (scalar + 1e-10);
         });
 
-        return res;
+        return std::move(res);
     }
 
  private:
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -108,14 +108,12 @@ class Tensor {
 
     // zero-overhead version (same performance to raw pointer access.
     // have an assertion for out-of-range error)
-    template<typename T>
-    T& operator[] (const size_t index) {
-        return *access_data<T>(index);
+    U& operator[] (const size_t index) {
+        return *access_data<U>(index);
     }
 
-    template<typename T>
-    const T& operator[] (const size_t index) const {
-        return *access_data<T>(index);
+    const U& operator[] (const size_t index) const {
+        return *access_data<U>(index);
     }
 
     // this is only a proof of concept to copy data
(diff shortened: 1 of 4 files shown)


beru:
@pansk I have some questions.

- Talking about `add`, `sub`, `mul`, `div`, `sqrt`, `exp` methods, do they need to be implemented as Tensor class's methods? Is it on purpose so that they are chainable?
- In `binary_scalar_operation`, `unary_element_wise_operation`, `binary_element_wise_operation` methods implementation, Tensor instance is created and returned but that means just calling the methods would result abusing heap manager (inevitable memory allocation happens everytime you call the methods) and that's not suitable for embedding computing. Those methods may be convenient but not very efficent from my view.
- What about Tensor Iterator class? Personally speaking, I don't utilize iterators and algorithms very well but if Tensor Iterator will not be implemented, I fear someday hardcore C++ programmers may visit your home with pitchforks and torches in their hands.

edgarriba:
@beru what do you propose to avoid inevitable memory allocation when calling methods?

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,481 @@
...
+        for_i(true, res.size(), [dst, src1, &f](size_t i) {
+            dst[i] = f(src1[i]);
+        });
+
+        return std::move(res);
+    }
+
+    template<typename F> Tensor binary_scalar_operation(U scalar, F f) const
+    {
+        if (this->shape() != src.shape()) {
```
beru: Where is src variable in this `binary_scalar_operation` method?
Unlike binary_element_wise_operation method, this doesn't have src argument.

beru:
// edgarriba wrote:
// @beru what do you propose to avoid inevitable memory allocation when calling methods?

@edgarriba What about passing `res` Tensor class instance variable by an arguement?
But I wonder if all the values in Tensor class should be targeted in any circumstances.
What if tiny-dnn library code wants to partially edit elements (such as only 1 batch data) inside Tensor class instance? I suppose current Tensor class public interface does not readily provide those functionalities.

edgarriba:
@beru maybe it's a stupid question but what's the difference between declaring a Tensor before the call or inside? Correct me if I'm wrong but with this simple ops I think that we can assume that all values should be targeted since are element-wise operations. At this first stage I don't think we have to handle the span concept, but indeed it's necessary!

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,481 @@
...
+            shape_[1] * shape_[2] * d3 +
+            shape_[1] * d2 +
+            d1
+            );
+    }
+
+    // zero-overhead version (same performance to raw pointer access.
+    // have an assertion for out-of-range error)
+    U& operator[] (const size_t index) {
+        return mutable_host_data()[index];
```
beru: There is a comment `zero-overhead`, but is it really so?
Inside `mutable_host_data`, I see `if` statement so I thought the performance of this method is slower than raw pointer access.

beru:
> edgarriba wrote:
> what's the difference between declaring a Tensor before the call or inside?

@edgarriba 
By readying Tensor instance outside of the methods, caller have opportunity to save number of memory allocations/deallocations.

If the methods declare Tensor instance inside, calling the methods mean memory allocation and deallocation happen frequently. If the size of data is small or operations are not frequently repeated, it may not be a big issue but there are situations when the data is like >= 300MB and the methods are called inside a looong loop.

Probablly you've already understood this but let me write this, using STL vector container is almost like using new/delete (malloc/free in C). It dynamically allocates heap memory.

Severely repeating heap memory allocation/deallocation may cause heap fragmentation issue if heap manager is not implemented well. Low-spec embedding systems tend to not have decent hardware/operating systems/libraries, so overly using dynamic heap memory can be very troublesome or should be avoided.

Besides, memory allocation/deallocation/copy slows down your program. They are very handy but it comes with performance penalty.

edgarriba:
@beru  thanks for this excellent explanation!
Some possibilities that come to my mind are the following:

- Keep same signature and modify current Tensor state (removed const)

```cpp
Tensor add(const Tensor& src) {
     const U* dst  = mutable_host_data();
     const U* src1 = host_data();
     const U* src2 = src.host_data();

     for_i(true, res.size(), [dst, src1, src2, &f](size_t i) {
         dst[i] = f(src1[i], src2[i]);
     });
     return *this
}
```
But what if we want to overload `operator +()` to behave like this?
Tensor t2 will be modified and cannot be reused for other operations.
```cpp
Tensor<float_t> t1(2,2,2,2);
Tensor<float_t> t2(2,2,2,2);

Tensor<float_t> t3 = t1 + t2;
```

- Get rid off `add()` from  `Tensor` structure and have it as standalone methods
````cpp
void add(const Tensor& src1, const Tensor& src2, Tensor& dst) {
     // Do we need to force users to resize Tensor dst before the call?
}
```

So, before going further to this discussion: what features do we want to support for `Tensor` structure?

edgarriba:
@pansk @beru @bhack @nyanp Another feature that most frameworks have is Tensor reshape. Does it make sense to have tensors with dynamic size/shape in tiny-dnn?

bhack:
Something like http://www.boost.org/doc/libs/1_62_0/libs/multi_array/doc/user.html#sec_reshape?

edgarriba:
exactly, but again, I don't know if there's any use case for this feature right know in the pipeline and the same for this basic operations.

bhack:
See how reshape is introduced in a basic pipeline like MNIST https://www.tensorflow.org/versions/r0.8/tutorials/mnist/pros/index.html

edgarriba:
sure, but I see it more as a pre-preprocessing operation rather that something really needed during the network computation, no?

bhack:
I think that [Caffe2](https://github.com/caffe2/caffe2/search?p=1&q=Reshape&utf8=✓) and [Keras](https://github.com/fchollet/keras/search?q=reshape&type=Code&utf8=✓) uses of reshape could give a fair complete use cases overview. 

For span have you seen 1.2 section of https://github.com/kokkos/array_ref/blob/master/proposals/P0009.rst?

edgarriba:
caffe2 uses reshape to resize the vector containing the tensor shape (note: I assume they support N dimensional tensors) but then I'm not sure what will happen to a tensor that already contains data and you trigger reshape method.

bhack:
See https://github.com/caffe2/caffe2/blob/master/caffe2/core/tensor.h#L239

tiny_dnn/core/framework/tensor.h
```diff
@@ -0,0 +1,481 @@
...
+        for_i(true, res.size(), [dst, src1, src2, &f](size_t i) {
+            dst[i] = f(src1[i], src2[i]);
+        });
+
+        return std::move(res);
+    }
+
+    template<typename F> Tensor unary_element_wise_operation(F f) const
+    {
+        if (this->shape() != src.shape()) {
```
pansk: Oh sh... It's supposed to read shape().

tiny_dnn/core/framework/tensor.h
pansk: See above.

tiny_dnn/core/framework/tensor.h
pansk: Indeed, it's no longer zero overhead - my fault, I didn't remove the comment._ I left the functionality here, but I would rather get rid of this accessor.

pansk:
@beru I agree with you with your considerations about add, sub, ... methods. They shouldn't be in the class, I just left them where they were, but (as discussed with @edgarriba) they would better be outside the class, and probably have a completely different design.

pansk:
@edgarriba, @bhack I'm wouldn't implement reshape in a first implementation, I think this decision is quite related to how you implement its connection to the graph structure.

pansk:
@beru instead of iterators, I'd rather like using slicing (a la valarray). Actually, when discussing with @edgarriba I also suggested we should have a look at the possibility of replacing the vector with a (possibly gsliced) valarray.

pansk:
@edgarriba: for implementing the syntax `Tensor<float_t> t3 = t1 + t2;` the standard workaround is to build the expression template returning dummy templated objects, and then perform the computation in an implicit converting assignment or converting construcor.
So, t1+t2 returns e.g. a tensor_sum<decltype(t1), decltype(t2)>, and the computation is triggered by constructing a Tensor taking with a tensor_sum.

beru:
> @beru instead of iterators, I'd rather like using slicing (a la valarray). Actually, when discussing with @edgarriba I also suggested we should have a look at the possibility of replacing the vector with a (possibly gsliced) valarray.

I didn't know about std::valaray, And after reading [this](http://stackoverflow.com/questions/1602451/c-valarray-vs-vector) stackoverflow page, I think we should stick with std::vector or raw memory.

Talking aboute implementing Iterator, I was just kidding. But it seems like current Tensor class does not provide end() like position of inner data. So it can be a bit troublesome.
```
const U* end() const {
    return mutable_host_data() + shape_[0] * shape_[1] * shape_[2] * shape_[3];
}
U* end() {
    return mutable_host_data() + shape_[0] * shape_[1] * shape_[2] * shape_[3];
}
```

nyanp:
Great works :)
@pansk Could you rebase from master?

pansk:
Today I'll try to rebase the PR.

tiny_dnn/core/framework/tensor.h
pansk: Fixed.

tiny_dnn/core/framework/tensor.h
pansk: I removed the operators in next commit.

pansk:
Aparently my rebase is a mess, but it should be rebased now.

bhack:
Tests are failing

edgarriba:
@pansk to fix duplicated commits from rebase:
1. Save your work
2. `git pull --rebase upstream master`
3. `git push -f origin tensor`

pansk:
Hope it's ok now...

edgarriba:
yes!

bhack:
Also tests passed

nyanp:
LGTM :)

edgarriba:
@nyanp @pansk  the file history was removed :scream:

PR #416, opened 2016-11-21
add NNPackInitializer, update NNPack backend, fix broken tests and warnings

edgarriba:
- add NNPackInitializer as a singleton in order to have a global variable to check if nnpack is initialized.
- update NNPack codes: found a bug in inference -> replaced by output. /cc @azsane 
- fix maxpool and fully connected broken tests
- fix some warnings

### DIFF
--- a/test/test_caffe_converter.h
+++ b/test/test_caffe_converter.h
@@ -81,7 +81,7 @@ TEST(caffe_converter, rectangle_input) {
     auto model = create_net_from_json(json);
 
     // conv->pool->conv->pool->fc->relu->fc->softmax
-    ASSERT_EQ(model->depth(), 2);
+    ASSERT_EQ(model->depth(), size_t(2));
 
     EXPECT_EQ((*model)[0]->in_shape()[0], shape3d(24, 40, 1));
     EXPECT_EQ((*model)[0]->out_shape()[0], shape3d(22, 38, 96));
@@ -270,7 +270,7 @@ TEST(caffe_converter, lenet) {
     auto model = create_net_from_json(json, shape3d(28, 28, 1));
 
     // conv->pool->conv->pool->fc->relu->fc->softmax
-    ASSERT_EQ(model->depth(), 8);
+    ASSERT_EQ(model->depth(), size_t(8));
 
     // conv1 28x28x1 -> 24x24x20
     EXPECT_EQ((*model)[0]->in_shape()[0], shape3d(28, 28, 1));   // in: 28x28x1
@@ -442,7 +442,7 @@ TEST(caffe_converter, lenet_v1) {
     auto model = create_net_from_json(json, shape3d(28, 28, 1));
 
     // conv->pool->conv->pool->fc->relu->fc->softmax
-    ASSERT_EQ(model->depth(), 8);
+    ASSERT_EQ(model->depth(), size_t(8));
 
     // conv1 28x28x1 -> 24x24x20
     EXPECT_EQ((*model)[0]->in_shape()[0], shape3d(28, 28, 1));   // in: 28x28x1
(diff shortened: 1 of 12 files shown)


bhack:
If we need this singleton thread safe we requite [VS 2015](https://msdn.microsoft.com/en-au/library/hh567368.aspx#concurrencytable)

azsane:
@edgarriba inference has a bug? Can you provide more detail?

And I just digged into output function found that 
```
enum nnp_status nnp_convolution_output(
	enum nnp_convolution_algorithm algorithm,
	size_t batch_size,
	size_t input_channels,
	size_t output_channels,
	struct nnp_size input_size,
	struct nnp_padding input_padding,
	struct nnp_size kernel_size,
	const float input[],
	const float kernel[],
	const float bias[],
	float output[],
	pthreadpool_t threadpool,
	struct nnp_profile* profile);
```

implementation
```
const struct nnp_size output_size = {
		.width = input_padding.left + input_size.width + input_padding.right - kernel_size.width + 1,
		.height = input_padding.top + input_size.height + input_padding.bottom - kernel_size.height + 1
	};
```
there is no *output_sampling* (or stride) provided  so it will compute incorrect value when conv has stride more than 1

edgarriba:
@azsane Are you sure that your tests pass? I'm getting a segfault with nnp_convolution_inference

http://pastebin.com/ua9gMuYy

@Maratyszcza Did you notice similar error before?

azsane:
@edgarriba 
Yes ,all 30 tests were passed.May be machine or enviroment related?

since nnp_kwt8x8_3x3_and_stream.avx2 cause SIGSEGV,a simple way can bypass this
change algorithm in conv2d_op_nnpack.h (I ran test with algorithm auto not the one changed below)
```
inline nnp_convolution_algorithm nnp_algorithm() {
    return nnp_convolution_algorithm_auto; //change this to nnp_convolution_algorithm_implicit_gemm
}
``` 

this algorithm doesn't use any avx2 related function and I use it in a converted caffemodel for googlenet.

The reason why I use the algorithm ,it's that on some machines auto one was quite slow than others(expecially on kernel size 3*3).Then it's unbelievable that gemm is fast than auto!

You can change algorithm to gemm and perform origin tests to see if it's correct.

edgarriba:
@azsane I've tried with that flag and I end up at this exception: 
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_op_nnpack.h#L124

with a status error == 15 : https://github.com/Maratyszcza/NNPACK/blob/master/include/nnpack.h#L44

azsane:
@edgarriba 
error code =15 ?
that's nnp_status_invalid_pooling_stride returned by validate_pooling_arguments

But this function is called by pool related function.Is conv test or pool test report this error ?

edgarriba:
I'm trying this: ./test/tiny_dnn_test --gtest_filter=convolutional.fprop_nnp

Maratyszcza:
@edgarriba `found a bug in inference -> replaced by output.` What is the bug in inference?

edgarriba:
@Maratyszcza I'm not sure if it's a bug or what it is but as I mention before I got the following:
    -  `nn_convolution_inference`:
              `nnp_convolution_algorithm_implicit_auto` -> segfault
              `nnp_convolution_algorithm_implicit_gemm` -> status error == 15 (related pool error)
    -  `nn_convolution_output`:
              `nnp_convolution_algorithm_implicit_auto` -> OK
              `nnp_convolution_algorithm_implicit_gemm` -> need to check yet

Maratyszcza:
@edgarriba I checked the code, and it is impossible: `nnp_convolution_inference` never returns `nnp_status_invalid_pooling_stride`. Only pooling functions return this error code.

edgarriba:
@Maratyszcza now is okay, solved.
My bad ... I had an older version .so file installed. Thanks anyway.

PR #418, opened 2016-11-22
add Tensor class

edgarriba:
This pull requests pretends to substitute the current [`tensor_t`](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/util.h#L71) by a full OOP Tensor class.

This is subsequent work from https://github.com/tiny-dnn/tiny-dnn/pull/319 and https://github.com/tiny-dnn/tiny-dnn/pull/411.

### DIFF
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -65,8 +65,8 @@ using namespace tiny_dnn::activation;
 #include "test_caffe_converter.h"
 #endif
 
+#include "test_tensor.h"
 #include "test_image.h"
-
 
 int main(int argc, char **argv) {
     ::testing::InitGoogleTest(&argc, argv);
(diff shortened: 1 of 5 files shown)


pansk:
In my version, when I merged the change from cnn_size_t to tensor_size_t (which reflects in your commit 0012555) I also fixed the alignment of parameters from all the methods of class Template.
It's quite confusing having the parameters aligned with the code, you won't see when the parameter list end and when the code begins, I rather have them in a single line (exceeding line length), or having them aligned with the first parameter. 
Can you bring in that version?

edgarriba:
@pansk but now it's okay no?
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/framework/tensor.h#L86-L89

pansk:
Oh, I see.
That commit was modified by me during the merge, and I only did a partial work, and solved with the next commit.

PR #419, opened 2016-11-23
Subtle speed optimization

beru:
Achieved about 1% speed gain by adjusting 1st parameter (`bool parallelize`) of `for_i` function in `parallel_for.h` file. When the size of data to process is not big enough, suppress parallelization to reduce thread creation overhead.

The downsides of this change
- code gets difficult to read (can be difficult to debug too?)
- may not be a good idea when thread creation is very cheap (thread pooling.. Intel TBB not tested)
- execution flow is not consistent

Here's the outputs of example_mnist_train project release builds to compare the performance.
[before.txt](https://github.com/tiny-dnn/tiny-dnn/files/608955/before.txt)
[after.txt](https://github.com/tiny-dnn/tiny-dnn/files/608956/after.txt)

### DIFF
--- a/test/test_core.h
+++ b/test/test_core.h
@@ -69,11 +69,11 @@ device_t device_type(size_t *platform, size_t *device) {
 }
 
 #define TINY_DNN_GET_DEVICE_AND_PLATFORM       \
-    int cl_platform = 0, cl_device = 0; \
+    size_t cl_platform = 0, cl_device = 0; \
     device_t device = device_type(&cl_platform, &cl_device);
 #else
 #define TINY_DNN_GET_DEVICE_AND_PLATFORM       \
-    int cl_platform = 0, cl_device = 0; \
+    size_t cl_platform = 0, cl_device = 0; \
     device_t device = device_t::NONE;
 #endif  // defined(USE_OPENCL) || defined(USE_CUDA)
 
--- a/test/test_core.h
+++ b/test/test_core.h
@@ -69,11 +69,11 @@ device_t device_type(size_t *platform, size_t *device) {
 }
 
 #define TINY_DNN_GET_DEVICE_AND_PLATFORM       \
-    size_t cl_platform = 0, cl_device = 0; \
+    int cl_platform = 0, cl_device = 0; \
     device_t device = device_type(&cl_platform, &cl_device);
 #else
 #define TINY_DNN_GET_DEVICE_AND_PLATFORM       \
-    size_t cl_platform = 0, cl_device = 0; \
+    int cl_platform = 0, cl_device = 0; \
     device_t device = device_t::NONE;
 #endif  // defined(USE_OPENCL) || defined(USE_CUDA)
 
--- a/tiny_dnn/layers/layer.h
+++ b/tiny_dnn/layers/layer.h
@@ -538,6 +538,7 @@ class layer : public node {
                 std::transform(diff.begin(), diff.end(),
                                diff.begin(), [&](float_t x) { // NOLINT
                                   return x * rcp_batch_size; });
+                // parallelize only when target size is big enough to mitigate thread spawning overhead
                 bool parallelize = (target.size() >= 512);
                 o->update(diff, target, parallelize);
             }
--- a/tiny_dnn/layers/layer.h
+++ b/tiny_dnn/layers/layer.h
@@ -530,16 +530,16 @@ class layer : public node {
 
     void update_weight(optimizer *o, serial_size_t batch_size) {
         float_t rcp_batch_size = float_t(1) / float_t(batch_size);
+        vec_t diff;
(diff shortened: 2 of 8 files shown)


beru:
Achieved another 10% ish speed gain by tuning up iterative call of activation function.
Here's the output to compare. [after2.txt](https://github.com/tiny-dnn/tiny-dnn/files/609170/after2.txt)

bhack:
What code are you running?

beru:
@bhack I compiled and ran `examples/mnist/train.cpp` program.

bhack:
And with what engine? Internal?

beru:
@bhack I modified around layer and optimization function a little bit, so this PR isn't related to any backend engines.

tiny_dnn/layers/layer.h
```diff
@@ -530,16 +530,16 @@ class layer : public node {
...
             if (trainable() && is_trainable_weight(in_type_[i])) {
-                vec_t diff;
                 vec_t& target = *get_weight_data(i);
-
                 ith_in_node(i)->merge_grads(&diff);
                 std::transform(diff.begin(), diff.end(),
                                diff.begin(), [&](float_t x) { // NOLINT
                                   return x * rcp_batch_size; });
-                o->update(diff, target);
+                bool parallelize = (target.size() >= 512);
```
edgarriba: @beru what's the reason for this? Could you add a short comment to the source?

beru: I just added below comment in recent commit.
```
// parallelize only when target size is big enough to mitigate thread spawning overhead
bool parallelize = (target.size() >= 512);
o->update(diff, target, parallelize);
```
I added `bool parallelize` parameter to `optimizer::update` method and its descendants.
The newly added parameter is used for 1st parameter of `for_i` function (that has triple argument prototype).  `for_i` function can selectively create multi-threads but doing so is not favourable when size of data is small.

bhack:
@beru Are you sure that something like https://github.com/Maratyszcza/pthreadpool/ that we already used trough upstream cannot ammortise many of this costs?

test/test_core.h
```diff
@@ -69,11 +69,11 @@ device_t device_type(size_t *platform, size_t *device) {
 }
 
 #define TINY_DNN_GET_DEVICE_AND_PLATFORM       \
-    size_t cl_platform = 0, cl_device = 0; \
+    int cl_platform = 0, cl_device = 0; \
```
beru: I've just reverted the change in here. I'll send another PR for this in future.

The reason why I changed the type : 2nd and 3rd parameters of `Device::Device` in `tiny_dnn/core/framework/device.fwd.h` is `int`, maybe I should have used `size_t` -> `int` casts.

beru:
> bhack wrote:
> @beru Are you sure that something like https://github.com/Maratyszcza/pthreadpool/ that we already used trough upstream cannot ammortise many of this costs?

I still haven't extensively checked the performance difference on many platforms, I only tested it on Win10 64bit, YMMV.

bhack:
But what code path are you testing? This one https://github.com/tiny-dnn/tiny-dnn/blob/5693b92803add28d6b6dea8f4265f27925b0cbb2/tiny_dnn/util/parallel_for.h#L108?

patrikhuber:
@bhack The thread pool you linked looks horrible. Are we going C-style library now?

Have a look at this in comparison - proper modern C++: https://github.com/progschj/ThreadPool

bhack:
No I don't want to use that. I mean that our upstream NNPACK is using that internally and I talked about ThreadPool cause I don't know if we are handling in this optimizzation only the c++11 thread backend of our parallel_for

beru:
> bhack wrote:
> But what code path are you testing? This one https://github.com/tiny-dnn/tiny-dnn/blob/5693b92803add28d6b6dea8f4265f27925b0cbb2/tiny_dnn/util/parallel_for.h#L108?

@bhack Yes, that one. Here's simplified call stack.
```
parallel_for
for_
for_
for_i
kernels::conv2d_op_avx
Conv2dOp::compute
convolutional_layer
layer::forward
sequential::forward
network::fprop
network::train_onebatch
network::train_once
network::fit
network::train
train_lenet
main
```

beru:
> bhack wrote:
> @beru Are you sure that something like https://github.com/Maratyszcza/pthreadpool/ that we already used trough upstream cannot ammortised many of this costs?

@bhack I'm not so sure about that. I can only guess without having and measuring actual implementations.
We can realy on profilers but I'd like to add mechanism to profile the code so that we can easily see percentages of execution time.

edgarriba:
@beru can I merge this PR?

beru:
@edgarriba It's up to you. I think b0db424 is much effective. So, just merging it would be enough for a while. Do you agree with the idea?

edgarriba:
@beru sure

PR #424, opened 2016-11-27
add source documentation for layer class

edgarriba:


nyanp:
@edgarriba
Nice, is it ok to merge?

edgarriba:
there's still a lot of documentation to write but for now we can merge

tiny_dnn/layers/layer.h
```diff
@@ -67,9 +67,11 @@ class layer : public node {
     virtual ~layer() = default;
 
     /**
-     * construct N-input, M-output layer
+     * @brief Defaul layer constructor that instantiates a N-input, M-output layer
```
patrikhuber: typo in "Defaul", and it should be "an N-input" I think

PR #425, opened 2016-11-27
Switch to variable dimensions for tensor

Randl:
Tensors should be not only 4D, as mentioned in gitter. This is my draft attempt to switch to N-dimensional tensors.
cc @nyanp  @bhack @beru @edgarriba

### DIFF
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -481,47 +481,47 @@ namespace details {
     }
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_add(Tensor<TD> &dst, TS1 src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_add(Tensor<TD, kDim> &dst, TS1 src1, const Tensor<TS2, kDim> &src2) {
     binary_scalar_tensor_operation(dst, src1, src2, details::plus<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_add(Tensor<TD> &dst, const Tensor<TS1> &src1, TS2 src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_add(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, TS2 src2) {
     binary_tensor_scalar_operation(dst, src1, src2, details::plus<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_add(Tensor<TD> &dst, const Tensor<TS1> &src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_add(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, const Tensor<TS2, kDim> &src2) {
     binary_tensor_tensor_elementwise_operation(dst, src1, src2, details::plus<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_sub(Tensor<TD> &dst, TS1 src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_sub(Tensor<TD, kDim> &dst, TS1 src1, const Tensor<TS2, kDim> &src2) {
     binary_scalar_tensor_operation(dst, src1, src2, details::minus<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_sub(Tensor<TD> &dst, const Tensor<TS1> &src1, TS2 src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_sub(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, TS2 src2) {
     binary_tensor_scalar_operation(dst, src1, src2, details::minus<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_sub(Tensor<TD> &dst, const Tensor<TS1> &src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_sub(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, const Tensor<TS2, kDim> &src2) {
     binary_tensor_tensor_elementwise_operation(dst, src1, src2, details::minus<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_mul(Tensor<TD> &dst, TS1 src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_mul(Tensor<TD, kDim> &dst, TS1 src1, const Tensor<TS2, kDim> &src2) {
     binary_scalar_tensor_operation(dst, src1, src2, details::multiplies<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_mul(Tensor<TD> &dst, const Tensor<TS1> &src1, TS2 src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_mul(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, TS2 src2) {
     binary_tensor_scalar_operation(dst, src1, src2, details::multiplies<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_mul(Tensor<TD> &dst, const Tensor<TS1> &src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_mul(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, const Tensor<TS2, kDim> &src2) {
     binary_tensor_tensor_elementwise_operation(dst, src1, src2, details::multiplies<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_div(Tensor<TD> &dst, TS1 src1, const Tensor<TS2> &src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_div(Tensor<TD, kDim> &dst, TS1 src1, const Tensor<TS2, kDim> &src2) {
     binary_scalar_tensor_operation(dst, src1, src2, details::divides_checked<TS1, TS2>);
 }
 
-template<typename TD, typename TS1, typename TS2> void layer_div(Tensor<TD> &dst, const Tensor<TS1> &src1, TS2 src2) {
+template<typename TD, typename TS1, typename TS2, size_t kDim> void layer_div(Tensor<TD, kDim> &dst, const Tensor<TS1, kDim> &src1, TS2 src2) {
     if (src2 == TS2(0.0)) {
         dst.reshape(src1.shape());
         dst.fill(std::numeric_limits<TD>::quiet_NaN());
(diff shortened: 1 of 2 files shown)


beru:
It would be great if it will have subscript operator methods that return low dimention TensorRef.
```
TensorRef<U, kDimensions - 1, /* is_const*/true> operator [] (int idx) const;
TensorRef<U, kDimensions - 1, /* is_const*/false> operator [] (int idx);
```
And maybe we need to add TensorBase interface class as well.

beru:
Though this post isn't directly related to the purpose of this PR, I thought below utility functions should be moved to another header file.
```
binary_tensor_tensor_elementwise_operation
unary_tensor_elementwise_operation
binary_tensor_scalar_operation
binary_scalar_tensor_operation
layer_add
layer_sub
layer_mul
layer_div
```

edgarriba:
@beru You are right. A good start could trying to implement as separate operations `add_op_internal.h`, `sub_op_internal.h`, and so on
https://github.com/tiny-dnn/tiny-dnn/tree/master/tiny_dnn/core/kernels

beru:
@edgarriba I thought about moving utility functions to another single header file such as `tensor_utils.h`

edgarriba:
as a first step yes, but ideally we should be able to have simple operations. 
That was my initial idea

beru:
@edgarriba add/sub/mul/div Aren't these operations too primitive to divide its implementation to individual files?

edgarriba:
yes, but if they can be implemented in standalone operations in a future we could use them to compose other complex operations at the same time to be considered as a node of the computational graph which maybe can let us to use auto-differentiation. That's how TF works and that's also the reason why I used this OpKernel pattern

bhack:
Also from OpenVx point of view

>Application can build an OpenVX graph that represents Convolution Network topologies where the layers are represented as OpenVX nodes (vx_node) and the vx_tensor as the data objects connecting the nodes (layers) of the OpenVX graph (Convolution Network).

beru:
@edgarriba I see... but chaining a lot of simple operations means longer pipelines and frequent data load/store. Without graph nodes optimization feature, I guess that will result in poor processing performance.

bhack:
I.e. Nvidia tensorrt and tensorrt 2 are fusing kernels https://devblogs.nvidia.com/parallelforall/production-deep-learning-nvidia-gpu-inference-engine/

edgarriba:
@Randl Good in general, please fix minor issues.

Besides, this is going to be one of the main basis of the library and has to be well documented. Please add as much documentation as possible: general ideas about shape, axis, general usage, etc.

We can use this as a reference of decent documentation
https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/mat.hpp#L740

Randl:
@edgarriba Minor issues are?

tiny_dnn/core/framework/tensor.h
```diff
@@ -148,58 +150,57 @@ class Tensor {
...
-                const size_t d3) {
-        if (d0 >= shape_[0] || d1 >= shape_[1] ||
-            d2 >= shape_[2] || d3 >= shape_[3]) {
-            throw nn_error("Access tensor out of range.");
+    // Returns the pointer to a specified index in the tensor
+    // Checked version (throw exceptions for out-of-range error)
+    template<typename... Args>
+    const U* host_ptr (const Args... args) const {
+        //  TODO(Randl): Should we throw here if lower number of arguments?
+        if (sizeof...(args) != shape_.size())  {
```
beru: If I understand correctly, the number of variadic template parameters in here should be equal to kDimensions.

Using `static_assert` does not incur runtime cost but then `host_ptr` method will be able to throw an exception.

```
static_assert(sizeof...(args) == kDimensions, "number of parameters must be... y'know");
static_assert(check<std::is_convertible, int, Args...>, "!");
```

tiny_dnn/core/framework/tensor.h
```diff
@@ -148,58 +150,57 @@ class Tensor {
...
-    const U* host_ptr(const size_t d0,
-                      const size_t d1,
-                      const size_t d2,
-                      const size_t d3) const {
-        if (d0 >= shape_[0] || d1 >= shape_[1] ||
-            d2 >= shape_[2] || d3 >= shape_[3]) {
+    template<typename... Args>
+    size_t host_pos(const size_t d,
+                const Args... args) {
+        size_t dim = shape_.size() - sizeof...(args) - 1;
```
beru: What if `sizeof...(args) >= kDimentions`? I suppose `dim` will be pretty high, and it is used for indexing `shape_` array.

beru:
> bhack wrote:
> I.e. Nvidia tensorrt and tensorrt 2 are fusing kernels https://devblogs.nvidia.com/parallelforall/production-deep-learning-nvidia-gpu-inference-engine/

@bhack Thank you for the link. Always very insightful.
But I don't think the same thing can be achieved with current design of Tensor utility function. As each of utility template functions are instantiated and linked together at compile time. So I suppose they can't be fused in meaningful ways.

Anyways... I guess I need to check TensorFlow code to learn its design.

bhack:
@beru If you are interested this is how Mx component approach handle fusion: http://dmlc.ml/2016/11/21/fusion-and-runtime-compilation-for-nnvm-and-tinyflow.html

beru:
@bhack @edgarriba IMO, It's much better directly using below functions with lambda function or sth instead of calling multiple layer_add/sub/mul/div functions.
```
binary_tensor_tensor_elementwise_operation
unary_tensor_elementwise_operation
binary_tensor_scalar_operation
binary_scalar_tensor_operation
```

Repeating multiple loops is unfavourable and not necessary since the same math operation can be achieved in single loop with user defined function argument. We should be careful about iterating memory. It's not very cheap when size of data is huge.

tiny_dnn/core/framework/tensor.h
Randl: @beru as for now we assume this is checked earlier. Not sure it's the best way to do, but we want to avoid too much checks too.

tiny_dnn/core/framework/tensor.h
Randl: @edgarriba @beru Unchecked version sounds useful, just need to decide about naming. I believe there was unchecked version in some version of PR, but can't recall exactly. Anyway, that's probably out of scope of PR.

tiny_dnn/core/framework/tensor.h
beru: @Randl  I can understand that if `host_pos` is private method but it's public. So its parameter should be checked.
And static_assert only takes place at compile time, so there's no need to worry about runtime cost.

Randl: @beru  @edgarriba then we probably should remove checks to the lower level, i.e. `host_pos`, and create separate set of non checked functions I think

edgarriba: +1 to use static_assert

tiny_dnn/core/framework/tensor.h
```diff
@@ -147,59 +147,88 @@ class Tensor {
...
-        return mutable_host_data() + (
-            shape_[1] * shape_[2] * shape_[3] * d0 +
-            shape_[1] * shape_[2] * d3 +
-            shape_[1] * d2 +
-            d1
-            );
+    template<typename... Args>
+    U* host_ptr(const Args... args) {
+        if (sizeof...(args) != kDimensions)  {
+            throw nn_error("Wrong number of dimensions");
```
edgarriba: @Randl use static assert. Better check error during compilation time than runtime

tiny_dnn/core/framework/tensor.h
```diff
@@ -294,10 +323,10 @@ class Tensor {
      * shape_[2]: height
      * shape_[3]: depth
      */
```
edgarriba: update this documentation

tiny_dnn/core/framework/tensor.h
```diff
@@ -147,59 +147,88 @@ class Tensor {
...
+
+    /**
+     * Checked version (throw exceptions for out-of-range error)
+     * @param args index of rest (k-1) dimensions.
+     * @return the pointer to a specified index in the tensor
+     */
+    template<typename... Args>
+    const U* host_ptr (const Args... args) const {
+        if (sizeof...(args) != kDimensions)  {
+            throw nn_error("Wrong number of dimensions");
```
edgarriba: add static_assert

tiny_dnn/core/framework/tensor.h
```diff
@@ -245,9 +274,9 @@ class Tensor {
         shape_[2] = d2;
         shape_[3] = d3;
         host_data_.resize(calcSize(), U(0));
```
edgarriba: remove this commented code if it's not used anymore

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,42 +61,41 @@
...
+     * It is assumed that a tensor will hold data in C-style nD array, i.e
+     * row-major order:  the rightmost index “varies the fastest”.
      *
      *  Data will be hold by a std::vector with 64bytes alignment.
      */
-    explicit Tensor(const size_t d0,
-                    const size_t d1,
-                    const size_t d2,
-                    const size_t d3) {
+    /*explicit Tensor(const size_t d0) {
```
edgarriba: remove commented code

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,42 +61,41 @@
...
         reshape(d0, d1, d2, d3);
-    }
+    }*/ //TODO(Randl): variadic template version
 
-    explicit Tensor(const std::array<size_t, 4>& shape) {
-        reshape(shape[0], shape[1], shape[2], shape[3]);
+    explicit Tensor(const std::array<size_t, kDimensions>& shape) {
```
edgarriba: could you write a couple of lines and put an example how to use this constructors?

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,42 +61,41 @@
...
     }
 
     explicit Tensor(const std::vector<size_t>& shape) {
-        assert(shape.size() == 4);
-        reshape(shape[0], shape[1], shape[2], shape[3]);
+        reshape(shape);
+    }
+
+    explicit Tensor(std::initializer_list<size_t> const& shape)  {
+        assert(shape.size() == kDimensions);
```
edgarriba: maybe an static assert here too?

Also quick documentation and usage example

Randl: @edgarriba I think shape.size() isn't known at compile time. Not sure about it though. Gotta check.

tiny_dnn/core/framework/tensor.h
```diff
@@ -147,59 +147,88 @@ class Tensor {
...
+    const std::array<size_t, kDimensions>& shape() const { return shape_; }
+
+    /**
+     * Checked version of access to indexes in tensor (throw exceptions
+     * for out-of-range error)
+     * @param args indexes in tensor
+     * @return the value of a specified index in the tensor
+     */
+    template<typename... Args>
+    U& host_at(const Args... args) {
```
Randl: @edgarriba Do we need non const verison, if it returns value?

edgarriba: the motivation behind this is for testing so that you can modify at runtime values

tiny_dnn/core/framework/tensor.h
Randl: @edgarriba Yeah, `static_assert()` doesn't work here?

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,42 +61,58 @@
...
+     * row-major order:  the rightmost index “varies the fastest”.
      *
      *  Data will be hold by a std::vector with 64bytes alignment.
      */
-    explicit Tensor(const size_t d0,
-                    const size_t d1,
-                    const size_t d2,
-                    const size_t d3) {
-        reshape(d0, d1, d2, d3);
-    }
```
edgarriba: you removed code but comments still there :)

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,42 +61,58 @@
...
-     * where:
-     *  N the batch axis
-     *  W the width axis
-     *  H the heigth axis
-     *  D the depth axis
+     * It is assumed that a tensor will hold data in C-style nD array, i.e
+     * row-major order:  the rightmost index “varies the fastest”.
      *
      *  Data will be hold by a std::vector with 64bytes alignment.
      */
```
edgarriba: you removed code but comments still there :)

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,42 +61,58 @@
...
-    explicit Tensor(const std::array<size_t, 4>& shape) {
-        reshape(shape[0], shape[1], shape[2], shape[3]);
+    //TODO(Randl): variadic template version of reshape
+
+    /**
+     * Constructor that assepts an array of shape and create a Tensor with that
+     * shape
+     * @param shape array containing N integers, sizes of dimensions
+     * @return
+     */
```
edgarriba: could you add a line with usage for different constructors?

edgarriba:
:+1:

PR #437, opened 2016-12-01
Tensor division into TensorStorage and Tensor (TensorView)

Randl:


tiny_dnn/core/framework/tensor.h
```diff
@@ -329,18 +375,10 @@ class Tensor {
...
-
-#if defined(USE_OPENCL) || defined(USE_CUDA)
-    /* Pointer to the Tensor data in the device */
-    std::unique_ptr<CLCudaAPI::Buffer<U> > device_data_;
-#endif
-    mutable bool data_is_on_host_;    //< current data is on host if true, on device if false.
-    mutable bool data_dirty_;         //< set to true if current data might have been modified
+    /* Offset from the beginningng of TensorStorage */
+    size_t offset;
+    std::shared_ptr<TensorStorage> host_data_;
```
edgarriba: to avoid confusions rename this as `storage_` or sth similar

tiny_dnn/core/framework/tensor.h
```diff
@@ -128,16 +128,22 @@ class TensorStorage {
...
     /**
      * Initializes an empty tensor.
      * @return
      */
     Tensor() {
-        host_data_ =
+        storage_pointer_ =
             std::make_shared<std::vector<U, aligned_allocator<U, 64>>>();
-        host_data_iter_ = host_data_->begin();
+        storage_pointer_iter_ = host_data_->begin();
```
pansk: host_data_ -> storage_pointer_? (similar cases follow)

tiny_dnn/core/framework/tensor.h
```diff
@@ -128,16 +128,22 @@ class TensorStorage {
  */
 template<typename U = float_t, size_t kDimensions = 4, bool kConst = false>
```
pansk: Adding the constness as a template parameter is a bit confusing for me.

Can't the uses of kConst be changed into is_const_v<U>? (Or U be forwarded to the underlying TensorStorage, to make it type-safe?)

tiny_dnn/core/framework/tensor.h
```diff
@@ -62,6 +63,61 @@
 
 namespace tiny_dnn {
 
+
+template<typename U = float_t>
+class TensorStorage {
+ public:
+    void toDevice() const {
```
pansk: I'd keep toDevice and fromDevice private, and move the host_data, mutable_host_data, device_data and mutable_device_data into here. I think this class should be responsible for moving the data around (in the future, maybe even only part of it).

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,44 +63,115 @@
...
+ public:
 
     /**
      * Initializes an empty tensor.
      * @return
      */
     Tensor() {
+        storage_pointer_ =
+            std::make_shared<std::vector<U, aligned_allocator<U, 64>>>();
+        storage_pointer_iter_ = host_data_->begin();
```
edgarriba: Where is `storage_pointer_iter_` defined?

edgarriba: Is it correct to access to a member of a member like this?

tiny_dnn/core/framework/tensor.h
```diff
@@ -109,19 +182,22 @@ class Tensor {
      * @param shape array containing N integers, sizes of dimensions
      * @return
      */
-    explicit Tensor(std::initializer_list<size_t> const& shape)  {
+    explicit Tensor(std::initializer_list<size_t> const &shape) {
+        storage_pointer_ =
+            std::make_shared<std::vector<U, aligned_allocator<U, 64>>>();
+        host_data_iter_ = host_data_->begin();
```
edgarriba: Where is `host_data_iter_` defined?

tiny_dnn/core/framework/tensor.h
```diff
@@ -264,83 +347,49 @@ class Tensor {
...
-    std::vector<U, aligned_allocator<U, 64>> host_data_;
-
-#if defined(USE_OPENCL) || defined(USE_CUDA)
-    /* Pointer to the Tensor data in the device */
-    std::unique_ptr<CLCudaAPI::Buffer<U> > device_data_;
-#endif
-    mutable bool data_is_on_host_;    //< current data is on host if true, on device if false.
-    mutable bool data_dirty_;         //< set to true if current data might have been modified
+    /* Offset from the beginningng of TensorStorage */
+    size_t offset;
```
edgarriba: For what is `offset` used?

pansk: I think the rationale is to use "offset" for indexing subtensors.
E.G., if I have a 2x2x2 tensor, and I want to index the second in the first dimension, the subtensor will be a tensor class sharing the same storage, but with offset 4 (the size of the subtensor to be skipped to get to the element in the storage).

tiny_dnn/core/framework/tensor.h
Randl: I'm not sure I get you right. 
My idea is, that Tensor can be forced to be constant at compile time, i.e. incapable to modify underlying TensorStorage (surely mutable).

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
...
+     * @return
+     */
+    explicit TensorStorage(std::initializer_list<size_t> const &shape) {
+        std::vector<size_t> tmp(shape.size());
+        std::copy(shape.begin(), shape.end(), tmp.begin());
+        resize(tmp);
+    }
+
+    void sync() {
+        if (data_is_on_host_)
```
edgarriba: with this we are double checking `data_is_on_host_`

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
...
+     * shape. For example, given shape = {2,3,4,5,6}, tensor
+     * will be of size 2x3x4x5x6
      * @param shape array containing N integers, sizes of dimensions
      * @return
      */
-    explicit Tensor(const std::array<size_t, kDimensions>& shape) {
-        reshape(shape);
+    explicit Tensor(const std::array<size_t, kDimensions> &shape) {
+        storage_pointer_ =
+            std::make_shared<std::vector<U, aligned_allocator<U, 64>>>(shape);
```
edgarriba: why you don't initialize as TensorStorage?

tiny_dnn/core/framework/tensor.h
```diff
@@ -264,83 +405,61 @@ class Tensor {
...
     /**
      * A tensor holds data in C-style nD array, i.e row-major order:
      * the rightmost index “varies the fastest”.
      */
     std::array<size_t, kDimensions> shape_;
 
-    /* Pointer to the Tensor data in pure in the host device */
-    std::vector<U, aligned_allocator<U, 64>> host_data_;
+    /* Offset from the beginning of TensorStorage */
+    size_t offset;
```
edgarriba: rename to `offset_`

tiny_dnn/core/framework/tensor.h
```diff
@@ -264,83 +405,61 @@ class Tensor {
...
+        static_assert(!kConst, "Non-constant operation on constant Tensor");
         toDevice();
         data_dirty_ = true;
         return (*device_data_)();
     }
 #endif
 
     size_t size() const {
-        return host_data_.size();
+        return calcSize();
```
edgarriba: why not `storage_ptr_->size()` ? To avoid calculating each time
Where:
```cpp
size_t TensorStorage::size() const {
    return host_data_.size();
}
```

Randl: Size of tensor is not necessary size of underlying TensorStorage

edgarriba: ok, I see. I guess you mean when it's a View. Then we probably have a member variable `size_` that is calculated each time that `reshape()` is called.

Besides, show we also have a have variable saying whether a Tensor is a view?

Randl: I have defined it, but it's yet commented since requires changes in constructors. 
As for isView, we might want to check if Tensor is of size of TensorStorage, but that's all, I believe. So IMHO no need for member variable for that

edgarriba: you are right. Then a with a public method is enough

edgarriba:
just to keep in mind other solutions: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.h#L474

bhack:
Just to take a look to other adopted solutions https://github.com/tiny-dnn/tiny-dnn/pull/437

edgarriba:
@bhack wrong ref? What do you think about the constness?

bhack:
@edgarriba Yes wrong paste. Now it is corrected.

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
 
 namespace tiny_dnn {
 
+template<typename U = float_t>
```
beru: What about adding template parameter `Allocator`
```
template<
    typename U = float_t,
    typename Allocator = aligned_allocator<U, 64>
>
```

and the use it for member `host_data_`.
```
std::vector<U, Allocator> host_data_;
```

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
...
+
+    void sync() {
+        if (data_is_on_host_)
+            toDevice();
+        else
+            fromDevice();
+    }
+
+    typename std::vector<U, aligned_allocator < U, 64>>::
+    iterator host_data(size_t offset) {
```
beru: The iterator should be typedef-ed so that you don't need to use `std::vector<U, aligned_allocator < U, 64>>` again and again.

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
...
+        return host_data_.begin() + offset;
+    }
+
+    typename std::vector<U, aligned_allocator < U, 64>>::
+    const_iterator host_data(size_t offset) const {
```
beru: The iterator should be typedef-ed.

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
...
+
+    std::vector<U, aligned_allocator<U, 64>> host_data_;
+
+#if defined(USE_OPENCL) || defined(USE_CUDA)
+    /* Pointer to the Tensor data in the device */
+    std::unique_ptr<CLCudaAPI::Buffer<U>> device_data_;
+#endif
+    mutable bool data_is_on_host_; // is current data is on host?
+    mutable bool data_dirty_;      // have current data might been modified?
+    mutable size_t dirty_from, dirty_to; // range of indexes that can be modified
```
beru: You violate 80 characters per line rule, edgarriba will hunt us down.

tiny_dnn/core/framework/tensor.h
```diff
@@ -61,45 +63,173 @@
...
+    const_iterator host_data(size_t offset) const {
+        return host_data_.begin() + offset;
+    }
+
+    void resize(const std::vector<size_t> &sz) {
+        host_data_->resize(std::accumulate(std::begin(sz),
+                                           std::end(sz),
+                                           size_t(1),
+                                           std::multiplies<size_t>()), U(0));
+        //TODO(Randl): device data
```
beru: vector::resize with zero initialization is fine for most of the uses.
But please consider adding another resize method that uses below vector method to resize and initialize at once efficiently.
```
template <class InputIterator>
void assign (InputIterator first, InputIterator last);
```

and what about below methods?
```
begin
end
cbegin
cend
```

tiny_dnn/core/framework/tensor.h
Randl: Yeah, nice idea

tiny_dnn/core/framework/tensor.h
Randl: `host_data` is effectively `begin`/`cbegin` for given Tensor.

beru:
@Randl It's trivial but `std::accumulate` is used 8 times in a latest commited code.
```
std::accumulate(std::begin(sz),
                std::end(sz),
                size_t(1),
                std::multiplies<size_t>())
```
I think adding a static inline function/method would simplify the code.

tiny_dnn/core/framework/tensor.h
beru: I didn't know that comment lines can be longer than 80 characters. So you don't need to fix this one..
https://google.github.io/styleguide/cppguide.html#Line_Length

Hmmm I guess that means we can hide wide codes in comments and apply preprocessing tool to expand it before compilers strip out the comments.

tiny_dnn/core/framework/tensor.h
beru: As you've probably already known, cbegin and cend are for `auto` specifier.
I thought lack of absense of end/cend means users need to calculate offset for end position.

beru:
I'm not sure why member names follow `snake_case` mentioned in [here](https://github.com/tiny-dnn/tiny-dnn/blob/master/docs/developer_guides/How-to-contribute.md#preferred-coding-style) while class names don't. Rules are for dummies but rules are rules. :triangular_ruler: Maybe there's an exception like this one.

Randl:
@beru I'm trying to follow [Google style guide](https://google.github.io/styleguide/cppguide.html#Type_Names). I think it should be somehow mentioned in this report.

beru:
@Randl What about the rest of the code in the library? Will they be modified to match Google style guide? I can't wait 2 spaces indentation to come because of 80 characters per line rule.

Randl:
@beru You should talk to @edgarriba or @nyanp about that

beru:
@Randl Thank you for the suggestion. Let me talk to @edgarriba and @nyanp.
Baa.. Baa..

edgarriba:
@Randl thx for the contribution! :+1:

PR #436, opened 2016-12-01
Add Mxnet in ReadMe

beru:
excerpted from #432 which is written by Erwan BERNARD(edmBernard)

PR #438, opened 2016-12-03
transpose a table in README.md

beru:
Motivation for this PR comes from some comments in #436
The new layout can be checked at [here](https://github.com/beru/tiny-cnn/blob/comparison_with_other_libraries_table_transposed/README.md).


edgarriba:
@beru thx for the contribution! :+1:

PR #443, opened 2016-12-06
modification for old quantization functions and kernels

wangyida:
- [ ] rewrite old quantization functions and delete useless functions
- [ ] make fully-connected layer work under quantized mode
- [ ] make conv layer work under quantized mode
- [ ] remove unnecessary ops through networks


tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
```diff
@@ -39,66 +39,28 @@ inline void tiny_quantized_conv2d_kernel(const conv_params& params,
...
-    float_t max_input(in[0]);
-    for (serial_size_t inc = 0; inc < params.in.depth_; inc++) {
-        for (serial_size_t ins = 0; ins < params.in_padded.height_*params.in_padded.height_; ins++) {
-            serial_size_t idx = params.in_padded.get_index(0, 0, inc);
-            min_input = std::min(min_input, (&in[idx])[ins]);
-            max_input = std::max(max_input, (&in[idx])[ins]);
-        }
-    }
+    // Input quantization
+    vec_t range_in = tensor_range<float_t>(in);
```
wangyida: @edgarriba The range for each input and output tensors should also be stored in tensor to pass to the next layer in tiny-dnn, so do you have some good idea for this data passing strategy on codes? I think it's different from full precision kernels cause new tensors of ranges are introduced there

tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
```diff
@@ -135,25 +97,40 @@ inline void tiny_quantized_conv2d_kernel(const conv_params& params,
...
+
+      quantize_down_and_shrink_range<int32_t, uint8_t>(a_quantized,
+        range_output[0], range_output[1],
+        &(range_output_requantized[0]), &(range_output_requantized[1]),
+        &a_requantized);
+    }
+
+    // Transforming dequantize data in uint8_t to flaot,
+    // this could be removed within concatenated quantized network
+    a = quantized_tensor_to_float<uint8_t>(a_requantized,
```
wangyida: @bhack @edgarriba The range for each output tensors should also be stored in tensor to pass to the next layer in tiny-dnn, so do you have some good idea for this data passing strategy on codes? I think it's different from full precision kernels cause new tensors of ranges are introduced there as output, so this line of code is not necessary in an efficient quantization structure.
I want to only use a_requantized and range_output_requantized as output tensors.

tiny_dnn/core/kernels/tiny_quantized_fully_connected_kernel.h
```diff
@@ -40,106 +40,86 @@ inline void tiny_quantized_fully_connected_kernel(const fully_params& params,
...
-        float_to_quantized<int32_t>(0.0f, min_output_value, max_output_value);
+        float_to_quantized<int32_t>(0.0f, range_output[0], range_output[1]);
 
-    const int32_t offset_output = 0;
-    const int32_t mult_output = 1;
-    const int32_t shift_output = 0;
+    const int32_t offset_output = 0, mult_output = 1, shift_output = 0;
 
+    // Calculating inner product
     bool use_gemm = false;
```
wangyida: @beru Here you can modify the option for GEMM utilization

tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
edgarriba: maybe we can go for a inherited `QuantizedTensor` or if there is a real need, just add a small attributes to Tensor regarding data ranges.

tiny_dnn/core/kernels/tiny_quantized_fully_connected_kernel.h
beru: @wangyida I see. Thank you for the info. So, it looks like `gemmlowp` is currently off.

bhack:
What is the status of this? I really hope that we can have last Gsoc efforts in shape and mainlined.Opencv dnn module (not tiny-dnn based) [now has Enet](https://github.com/opencv/opencv_contrib/commit/e784f137f6143ab4844ab36818bdb262b6d197db) that was one of the orginal Gsoc goal in quantized version for mobile with an android example.

edgarriba:
ping @wangyida 

do you think you can finish this PR? we need to update the old backend structure

PR #444, opened 2016-12-06
tiny_dnn/util/math_functions.h: moments: Why 4th/5th argument is `vec_t *`?

yumetodo:
To avoid Null-Pointer-Dereference, It's recommended to use lvalue-reference instead.

### DIFF
--- a/tiny_dnn/layers/batch_normalization_layer.h
+++ b/tiny_dnn/layers/batch_normalization_layer.h
@@ -127,8 +127,8 @@ public:
             }
         }
  
-        moments(delta_dot_y, in_spatial_size_, in_channels_, mean_delta_dot_y, nullptr);
-        moments(curr_delta, in_spatial_size_, in_channels_, mean_delta, nullptr);
+        moments(delta_dot_y, in_spatial_size_, in_channels_, mean_delta_dot_y);
+        moments(curr_delta, in_spatial_size_, in_channels_, mean_delta);
         // if Y = (X-mean(X))/(sqrt(var(X)+eps)), then
         //
         // dE(Y)/dX =
@@ -159,7 +159,7 @@ public:
 
         if (phase_ == net_phase::train) {
             // calculate mean/variance from this batch in train phase
-            moments(*in_data[0], in_spatial_size_, in_channels_, mean, &variance);
+            moments(*in_data[0], in_spatial_size_, in_channels_, mean, variance);
         }
 
         // y = (x - mean) ./ sqrt(variance + eps)
--- a/tiny_dnn/layers/batch_normalization_layer.h
+++ b/tiny_dnn/layers/batch_normalization_layer.h
@@ -126,9 +126,9 @@ public:
                 delta_dot_y[i][j] *= curr_delta[i][j];
             }
         }
-        moments(delta_dot_y, in_spatial_size_, in_channels_, &mean_delta_dot_y, nullptr);
-        moments(curr_delta, in_spatial_size_, in_channels_, &mean_delta, nullptr);
  
+        moments(delta_dot_y, in_spatial_size_, in_channels_, mean_delta_dot_y, nullptr);
+        moments(curr_delta, in_spatial_size_, in_channels_, mean_delta, nullptr);
         // if Y = (X-mean(X))/(sqrt(var(X)+eps)), then
         //
         // dE(Y)/dX =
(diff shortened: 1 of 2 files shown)


nyanp:
@yumetodo 
Hi, thank you for your PR!
It's nice to use reference in local variables in ```forward_propagation```. Overloading ```moments``` is also great idea :)

We are usually using google's style guide, and using pointer as outputs is derived from that guide:
https://google.github.io/styleguide/cppguide.html#Reference_Arguments

So, it would be nice if you'll keep 4th/5th arguments as ```vec_t*``` (the rest is perfect for me).

yumetodo:
Thank you for your review.

---

>We are usually using google's style guide

Before post this, I read [Coording style](https://github.com/tiny-dnn/tiny-dnn/blob/master/docs/developer_guides/How-to-contribute.md#coding-guides).
However, It's seem that there is no mention of that.

---

>and using pointer as outputs is derived from that guide

Using poiner without null-check will cause Undefined Behavior, and [Undefined behavior can result in time travel](https://blogs.msdn.microsoft.com/oldnewthing/20140627-00/?p=633/).

So, to use pointer as argument, null-check will be required.
However, **there is small but unignorable cost to null-check**.
Using lvalue reference(not const-lvalue-reference, **so that we can easily understand this is output**) will solve that. 
Google coording style force us null-check cost.

nyanp:
@yumetodo 
> Before post this, I read Coording style.
> However, It's seem that there is no mention of that.

Sorry it's my fault. I'll edit the guide page.

> However, there is small but unignorable cost to null-check.
> Using lvalue reference(not const-lvalue-reference, so that we can easily understand this is output) will solve that.

Ok, it's reasonable for tiny-dnn because it is called by performance-critical method. I'll merge it as-is. Thanks!

tiny_dnn/layers/batch_normalization_layer.h
```diff
@@ -152,32 +152,25 @@ class batch_normalization_layer : public layer {
...
 
         // y = (x - mean) ./ sqrt(variance + eps)
-        calc_stddev(*variance);
+        calc_stddev(variance);
 
         for_i(parallelize_, in_data[0]->size(), [&](int i) {
             const float_t* inptr  = &in[i][0];
             float_t*       outptr = &out[i][0];
 
             for (serial_size_t j = 0; j < in_channels_; j++) {
```
edgarriba: could you rename all `serial_size_t` to `size_t` ?

tiny_dnn/util/math_functions.h
```diff
@@ -36,52 +37,56 @@ inline void vector_div(vec_t& x, float_t denom) {
...
         }
-
-        vector_div(*variance, std::max(1.0f, num_examples*spatial_dim-1.0f));
+        vector_div(variance, std::max(1.0f, num_examples*spatial_dim-1.0f));
     }
 }
+/**
+ * calculate mean/variance across channels
+ */
+inline void moments(const tensor_t& in, serial_size_t spatial_dim, serial_size_t channels, vec_t& mean) {
```
edgarriba: check that < 80 rule is followed. Same in above method

tiny_dnn/util/math_functions.h
```diff
@@ -36,52 +37,56 @@ inline void vector_div(vec_t& x, float_t denom) {
...
+}
+inline void moments(const tensor_t& in, serial_size_t spatial_dim, serial_size_t channels, vec_t& mean, vec_t& variance) {
+    const serial_size_t num_examples = static_cast<serial_size_t>(in.size());
+    assert(in[0].size() == spatial_dim * channels);
+
+    mean.resize(channels);
+    std::fill(mean.begin(), mean.end(), (float_t)0.0);
+    variance.resize(channels);
+    std::fill(variance.begin(), variance.end(), (float_t)0.0);
+    detail::moments_impl_calc_mean(num_examples, channels, spatial_dim, in, mean);
```
edgarriba: you can just reuse code here by calling moments to compute mean

edgarriba:
@yumetodo too late to for this PR but maybe you can open a new one with my comments

tiny_dnn/util/math_functions.h
yumetodo: According to my research, there are too maby ``< 80 rule`` violation in this project.
To solve that, change coding style or **use clang-format**.

ref: #453

beru: I wish the restriction will be relaxed to `< 100` or `< 120`.

Randl: @beru This is discussed in https://github.com/tiny-dnn/tiny-dnn/pull/453 . IMHO 2-space + 80 char restriction is OK, though is not best for modern displays, 2/4-space with 120 char is the best.

PR #450, opened 2016-12-10
implement view functionality

edgarriba:


tiny_dnn/core/framework/tensor.h
```diff
@@ -335,7 +318,36 @@ class Tensor {
...
+    Tensor view(std::initializer_list<size_t> const &start,
+                std::initializer_list<size_t> const &shape) {
+	if (start.size() > kDimensions || shape.size() > kDimensions) {
+            throw nn_error("Overpassed number of existing dimensions.");
+	}
+        return Tensor(storage_ptr_, compute_offset(start, shape_), shape);
+    }
+
+    bool isView() const {
+        return size_ != storage_ptr_->size();
```
Randl: You can check first if `offset!=0`, which is cheaper than call `size()``

tiny_dnn/core/framework/tensor.h
```diff
@@ -335,7 +318,36 @@ class Tensor {
                                                           shape_.end()));
     }
 
+    Tensor view(std::initializer_list<size_t> const &start,
```
Randl: Also, a default value of `start` equals to 0 might be nice

edgarriba: ok so then we have to switch arguments order

edgarriba: or maybe having an overloaded method?

tiny_dnn/core/framework/tensor.h
edgarriba: ok makes sense but silly question: is it really than cheaper?

Randl: @edgarriba only way to know for sure is to run benchmark

tiny_dnn/core/framework/tensor.h
Randl: Current order is kinda more intuitive, so I'd vote for overloaded method

tiny_dnn/core/framework/tensor.h
edgarriba: @Randl I have been checking and not all views have `offset != 0` since we can take a view starting from zero position

Randl: Right, you still need to check size. But given book check is significantly faster then vector size computation, this still usually will give a speedup.

edgarriba:
@Randl any idea about the win errors?

tiny_dnn/core/framework/tensor.h
```diff
@@ -335,7 +318,57 @@ class Tensor {
...
+
+    Tensor view_impl(std::initializer_list<size_t> const &start,
+                     std::initializer_list<size_t> const &new_shape) {
+	if (start.size() > kDimensions || new_shape.size() > kDimensions) {
+            throw nn_error("Overpassed number of existing dimensions.");
+	}
+
+	// compute the new offset and check that it's feasible to create
+	// the new view.
+	const size_t new_offset = compute_offset(start, shape_);
```
Randl: As far as I understand, new offset should be just `offset+product(start)`

test/test_tensor.h
```diff
@@ -129,6 +129,62 @@ TEST(tensor, check_bounds) {
...
+
+    Tensor<float_t, 4> t1({2,2,2,2});
+    Tensor<float_t, 4> t2 = t1.view({2,4});
+
+    for (size_t i = 0; i < 8; ++i) {
+        EXPECT_EQ(t1.host_data()[i], float_t(0.0));
+        EXPECT_EQ(t2.host_data()[i], float_t(0.0));
+    }
+
+    t2.fill(float_t(2.0));
```
Randl: Try to fill with different values to check not only that function works, but also that you refer to right element in Tensor. This, BTW, applies to other tests too.

tiny_dnn/core/framework/tensor.h
Randl: Ah, ok, I misunderstood the meaning of `start` and the way `compute_offset` works. Please add some explanation to the function.

Randl: Anyway you need to take old offset in account

tiny_dnn/core/framework/tensor.h
pansk: Oh. Actually my comment didn't come through (was still "pending"). Sorry, I'm catching up with this "modern" tools ;)
size for vectors is just a pointer difference, probably inlined, I wouldn't bother about speed here.

tiny_dnn/core/framework/tensor_storage.h
```diff
@@ -68,6 +68,20 @@ static inline size_t product(Container &c) {
                            std::multiplies<size_t>());
 }
 
+template <typename C1, typename C2>
+static inline size_t compute_offset(const C1 &c1, const C2 &c2) {
```
pansk: The first one could be a range for, and the second a iterator-based one.
Even better, actually: the whole thing could be reduced to a single iteration (adding the constraint for both C1 and C2 to be bidirectional iteratable).

tiny_dnn/core/framework/tensor.h
```diff
@@ -348,6 +353,22 @@ class Tensor {
...
+	if (new_offset + product(new_shape) > size_) {
+            throw nn_error("Cannot create a view from this tensor");
+	}
+
+        return Tensor(storage_ptr_, new_offset, new_shape);
```
pansk: Does this really works if I create a view from a tensor (not a view) and then take the view out of it?
Actually, I think the first view will give bad results already: offset and shape are not enough information to compute correct offsets, you'll also need the original shape.

tiny_dnn/core/framework/tensor_storage.h
edgarriba: not sure that we can reduce the operation since c1 (offset starting point) can have up to `kDimensions` and we have to iterate over c2 (current tensor shape) to compute the offset

test/test_tensor.h
Randl: @edgarriba I mean that if for some reason `host_data[i]` returns `i+2`nd element instead of `i`th, it will still pass tests. So I suppose to fill tensor with something like `0,1,2,3,4,5,6,7,8`... such that you make sure right element access.

tiny_dnn/core/framework/tensor_storage.h
Randl: @edgarriba You can fill new array with multiplication of last `n` elements of `c2`, reducing complexity, but that's not really needed given sizes of `c1` and `c2`.

pansk: The inner loop calculates the accumulated product of the last N dimensions and the external loop uses just that value, mutliplies by a value from C1 and accumulates the sum.
If you do everything backwards, you can do everything in 1 loop, without a temporary array.

tiny_dnn/core/framework/tensor.h
pansk: Let's make a simple example with a 3x3 tensor, and a 2x2 subtensor.

The storage is:
1 2 3 4 5 6 7 8 9

And the value for the offset and shape are
offset = 0, shape = {3, 3}

Now, I want to take a subtensor, in different ways:

1) first row
2) second row,
3) first column
4) top-left 2x2 submatrix

1) Offset = 0, shape = {1, 3} <- ok: {{1, 2, 3}}
2) Offset = 3, shape = {1, 3} <- ok: {{1, 2, 3}}
3) Offset = 0, shape = {3, 1} <- Broken! It will result {{1}, {2}, {3}}
4) Offset = 0, shape = {2, 2} <- Broken! It will result {{1, 2}, {3, 4}}

Am I missing something obvious?

pansk: I think the problem is that the code is assuming the subtensor is contiguous:
- if we want to limit views to contiguous subtensor, we should at least assert that the last N dimension match
- if we don't want to limit to contiguous subtensor, we should pass the size of the "base" tensor, and store it somewhere.

Randl: @pansk You are right, I assumed contiguous views.

tiny_dnn/core/framework/tensor_storage.h
edgarriba: @pansk I see more @Randl's solution. Enlight me :smile:

edgarriba:
@pansk @Randl I want to merge this and let you fix/add the features in your comments.
Still no idea why appveyeour doesn't like my tests :sob:

bhack:
@edgarriba http://stackoverflow.com/questions/13157671/seh-exception-with-code-0xc0000005-thrown-in-the-test-body

beru:
@edgarriba I see this error message on Output of Visual Studio 2015 (VS14)
```
Exception thrown at 0x00007FFEBFD37788 in tiny_dnn_test.exe: Microsoft C++ exception: tiny_dnn::nn_error at memory location 0x0000005F43FDEB80.
```

Bounds checking code like this one causes the exceptions.
```
EXPECT_THROW(tensor.host_at(0,0,0,1), nn_error);
```

But I couldn't reproduce below error that was logged in AppVeyor, so I'll install Visual Studio 2013 (VS12) later.
`unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.`

tiny_dnn/core/framework/tensor_storage.h
pansk: Even simpler, you don't even need to do it in reverse order:

`template <typename C1, typename C2> 
static inline size_t compute_offset2(const C1 &c1, const C2 &c2) { 
    //assert(end(c1)-begin(c1) == end(c2)-begin(c2));
    size_t res = 0;
	for (size_t i = 0; i < c2.size(); ++i) { 
	    res *= c2[i];
	    res += c1[i];
	}
	return res;
}`

tiny_dnn/core/framework/tensor.h
```diff
@@ -335,7 +318,95 @@ class Tensor {
...
+
+    /*
+     * Implementation method to extract a view from a tensor
+     * Raises an exception when sizes of the starting offset and new_shape
+     * are bigger than the current dimensions number. Also raises an exception
+     * when the requested view size is not feasible.
+     */
+    Tensor view_impl(std::initializer_list<size_t> const &start,
+                     std::initializer_list<size_t> const &new_shape) {
+	if (start.size() > kDimensions || new_shape.size() > kDimensions) {
```
beru: Between L395 to L407, tabs are used instead of 4 spaces.

tiny_dnn/core/framework/tensor.h
```diff
@@ -140,23 +140,6 @@ class Tensor {
...
-                    const std::array<size_t, kDimensions> &shape) {
-        offset_ = offset;
-        shape_ = shape;
-        size_ = product(shape);
-        storage_ptr_ = storage;
-    }
-
     ~Tensor() = default;
 
     Tensor(const Tensor&other) { //TODO(Randl):deep copy
```
beru: This constructor should be removed with `#ifdef 0 ~ #endif` at the moment. Visual Studio 2013 uses this copy ctor and makes uninited instance.

beru:
```
    template<typename... Args>
    U& host_at(const Args... args) {
        static_assert(!kConst, "Non-constant operation on constant Tensor");
        return *host_ptr(args...);
    }
    
    template<typename... Args>
    U host_at(const Args... args) const {
        return *host_ptr(args...);
    }
```

for these methods, maybe it would be simpler to use `host_iter` instead of `host_ptr` since the iterator is also dereferenceable. But I wonder if it is meaningful to use `vector::iterator`.

I'd like to set existing memory to TensorStorage for usability. Current implementation allocated memory with vector container in TensorStorage but in this way, we cannot use external memory storage. I wish to see more flexibility with TensorStorage so the use of STL vector container would be a burden in that case.

tiny_dnn/core/framework/tensor.h
```diff
@@ -140,52 +140,32 @@ class Tensor {
...
+        data_dirty_ = true;
         //device_data_ is intentionally left uninitialized.
     }
-/*
- *
- * TODO(Randl): Move constructors for Tensor and TensorStorage
-    Tensor &operator = (const Tensor& other) {
-    }
+    //TODO(Randl): Move constructors for Tensor and TensorStorage
+    Tensor &operator = (const Tensor& other) {}
```
beru: You should remove or comment out this constructor. VS2013 will call this and make an uninitialized instance.

test/test_tensor.h
```diff
@@ -129,6 +129,93 @@ TEST(tensor, check_bounds) {
...
+    for (size_t i = 0; i < t2.size(); ++i) {
+        t2.host_data()[i] = float_t(i+1);
+    }
+
+    // check that root tensor has been modified assuming
+    // continous memory-
+    EXPECT_EQ(t1.host_at(0,0), float_t(1.0));
+    EXPECT_EQ(t1.host_at(0,1), float_t(2.0));
+    EXPECT_EQ(t1.host_at(0,2), float_t(3.0));
+    EXPECT_EQ(t1.host_at(0,3), float_t(4.0));
```
beru: Size of Tensor instance t1 is 3x3, so index 3 shouldn't be used.
You should change it to
```
EXPECT_EQ(t1.host_at(1,0), float_t(4.0));
```
to pass tests.

test/test_tensor.h
```diff
@@ -129,6 +129,93 @@ TEST(tensor, check_bounds) {
...
+   
+    // check that's not the top-left matrix
```
beru: This comment doesn't make sense.

I suppose the idea in here is like...

Ideal
```
1 1 1
1 1 1
1 1 1

2 2
2 2
```

Reality
```
1 1 1
1 1 1
1 1 1

2 2 2
2 

```

test/test_tensor.h
edgarriba: right! my bad

edgarriba:
@beru @Randl @pansk tests passing. Give it a last glimpse before so that I can merge

Randl:
IMHO, good to merge

PR #451, opened 2016-12-10
AVX backend optimization

beru:
Hi, this PR optimizes AVX backend. About 5% speed gain with cifar10 training.

### DIFF
--- a/tiny_dnn/core/kernels/conv2d_op_avx.h
+++ b/tiny_dnn/core/kernels/conv2d_op_avx.h
@@ -168,7 +168,6 @@ void avx_conv2d_5x5_kernel(const core::conv_params& params,
                             __m256 i2 = _mm256_loadu_ps(pi2);
                             __m256 i3 = _mm256_loadu_ps(pi3);
                             __m256 i4 = _mm256_loadu_ps(pi4);
-                            __m128 sum = _mm_loadu_ps(ppa2);
                             dst0 = _mm256_mul_ps(w0a, i0);
                             dst1 = _mm256_mul_ps(w0b, i0);
                             dst2 = _mm256_mul_ps(w0c, i0);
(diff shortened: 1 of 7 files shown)


beru:
By the way, CPU utilization is much better when minibatch size is multiple of `CNN_TASK_SIZE`.

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -256,7 +254,7 @@ void avx_conv2d_5x5_kernel(const core::conv_params& params,
 
     if (out.height_ == 1 && out.width_ == 1) {
         const double* pw = &W[0];
-        for (size_t o = 0; o < out.depth_; ++o) {
+        for (serial_size_t o = 0; o < out.depth_; ++o) {
```
edgarriba: why `serial_size_t` here?

tiny_dnn/node.h
```diff
@@ -113,7 +113,8 @@ class edge {
 
     void clear_grads() {
 		for (size_t sample = 0, sample_count = grad_.size(); sample < sample_count; ++sample) {
-			std::fill(grad_[sample].begin(), grad_[sample].end(), (float_t)0);
+		    auto& g = grad_[sample];
+		    vectorize::fill(&g[0], g.size(), (float_t)0);
```
edgarriba: make safe type cast `static_cast<float_t>(0.0)`. And why not all in the same line?

tiny_dnn/core/kernels/conv2d_op_avx.h
beru: Because the type of `tiny_dnn::core::conv_params::out::depth_` is `serial_size_t`.
The type of `tiny_dnn::core::conv_params::out` is `serial_size_t`.
I've already forgotten why nyanp started using `serial_size_t` with `cereal` library though he mentioned the reason several times in the past.

tiny_dnn/node.h
beru: Instead, I used concise brace initialization mentioned in [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Casting).

tiny_dnn/core/kernels/conv2d_op_avx.h
edgarriba: oh, right. I guess `out.depth_` is one of the serialized attributes

Issue #442, opened 2016-12-05
Windows Support for Caffe


mrgloom:
Hello, Caffe now have `Windows Support`, you can build it with VS2013, so your table is outdated.

beru:
Anyone can send PRs to update it but what about moving [the table](https://github.com/tiny-dnn/tiny-dnn#comparison-with-other-libraries) to a new Wiki page?

beru:
@mrgloom I updated the `README.md`. The table is in Wiki section, so please check it.
https://github.com/tiny-dnn/tiny-dnn/wiki/Comparison-with-other-libraries
Now, Caffe's `Windows support` is `Yes`.

edgarriba:
@mrgloom @beru can we close this?

beru:
@edgarriba It's OK for me. Is there anything to add more? @mrgloom

PR #452, opened 2016-12-10
update README.md

beru:
I recently see some issues posted about `Comparison with other libraries` table in `README.md`, so I've opened new wiki [page](https://github.com/tiny-dnn/tiny-dnn/wiki/Comparison-with-other-libraries) and removed the table with this PR. #442

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,18 +49,7 @@ Check out the [documentation](http://tiny-dnn.readthedocs.io/) for more info.
 
 ## Comparison with other libraries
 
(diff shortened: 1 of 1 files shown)


nyanp:
@beru 
LGTM 👍

edgarriba:
@beru @nyanp do you think that we can do the same for features and build sections? I think that readme is overloaded and should contain minimum information.

beru:
Maybe so. Would it be alright if I moved some of contents in readme to Wiki?

PR #453, opened 2016-12-10
Add a pointer to google coding style guide

nyanp:


beru:
I think [this rule](https://google.github.io/styleguide/cppguide.html#Type_Names) in Google coding style guide collides with below preferred coding style.
> - Use ```snake_case``` for identifiers

Maybe they are not absolutely compulsory.

Randl:
Well, `snake_case` for class names is really weird. I think we should remove it, and hold with Google style. 

Naming for macros is also part of Google style, so only the part `CNN_IN_BEGINNING` (or maybe `DNN_IN_BEGINNING`/`TINY_DNN_IN_BEGINNING`?) should be kept, and located after Google style point.

4 space indentation in once again exception from Google style and should be in "exception from Google style" part.

BTW what's the reasoning behind 4 spaces instead of 2 spaces? @edgarriba @nyanp

edgarriba:
Agree with @Randl. Also we should take into account this https://github.com/tiny-dnn/tiny-dnn/issues/454#issuecomment-266232389

nyanp:
@Randl 
Some people prefer to use ```snake_case``` for class names/identifiers because they think it is easier to read. Or other people think ```snake_case``` is good to keep consistent with STL and boost.

http://stackoverflow.com/questions/1785416/c-naming-read-input-vs-readinput
http://softwareengineering.stackexchange.com/questions/27264/naming-conventions-camelcase-versus-underscore-case-what-are-your-thoughts-ab

Anyway, I think we need a strong reason to change ```snake_case```, because it will break all existent APIs. 

As for 4-spaces vs 2-spaces, I have no strong opinion and either is fine :) 
https://ukupat.github.io/tabs-or-spaces/

@edgarriba Do you prefer to use 2-spaces?

Randl:
@nyanp Agreed about `snake_case`, so it should be moved under "Exceptions".

As for 4-spaces vs. 2-spaces, 4-spaces is kinda wide if we are keeping with 80-char wide rule

bhack:
Ros adopted google style guide but with 120 limit http://wiki.ros.org/CppStyleGuide

yumetodo:
# line length limit

>https://github.com/tiny-dnn/tiny-dnn/pull/444#discussion_r91942876
>...is not best for modern displays, 2/4-space with **120 char** is the best.

I totally agree with this opinion. 80 limit is too small to write C++.

---

# Disagree with the rule using pointer not lvalue-reference as outputs

ref: https://github.com/tiny-dnn/tiny-dnn/pull/444#issuecomment-266211292

## motivation

According to Google Coding Style Guide, on function argument, using pointer not lvalue-reference as outputs is recommended.

However, I disagree with this rule. The strong reason is **to avoid null-check**, which is required to avoid null-pointer dereference.
Null-pointer dereference is undefined behavior, C++ standard says. [Undefined behavior can result in time travel](https://blogs.msdn.microsoft.com/oldnewthing/20140627-00/?p=633/).
**The cost of null-check is small but unignorable**, I think.

Skipping null-check also give us code-readability because there are no output-less case.

## using pointer as outputs

Library code

```cpp
void library_function(some_class* output){
    if(nullptr == output){
        //How should we do on this case?
    }
    //main process
}
```

User code

```cpp
void user_function(){
    some_class c;
    library_function(&c);//on this case, null-check is not required
    library_function(nullptr);//OK, because library_function accept pointer.
}
```

Because library has null-check responsibility, Not required null-check occur.

## using lvalue-reference  as outputs

Library code

```cpp
void library_function(some_class& output){
    //main process. null check code is not exist!
}
```

User code

```cpp
void user_function(){
    some_class c1;
    library_function(c1);//OK
    some_class* c2 = nullptr;
    library_function(*c2);//NG. null pointer dereference occur.
}
```

**Null pointer dereference occur at user code not library code**. this means user has null-check responsibility.
Using lvalue-reference simplify library code and booost code-readability, I think.

## this proposal make API break change?

No. Just add overload.

edgarriba:
@yumetodo @nyanp for me it's okay 2spaces +  length limit < 80. As for output values I would use references since almost all the code we have is done like this. For those arguments that could not been modified just put a const.

edgarriba:
should we merge this and take a final decision regarding codes styles?

Randl:
We definitely should.

edgarriba:
@yumetodo are you going to work on this https://github.com/tiny-dnn/tiny-dnn/pull/455#issuecomment-266426229? Because we can iterate this topic over your commits. If so, don't make big PR.

yumetodo:
Does this merge commit means this project adopt `` length limit < 80`` rule?

edgarriba:
@yumetodo no decision yet, but we cannot get stuck only for this. Just asking I you had time since from your comments I assumed that could take this task.

PR #363, opened 2016-10-25
Add clang tooling

NewProggie:
- Normalize the license boilerplate, redirecting to the already existing LICENSE file in the root of the project. This reduces clutter and is common practice (as seen in the Chromium project, Mozilla and others)
- Add clang-format configuration file based on the Google coding style guide, using the most current version 3.9
- Apply all reformatting based on the coding style guide

PR #455, opened 2016-12-12
Fix/#444

yumetodo:
fix #144 based on @edgarriba 's review.

To apply <80 rule, I use clang-format.

```yaml
---
BasedOnStyle: 'Google'
ColumnLimit: 80
IndentWidth: 4
```

Put this file to root directory, and 

```sh
$ clang-format -i ./tiny_dnn/util/math_functions.h
```

note: LICENSE comment doesn't satisfy <80 rule, so I revert the change clang-format done at that part.

### DIFF
--- a/tiny_dnn/util/math_functions.h
+++ b/tiny_dnn/util/math_functions.h
@@ -34,58 +34,67 @@ namespace tiny_dnn {
 
 // x = x / denom
 inline void vector_div(vec_t& x, float_t denom) {
-    std::transform(x.begin(), x.end(), x.begin(), [=](float_t x) { return x / denom; });
+    std::transform(x.begin(), x.end(), x.begin(),
+                   [=](float_t x) { return x / denom; });
 }
 
 namespace detail {
-    inline void moments_impl_calc_mean(serial_size_t num_examples, serial_size_t channels,
-                                       serial_size_t spatial_dim, const tensor_t& in, vec_t& mean) {
-        for (serial_size_t i = 0; i < num_examples; i++) {
-            for (serial_size_t j = 0; j < channels; j++) {
-                float_t& rmean = mean.at(j);
-                const auto it = in[i].begin() + (j * spatial_dim);
-                rmean = std::accumulate(it, it + spatial_dim, rmean);
-            }
+inline void moments_impl_calc_mean(serial_size_t num_examples,
+                                   serial_size_t channels,
+                                   serial_size_t spatial_dim,
+                                   const tensor_t& in, vec_t& mean) {
+    for (serial_size_t i = 0; i < num_examples; i++) {
+        for (serial_size_t j = 0; j < channels; j++) {
+            float_t& rmean = mean.at(j);
+            const auto it = in[i].begin() + (j * spatial_dim);
+            rmean = std::accumulate(it, it + spatial_dim, rmean);
         }
     }
-    inline void moments_impl_calc_variance(serial_size_t num_examples, serial_size_t channels,
-                                           serial_size_t spatial_dim,
-                                           const tensor_t& in, const vec_t& mean, vec_t& variance) {
-        for (serial_size_t i = 0; i < num_examples; i++) {
-            for (serial_size_t j = 0; j < channels; j++) {
-                float_t& rvar = variance.at(j);
-                const auto it = in[i].begin() + (j * spatial_dim);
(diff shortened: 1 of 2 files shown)


Randl:
@yumetodo If you could make https://github.com/tiny-dnn/tiny-dnn/pull/363 work, that would be nice

tiny_dnn/util/math_functions.h
```diff
@@ -34,59 +34,67 @@ namespace tiny_dnn {
...
 namespace detail {
-    inline void moments_impl_calc_mean(serial_size_t num_examples, serial_size_t channels,
-                                       serial_size_t spatial_dim, const tensor_t& in, vec_t& mean) {
-        for (serial_size_t i = 0; i < num_examples; i++) {
-            for (serial_size_t j = 0; j < channels; j++) {
-                float_t& rmean = mean.at(j);
-                const auto it = in[i].begin() + (j * spatial_dim);
-                rmean = std::accumulate(it, it + spatial_dim, rmean);
-            }
+inline void moments_impl_calc_mean(serial_size_t num_examples,
```
edgarriba: @yumetodo sorry, didn't mention that you should rename all of them :smile:

yumetodo:
@edgarriba Oh...I fixed.
BTW, what is ``serial_size_t``?

tiny_dnn/util/math_functions.h
```diff
@@ -34,59 +34,67 @@ namespace tiny_dnn {
...
-                const auto it = in[i].begin() + (j * spatial_dim);
-                rmean = std::accumulate(it, it + spatial_dim, rmean);
-            }
+inline void moments_impl_calc_mean(size_t num_examples,
+                                   size_t channels,
+                                   size_t spatial_dim,
+                                   const tensor_t& in, vec_t& mean) {
+    for (size_t i = 0; i < num_examples; i++) {
+        for (size_t j = 0; j < channels; j++) {
+            float_t& rmean = mean.at(j);
```
beru: http://www.cplusplus.com/reference/vector/vector/at/
`std::vector::at` does bounds checking, so using in nested loop may not be favourable.

tiny_dnn/util/math_functions.h
```diff
@@ -34,59 +34,67 @@ namespace tiny_dnn {
...
+            const auto it = in[i].begin() + (j * spatial_dim);
```
beru: I thought we can reduce dereferencing array in here.
What about moving `in[i]` outside of `channels` loop?
Please see below example code.
```
assert(mean.size() >= channels);
for (size_t i = 0; i < num_examples; i++) {
    const auto& idat = in[i];
    for (size_t j = 0; j < channels; j++) {
        float_t& rmean = mean[j];
        const auto it = idat.begin() + (j * spatial_dim);
```

yumetodo: @beru Compiler should do it [such optimization](https://en.wikipedia.org/wiki/Common_subexpression_elimination). I don't want to write such code.

tiny_dnn/util/math_functions.h
edgarriba: @yumetodo ping

yumetodo: https://github.com/tiny-dnn/tiny-dnn/pull/455/commits/257cdd0c61b99eb16372cb11ed9db34f7c57a88e

edgarriba: umm, sorry. Didn't see that! :smile:

tiny_dnn/util/math_functions.h
edgarriba: @yumetodo for me the word 'should' means that it might not happen and we want this as much optimized as possible. But don't worry we can fix this later.

yumetodo: > for me the word 'should' means that it might not happen

I believed that compiler optimize this.
However, according to my investigation, Visual Studio 2015 Update 3 cannot.

x86: https://gist.github.com/yumetodo/a2a15bb65c157d73b7cb156de80d8f7b
x64: https://gist.github.com/yumetodo/64e244aa392bf0c5751b02c16e80f4aa

(for me, x64 assembly is difficult, however, x86 assembly shows VS 2015 Update 3 cannot optimize.)

>But don't worry we can fix this later.

Thank you.

PR #456, opened 2016-12-12
Fix predict_max_value

jpauwels:
Had to change `fprop_max` to make `max predict_max_value` work, since `fprop` does not accept two parameters (any more?).

### DIFF
--- a/tiny_dnn/network.h
+++ b/tiny_dnn/network.h
@@ -727,8 +727,8 @@ class network {
     }
 
  protected:
-    float_t fprop_max(const vec_t& in, int idx = 0) {
-        const vec_t& prediction = fprop(in, idx);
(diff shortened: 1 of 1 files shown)


edgarriba:
@jpauwels very good question! Actually I'm not able to find where `fprop` is defined accepting two input arguments. Anyways, nice catch!

Besides could you setup a small test to check the usage of this functionality?

pansk:
fprop was defined having two arguments some time ago, the second one being the thread slot where to save the results (for multi-threading purposes).

beru:
@edgarriba @pansk @jpauwels fprop stopped having two arguments around 5cf862915bcc351400aec329d184e41d21ca25ee

PR #462, opened 2016-12-19
integrate Tensor class to fully_connected forward

edgarriba:


tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -32,28 +32,25 @@ namespace tiny_dnn {
 namespace kernels {
 
 inline void
-fully_connected_op_avx(const tensor_t& in_data,
-                       const vec_t&    W,
-                       const vec_t&    bias,
-                       tensor_t&       out_data,
-                       const fully_params& params,
-                       const bool      layer_parallelize) {
+fully_connected_op_avx(const Tensor<float_t, 2>& in_data,
```
Randl: `Tensor<float_t,2, true>`

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -32,24 +32,28 @@ namespace tiny_dnn {
...
+                            Tensor<float_t, 2>*       out_data,
+                            const bool             parallelize) {
+    auto in_shape  = in_data.shape();
+    auto out_shape = out_data->shape();
 
-        for (serial_size_t i = 0; i < params.out_size_; i++) {
-            out[i] = float_t(0);
-            for (serial_size_t c = 0; c < params.in_size_; c++) {
-                out[i] += W[c * params.out_size_ + i] * in[c];
+    const float_t*  W = &*weights.host_begin();
```
Randl: I'd prefer `host_ptr(0,0)`

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -32,24 +32,28 @@ namespace tiny_dnn {
...
+    const float_t*  b = &*bias.host_begin();
+
+    for_i(parallelize, in_shape[0], [&](size_t sample) {
+	const float_t* in = in_data.host_ptr(sample, 0);
+	float_t* out = out_data->host_ptr(sample, 0);
+
+        for (size_t i = 0; i < out_shape[1]; i++) {
+            for (size_t c = 0; c < in_shape[1]; c++) {
+                out[i] += W[c * out_shape[1] + i] * in[c];
```
Randl: Maybe use view or `reshape` for instead of `W[c * out_shape[1] + i]`?

tiny_dnn/core/kernels/fully_connected_op_internal.h
edgarriba: we could but it might happens that there's no bias and then an exception is raised. Probably it's easier to provide a method that return a pointer to beginning of the vector

tiny_dnn/core/kernels/fully_connected_op_avx.h
edgarriba: yes ... but too much parameters ... we have to rethink about this

tiny_dnn/core/kernels/fully_connected_op_internal.h
edgarriba: probably, I'll check.

tiny_dnn/core/kernels/fully_connected_op_avx.h
Randl: Three parameters seems ok. We can rearrange them `<kDim, kConst, type, allocator>``, so you won't have to put `float_t ` every time

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -31,25 +31,33 @@
...
-                            const Tensor<float_t, 2>& weights,
-                            const Tensor<float_t, 2>& bias,
-                            Tensor<float_t, 2>*       out_data,
-                            const bool             parallelize) {
+template<typename T1, typename T2, typename T3, typename T4>
+inline void fully_connected_op_internal(const T1&  in_data,
+                                        const T2&  weights,
+                                        const T3&  bias,
+                                        T4*        out_data,
+                                        const bool parallelize) {
```
Randl: Though shorter, I'd prefer previous style, since it's tells much more about what happening. Also previous style gives more place for optimization I believe

edgarriba: @Randl with previous style what allocator and constness? If a user define a different from the default settings we'll have a problem. Starting to think that maybe is not that good idea to have parameters as templates. Not sure if all of them, but at least kDimensions can be defined by the list constructor. /cc @beru @pansk

Randl: Actually after more thinking I suppose to leave it as it is now, leaving parameter deduction to compiler. To make necessary checks and have access to Tensor parameters, we should add member variables for each of 4 parameters.

edgarriba: nice! Definitely I think it's more elegant in this way. In this cases we can add extra documentation regarding tensor members

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -31,25 +31,37 @@
...
+    auto in_shape  = in_data.shape();
+    auto out_shape = out_data->shape();
 
-        for (serial_size_t i = 0; i < params.out_size_; i++) {
-            out[i] = float_t(0);
-            for (serial_size_t c = 0; c < params.in_size_; c++) {
-                out[i] += W[c * params.out_size_ + i] * in[c];
+    // ideally the tensor should ahave this shape. In a future
```
beru: Found a typo. `ahave`

Randl:
@edgarriba Any reason we aren't merging this?

edgarriba:
waiting for decision of non-contiguous memory, shared Tensor API, etc

bhack:
I think we could finalize some initial choiches and merge as soon as we can.

edgarriba:
I need to finish non-contiguous version for viewed tensor, otherwise we probably will have to rewrite codes due to the change of accessors
https://github.com/edgarriba/tiny-cnn/tree/feat/non-contiguous-tensor

@Randl in case you are interested, feel free to commit into my branch

bhack:
Can we close this?

Issue #315, opened 2016-09-21
a  problem that makes me crazy


yyfsss:
when i run the tiny_cnn.sln that you give in the vc,
it always crashed and notes that
错误	2	error C2275: “T”: It is illegal to have this type used as an expression	
错误	4	error C2975: “_Align”:“std::aligned_storage”	Invalid template argument, which should be a compile time constant expressio
错误	3	error C3861: “alignof”:  Cannot find identifier

nyanp:
@yyfsss 
If you use visual studio 2013, compile errors are fixed from b1fe7fc. Please retry at the latest version.

zhangqianhui:
@yyfsss  Just use "__alignof" replace "alignof" . You can try! Or Please vs 2015.

yyfsss:
@nyanp  thanks ,it works successfully in the vs 2013 version.

nyanp:
@yyfsss 
I'm happy to hear that :) Thanks for your report!

PR #464, opened 2016-12-20
Print tensor

Randl:
Add print function for tensor

### DIFF
--- a/test/test_tensor.h
+++ b/test/test_tensor.h
@@ -926,14 +926,25 @@ TEST(tensor, nd1) {
 
 }
 
-TEST(tensor, print) {
-    Tensor<float_t, 5> t({3,2,2,2,2});
+template<size_t T> std::ostream& print_tester(std::ostream &os) {
+    os << "\nPrinting Tensor of size " << T << ":\n\n";
+    std::vector<size_t> tmp(T, 2);
+    tmp.back() = 3;
 
+    Tensor<float_t, T> t(tmp);
     t.fill(float_t(1.0));
-
-    std::cout << t;
-
+    os << t;
+    print_tester<T-1>(os);
 }
+
+template<> std::ostream& print_tester<0>(std::ostream &os) {
+    return os;
+}
+
+TEST(tensor, print) {
+    print_tester<5>(std::cout);
+}
+
 //TEST(tensor, exp) {
 //    Tensor<float_t> t(2, 2, 2, 2);
 //
--- a/test/test_tensor.h
+++ b/test/test_tensor.h
@@ -927,11 +927,11 @@ TEST(tensor, nd1) {
 }
 
 TEST(tensor, print) {
-    Tensor<float_t, 4> t({3,2,2,2});
+    Tensor<float_t, 5> t({3,2,2,2,2});
 
     t.fill(float_t(1.0));
 
-    //std::cout << t;
+    std::cout << t;
 
 }
 //TEST(tensor, exp) {
--- a/test/test_tensor.h
+++ b/test/test_tensor.h
@@ -926,15 +926,16 @@ TEST(tensor, nd1) {
 
 }
 
-template<size_t T> std::ostream& print_tester(std::ostream &os) {
-    os << "\nPrinting Tensor of size " << T << ":\n\n";
-    std::vector<size_t> tmp(T, 2);
-    tmp.back() = 3;
+template<size_t N> std::ostream& print_tester(std::ostream &os) {
+    os << "\nPrinting " << N << "-dimensional Tensor" << ":\n\n";
+    std::vector<size_t> shape(N, 2);
+    shape.back() = 3;
 
-    Tensor<float_t, T> t(tmp);
-    t.fill(float_t(1.0));
+    Tensor<float_t, N> t(shape);
+    t.fill(float_t{1.0});
     os << t;
-    print_tester<T-1>(os);
+    print_tester<N-1>(os);
+    return os;
 }
 
 template<> std::ostream& print_tester<0>(std::ostream &os) {
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -121,8 +121,8 @@ class Tensor {
      */
     explicit Tensor(const std::vector<size_t> &shape) {
         offset_ = size_t(0);
-        shape_ = shape;
         size_ = product(shape);
+        std::copy(shape.begin(), shape.end(), shape_.begin());
         storage_ptr_ = std::make_shared<TensorStorageType>(shape);
     }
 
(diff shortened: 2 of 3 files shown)


beru:
I've checked the branch with Visual Studio 2013.
Please check attached file. 
[tensor_utils.h.txt](https://github.com/tiny-dnn/tiny-dnn/files/664488/tensor_utils.h.txt)

I guess Visual Studio 2013 cannot substitute well with sizeof...(Variadic_template_parameter).
So I explicitly added a number of variadic args as another template parameter for a workaround.

Randl:
@beru The implicit number of arguments seems a bit artificial, given proposed drop of VS12 support I'm not sure it's needed.

beru:
@Randl I vote for dropping support for (all versions of) VS since majority of developers and users seem to be not using the IDE anymore. A lot of developers simply don't test their code with it.

Here's another way.
```
#if !defined(_MSC_FULL_VER) || _MSC_FULL_VER >= 190024210

TEST(tensor, print) {
    Tensor<float_t, 5> t({3,2,2,2,2});

    t.fill(float_t(1.0));

    std::cout << t;

}

#endif
```

test/test_tensor.h
```diff
@@ -927,11 +927,11 @@ TEST(tensor, nd1) {
 }
 
 TEST(tensor, print) {
-    Tensor<float_t, 4> t({3,2,2,2});
+    Tensor<float_t, 5> t({3,2,2,2,2});
```
beru: We'd be able to check the printing routine further by testing many variations of tensors.
But well... I wonder if there is any specific plans or goals for the printing routine.

pansk:
> I vote for dropping support for (all versions of) VS since majority of developers and users seem to be not using the IDE anymore.

I strongly disagree with this point.
Actually, I find the current cmake configuration quite unfriendly with Visual Studio: it's already quite difficult to work on the library with VS, because the include files are not included in any project.
Making impossible to test some parts on VS will only make things worse.

test/test_tensor.h
Randl: @beru Usage is: a) Debugging for developers b) Debugging for users

beru: @Randl Thank you for the explanation. For debugging, writing a Pretty-Printer with Python code for gdb and Natvis XML file for Visual Studio would be beneficial.

beru:
> I find the current cmake configuration quite unfriendly with Visual Studio: it's already quite difficult to work on the library with VS, because the include files are not included in any project.

@pansk I've sent #473 to address this issue. Would it be possible for you to check it?

edgarriba:
@Randl could you update with @beru's proposal?

Randl:
If anyone has an idea how to test prints without huge string constants, you're welcome to suppose.

test/test_tensor.h
```diff
@@ -926,14 +926,25 @@ TEST(tensor, nd1) {
 
 }
 
-TEST(tensor, print) {
-    Tensor<float_t, 4> t({3,2,2,2});
+template<size_t T> std::ostream& print_tester(std::ostream &os) {
```
beru: I'd prefer the template parameter name in here to be `N`.

test/test_tensor.h
```diff
@@ -926,14 +926,25 @@ TEST(tensor, nd1) {
...
+    os << "\nPrinting Tensor of size " << T << ":\n\n";
+    std::vector<size_t> tmp(T, 2);
```
beru: Local variable name `tmp` makes the code ambiguous.
What about renaming it to `shape`?

test/test_tensor.h
```diff
@@ -926,14 +926,25 @@ TEST(tensor, nd1) {
...
+    tmp.back() = 3;
 
+    Tensor<float_t, T> t(tmp);
     t.fill(float_t(1.0));
```
beru: I personally don't care but @edgarriba dislikes use of C-style casts.
So we can rewrite it in this way.
```
t.fill(float_t{1.0});
```

test/test_tensor.h
```diff
@@ -926,14 +926,25 @@ TEST(tensor, nd1) {
 
 }
 
-TEST(tensor, print) {
-    Tensor<float_t, 4> t({3,2,2,2});
+template<size_t T> std::ostream& print_tester(std::ostream &os) {
+    os << "\nPrinting Tensor of size " << T << ":\n\n";
```
beru: At a glance, it's not very clear to me what the `size` means in here.

```
    os << "\nPrinting " << T << "-dimensional Tensor" << ":\n\n";
```
I think above message would explain more about the Tensor.

tiny_dnn/core/framework/tensor.h
```diff
@@ -121,8 +121,8 @@ class Tensor {
      */
     explicit Tensor(const std::vector<size_t> &shape) {
         offset_ = size_t(0);
-        shape_ = shape;
         size_ = product(shape);
+        std::copy(shape.begin(), shape.end(), shape_.begin());
         storage_ptr_ = std::make_shared<TensorStorageType>(shape);
     }
```
beru: Is there any reason not to use constructor's member initializer list in here?

tiny_dnn/core/framework/tensor_utils.h
```diff
@@ -70,33 +70,60 @@ inline std::ostream& print_last_two_dimesions (std::ostream           &os,
     const std::array<size_t, kDim>& shape = tensor.shape();
     for (size_t k = 0; k < shape[kDim-1]; ++k) {
         for (size_t l = 0; l < shape[kDim-2]; ++l) {
-            os << " " << tensor.host_at(args..., l, k) << " ";
+            os << "\t" << tensor.host_at(args..., l, k);
```
beru:
```
    for (size_t i = 0; i < shape[kDim-2]; ++i) {
        for (size_t j = 0; j < shape[kDim-1]; ++j) {
```
I think printing 2d array as Row-major order would be better.

test/test_tensor.h
```diff
@@ -926,14 +926,25 @@ TEST(tensor, nd1) {
...
+    std::vector<size_t> tmp(T, 2);
+    tmp.back() = 3;
 
+    Tensor<float_t, T> t(tmp);
     t.fill(float_t(1.0));
+    os << t;
+    print_tester<T-1>(os);
+}
```
beru:
```cpp
    return os;
```
is missing in `print_tester` function. Adding it will solve Segmentation Fault issue with clang 3.5.0

beru:
LGTM 2.0

beru:
@edgarriba Is it OK to drop support of Visual Studio 2013?

If so, @Randl may update `appveyor.yml` file to remove `Environment: generator=Visual Studio 12` errors. Otherwise, AppVeyor build report will not be green.

edgarriba:
for me it's okay

beru:
#315 is related. I wish @yyfsss has stopped using VS2013.

edgarriba:
@beru I'll push directly in master. We'll be able to speed up appveyor tests with this.

edgarriba:
@Randl thx! I'll update `feat/tensor_integration` so that we can proceed with the task

PR #465, opened 2016-12-20
Reduce frequent memory allocation

beru:
This PR improves execution speed of the library.
I'm not sure about the reasons but I've found out memory allocation consumes a lot of time.
I tested with Ubuntu 16.04, Core i5-4670, x86_64, example_mnist_train Release build.

Speed of 1 epoch.
Before : 87.1654sec
After : 49.8546sec

On Windows, example_mnist_train built with Visual Studio 2015 takes about 16 seconds for 1 epoch, so there must be something wrong with current implementation.

### DIFF
--- a/tiny_dnn/layers/layer.h
+++ b/tiny_dnn/layers/layer.h
@@ -622,7 +622,7 @@ class layer : public node {
 
     void update_weight(optimizer *o, serial_size_t batch_size) {
         float_t rcp_batch_size = float_t(1) / float_t(batch_size);
-        vec_t diff;
+        auto& diff = weights_diff_;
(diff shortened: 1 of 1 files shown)


Randl:
That's why we should profile IMHO

edgarriba:
@beru good catch!
@beru @Randl how should we automate this?

bhack:
I think that we could run kernel perf tools on travis

Randl:
@edgarriba I think this should be done locally by developers

Randl:
Probably we should add a CMake option `CNN_PROFILE` which runs profiler

bhack:
@randl Yes but it is hard to support multiplatform lowlevel profiler

edgarriba:
@beru thx!

PR #466, opened 2016-12-20
Add profiling option

Randl:
Add option to profile to CMake


Randl:
Here is graph of calls acquired using [gprof2dot](https://github.com/jrfonseca/gprof2dot) with following options `gprof2dot -w -s -n 1 -e 0.3`:

Graph of calls

beru:
Visual Studio does not support `-pg` option. So, moving it inside of
```
if(CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
   CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
might be necessary.

CMakeLists.txt
```diff
@@ -257,6 +258,10 @@ set(CMAKE_CXX_FLAGS         "${CMAKE_CXX_FLAGS} ${EXTRA_C_FLAGS}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${EXTRA_C_FLAGS_RELEASE}")
 set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} ${EXTRA_C_FLAGS_DEBUG}")
 
+if(PROFILE AND (CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
+    CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
+    set(CMAKE_CXX_FLAGS     "${CMAKE_CXX_FLAGS} -pg")
```
beru: What about moving this to around L200, so adding same if statement can be avoided?

Randl: What is preferred - same action (compilation flags) or same condition (not VS) together? @edgarriba @nyanp

edgarriba: not sure if I'm missing something but I'd say that @beru is right, `-pg` only happens if and only if no VS.

beru: I think Randl is talking about how to organize/group flags in CMakeLists.txt

If this condition `(CMAKE_COMPILER_IS_GNUCXX OR MINGW OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")` is frequently used, using a variable can simplify the file. And what about dropping MSVC support? I think that's the best way...

edgarriba: oh, I see. +1 for first one.

 Regarding MSVC, we should discuss with @nyanp and the others

Randl: @beru I think dropping MSVC is bad, but yet I think you shoud create issue to discuss it there

PR #469, opened 2016-12-23
- Add cmakelists for caffe importer

rperrot:
This is a simple CMakeLists.txt file for ease compilation of caffe_importer that automatically calls protobuf, compile and link the caffe importer example. 

The usage is really simple : 

```
git clone https://github.com/tiny-dnn/tiny-dnn.git
mkdir caffe_importer_build 
cd caffe_importer_build
cmake . ../tiny-dnn/examples/caffe_converter/ 
make
```

I only tested it on OS X for now.

### DIFF
--- a/examples/caffe_converter/CMakeLists.txt
+++ b/examples/caffe_converter/CMakeLists.txt
@@ -1,0 +1,22 @@
+cmake_minimum_required( VERSION 3.2 )
+
+project( caffe_converter )
+
+find_package( Protobuf REQUIRED )
+
+# generate CPP file for connexion between protobuf and tiny_dnn 
+PROTOBUF_GENERATE_CPP( PROTO_SRC PROTO_HDR ${CMAKE_CURRENT_SOURCE_DIR}/../../tiny_dnn/io/caffe/caffe.proto )
+
+# set the executable (main cpp + protobuf generated caffe.pb.h/caffe.pb.cc )
+add_executable( caffe_converter caffe_converter.cpp ${PROTO_SRC} ${PROTO_HDR} )
+
+# Set the include dir (binary for caffe.pb.h, protobuf and location of tiny-dnn source)
+target_include_directories( caffe_converter PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
+                                               ${PROTOBUF_INCLUDE_DIR} 
+                                               ${CMAKE_CURRENT_SOURCE_DIR}/../../ )
+
+# Link with protobuf 
+target_link_libraries( caffe_converter ${PROTOBUF_LIBRARY} )
+
+# Set C++11 mode 
+set_target_properties( caffe_converter PROPERTIES CXX_STANDARD 11 )


bhack:
Need you to remove https://github.com/tiny-dnn/tiny-dnn/blob/master/examples/CMakeLists.txt#L33,L40?

rperrot:
No, this can work together, the only aim of my CMakeLists is to build the importer sample independently. 

Moreover, it follows some simple guidelines on CMakeLists that should be encouraged ( use only local changes that have minimum impact on the other part of the library : forbid include(), add_defintions(), ...) 

I think it should be preferable to have a CMakeLists.txt per sample not a global one. If in the future tiny-dnn provides more than a hundred examples, there is no reason to build all of them (it should be possible but, when you want a specific sample is not possible for now).

bhack:
Ok but makes it sense to have two entry in two hierarchical level?

rperrot:
I don't know your position about it, it's just a suggestion. 

In my opinion, it should be possible to rewrite the CMakeLists in examples to include CMakeLists on the individual examples. (the examples/CMakeLists.txt should be a succession of add_subdirectory() )

If you want I can try to propose a solution like that ? This would have no impact on the global CMakeLists.txt file (i.e.: the BUILD_EXAMPLES may works as usual). The only advantage is to provide a way to build individual example.

edgarriba:
@rperrot thanks for your contribution! I think that having separated sample is the good way since it will be easy to show how to implement tiny-dnn in other frameworks or applications. However, @bhack is right that we cannot have duplicated code. We should decide which approach to use: the existing one or yours. I prefer your,s and in that case you need to do slightly modifications to root  CMake.

What I'm not still convinced is about the fact to choose what examples compile. It would be really nice but I think that the logic behind that will be too much for what we really need.

edgarriba:
@beru Could you check this?

beru:
@edgarriba Sure, but I don't have any experience of `caffe` so it takes a lot of time.

beru:
I've just finished reading the conversation in this thread. I think it's pretty much OK to have `CMakeLists.txt` per samples as it is said to lead us to have more than 100 samples in the future. :astonished:

beru:
@edgarriba What about putting common codes in parent CMakeLists.txt? so that we can avoid having duplicates in each example folders CMakeLists.txt files.

edgarriba:
I think that for now is okay to introduce this PR but before we need to remove https://github.com/tiny-dnn/tiny-dnn/blob/master/examples/CMakeLists.txt#L33,L40. /cc @rperrot 
In case that we want to introduce more examples we can think in how to handle it later.

beru:
@edgarriba I'm not sure how
https://github.com/tiny-dnn/tiny-dnn/blob/876e27971c8cc1bd63d9e2a6382c3c1e7958f563/examples/CMakeLists.txt#L33,L40?
is related to caffe importer.

Maybe you guys are talking about these lines between L48 - L60?
https://github.com/tiny-dnn/tiny-dnn/blob/876e27971c8cc1bd63d9e2a6382c3c1e7958f563/examples/CMakeLists.txt#L48,L60?

bhack:
I think it is ok if we minimize duicated code for each example. So we can share common code at high hierarchical level as @beru proposed in https://github.com/tiny-dnn/tiny-dnn/pull/469#issuecomment-269629207

edgarriba:
@rperrot thx for the contribution! If you have time try to modify parent CMake in order to check that this code works

Issue #440, opened 2016-12-03
Compiler issues a lot of warnings during compilation


festlv:
I tried to use tiny-dnn in an existing project but got a lot of warnings by just including tiny-dnn.h
The list of warnings is available here: https://paste.ee/p/0mPze

Warnings are possibly related to project's compiler flags (which are the strictest that I know of):
> -Werror -Wall -Wextra -Wpedantic

GCC version:

> Configured with: /build/gcc-multilib/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --enable-multilib --disable-werror --enable-checking=release
> Thread model: posix
> gcc version 6.2.1 20160830 (GCC) 

Since I had -Werror, these warnings were treated as errors and I had to loosen the flags to build tiiny-dnn.

By quickly scanning the list, I see two types of warnings:

1. Issues in stb_image.h (should be reported upstream?)
2. Unused parameters when some feature is turned off:
```
inline void func(int param1) {
#if defined(USE_OPENCL) || defined(USE_CUDA)
  //code that uses params1
#else
    nn_warn("TinyDNN was not build with OpenCL or CUDA support.");
#endif
}
```

@bhack suggested to propose one of these mechanisms: https://sites.google.com/site/opensourceconstriubtions/ettl-martin-1/articles/suppress-unused-parameter-compiler-warning

Since I have very little knowledge about typical environments/compilers where tiny-dnn is being used I will refrain from suggesting any particular way to fix the warnings (or whether or not they need to be fixed at all).

nyanp:
@festlv 
Hi,
You can use a macro for this purpose:
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/macro.h#L29

It's equivalent to the solution 3 in your link.
I'd very happy if you can send a PR to solve them :)

[...]

PR #471, opened 2016-12-23
remove some warnings

beru:
Hi, this PR removes some warnings.
I also added `/W4` option for Visual Studio projects to find warnings.

### DIFF
--- a/tiny_dnn/core/backend.h
+++ b/tiny_dnn/core/backend.h
@@ -62,9 +62,12 @@ inline backend_t default_engine() {
 #ifdef CNN_USE_AVX
 #if defined(__AVX__) || defined(__AVX2__)
     return backend_t::avx;
+#else
+#error "your compiler does not support AVX"
 #endif
-#endif // CNN_USE_AVX
+#else
     return backend_t::internal;
+#endif
 }
 
 #ifdef CNN_USE_NNPACK
--- a/tiny_dnn/nodes.h
+++ b/tiny_dnn/nodes.h
@@ -326,6 +326,7 @@ class sequential : public nodes {
 
     template <typename InputArchive>
     void load_connections(InputArchive& ia) {
+        CNN_UNREFERENCED_PARAMETER(ia);
         for (serial_size_t i = 0; i < nodes_.size() - 1; i++) {
             auto head = nodes_[i];
             auto tail = nodes_[i + 1];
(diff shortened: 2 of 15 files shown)


tiny_dnn/core/backend.h
```diff
@@ -62,9 +62,12 @@ inline backend_t default_engine() {
...
 #if defined(__AVX__) || defined(__AVX2__)
     return backend_t::avx;
+#else
+#error "your compiler does not support AVX"
 #endif
-#endif // CNN_USE_AVX
+#else
     return backend_t::internal;
+#endif
 }
```
edgarriba: no comments for this. However, since we treat AVX as a backend do you think that we should enable it by default as it right now? I vote to not do it and use internal backend by default.

tiny_dnn/nodes.h
```diff
@@ -486,15 +487,21 @@ class graph : public nodes {
...
-                serial_size_t head_index = e.prev()->next_port(e);
-
-                for (auto n : next) {
-                    serial_size_t tail_index = n->prev_port(e);
-                    gc.add_connection(node2id[e.prev()], node2id[n], head_index, tail_index);
+            graph_traverse(
+                l,
+                [=](layer& l) {
+                    CNN_UNREFERENCED_PARAMETER(l);
+                },
```
edgarriba: is this format given by your IDE? (just curious)

beru: No. I disable Visual Studio's Text Editor C/C++ Formatting options because I prefer to indent code by myself.

I prefer to use below format to reduce eye positioning but abandoned it because of 80 characters per line rule.

```
            graph_traverse(l,
                           [=](layer& l) {
                               CNN_UNREFERENCED_PARAMETER(l);
                           },
                           [&](edge& e) {
                               auto next = e.next();
                               serial_size_t head_index = e.prev()->next_port(e);

                               for (auto n : next) {
                                   serial_size_t tail_index = n->prev_port(e);
                                   gc.add_connection(node2id[e.prev()],
                                                     node2id[n],
                                                     head_index,
                                                     tail_index);
                               }
                           }
                          );
```

tiny_dnn/core/backend.h
beru: The reason for the change in `default_engine` function is Visual Studio emits `warning C4702: unreachable code` with Warning Level4 option `/W4`.

Compiler flags for AVX are checked in `CMakeLists.txt` so another check with preprocessor macro can be a bit redundant but I think it's not harmful.

Using AVX backend as a default backend sounds a bit unnatural because it isn't a reference implementation.

tiny_dnn/nodes.h
edgarriba: oh, cool. Agree with your POV

tiny_dnn/core/backend.h
bhack: But we want to run always on the same target where we compile? Probably we coul evaluate dynamic loading of engines at runtime..

beru: @bhack That's for sure. This is one of many limitations of the library.

beru:
This PR is related to #440

edgarriba:
@beru rebase codes

beru:
@edgarriba I force-pushed rebased codes. Please check it before merging.
Warning mentioned in #440 are still left. I wait a response from @nyanp about forked `stb_image`.

PR #473, opened 2016-12-24
add a hierarchy of library files as Visual Studio project items

beru:
Hi, this PR changes `examples/CMakeLists.txt`.
I hope this will address the problem @pansk mentioned at https://github.com/tiny-dnn/tiny-dnn/pull/464#issuecomment-269076479

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,7 +1,7 @@
 
-file(GLOB_RECURSE tiny_dnn_headers "${CMAKE_SOURCE_DIR}/tiny_dnn/*.h")
 
 if(MSVC)
+    file(GLOB_RECURSE tiny_dnn_headers "${CMAKE_SOURCE_DIR}/tiny_dnn/*.h")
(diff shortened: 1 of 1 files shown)


pansk:
Yep, it definitely solves the problem!

examples/CMakeLists.txt
```diff
@@ -1,40 +1,55 @@
-add_executable(benchmarks_all benchmarks/main.cpp)
+
+file(GLOB_RECURSE tiny_dnn_headers "${CMAKE_SOURCE_DIR}/tiny_dnn/*.h")
```
edgarriba: @beru this is not needed since during the cmake configuration we include the sources in the target
https://github.com/tiny-dnn/tiny-dnn/blob/master/CMakeLists.txt#L299

beru: @edgarriba It is needed to add file items into example projects that can be seen in Solution Explorer  window of Visual Studio.

PR #421, opened 2016-11-26
separate serialization and deserialization

beru:
This PR covers the problem mentioned in #420 .
mainly @nyanp and @bhack Could you please review the change? If you noticed any problems, please feel free to let me know and I'll do my best to fix them.

Issue #420, opened 2016-11-23
un-Tiny compilation time


bhack:
Compilation time it is a little bit really out of control and on different system it is exhausting compilation memory. Actually on some system I needed to manually limit gcc with [some flags](http://jkroon.blogs.uls.co.za/it/scriptingprogramming/preventing-gcc-from-trashing-the-system)

bhack:
Add -ftime-report -fmemory-report in CMake and try to analyze reports on your environment

pansk:
This is one of the drawbacks of a header-only library.

bhack:
Yes if we enable the reports flags we could see how this issue is specially on serialization

bhack:
http://stackoverflow.com/questions/12290639/quantifiable-metrics-benchmarks-on-the-usage-of-header-only-c-libraries

bhack:
A little bit of feeback with serializzation ON that actually it is required for compiling mnist example for mnist/train.cpp:

```
Memory Total          1556M

Top Execution times (seconds)

 phase lang. deferred    :  37.50 (50%) usr   4.45 (35%) sys  41.95 (43%) wall 2069816 kB (52%) ggc
 phase opt and generate  :  30.19 (40%) usr   6.80 (53%) sys  38.03 (39%) wall 1573418 kB (40%) ggc
 template instantiation  :  16.79 (22%) usr   3.86 (30%) sys  20.88 (21%) wall 1672857 kB (42%) gg 
|overload resolution    :  13.20 (18%) usr   3.06 (24%) sys  16.41 (17%) wall 1316716 kB (33%) ggc
```

bhack:
I've also tried to run iowyu http://stackoverflow.com/questions/30951492/how-to-use-the-tool-include-what-you-use-together-with-cmake-to-detect-unused-he to see if we could minimize includes

bhack:
http://baptiste-wicht.com/posts/2016/02/use-templight-and-templar-to-debug-cpp-templates.html

bhack:
Some more tips at http://stackoverflow.com/questions/28919285/identify-slow-to-compile-function

nyanp:
- Unfortunately iowyu doesn't help much for header only libraries.
- Note that tiny-dnn's serialization obeys the zero-overhead principle in C++ - When we don't call serialization API, the compilation time is as fast as previous release.

beru:
@bhack You can decrease build time of `example_mnist_train` by following method.

[...]

PR #481, opened 2016-12-28
Decouple activation function template parameter from layer classes

beru:
Hi, this PR decouples activation function template parameter from layer classes.

### DIFF
--- a/examples/caffe_converter/caffe_converter.cpp
+++ b/examples/caffe_converter/caffe_converter.cpp
@@ -48,10 +48,10 @@ image<float> compute_mean(const string& mean_file, int width, int height)
 }
 
 void preprocess(const image<float>& img,
-                 const image<float>& mean,
-                 int width,
-                 int height,
-                 vec_t* dst)
+                const image<float>& mean,
+                int width,
+                int height,
+                vec_t* dst)
 {
     image<float> resized = resize_image(img, width, height);
 
@@ -83,17 +83,17 @@ vector<string> get_label_list(const string& label_file)
 
 void load_validation_data(const string& validation_file,
                           vector<pair<string, int>>* validation) {
-  string line;
-  ifstream ifs(validation_file.c_str());
+    string line;
+    ifstream ifs(validation_file.c_str());
 
-  if (ifs.fail() || ifs.bad()) {
-      throw runtime_error("failed to open:" + validation_file);
-  }
+    if (ifs.fail() || ifs.bad()) {
+        throw runtime_error("failed to open:" + validation_file);
+    }
 
-  vector<string> lines;
-  while (getline(ifs, line)) {
-      lines.push_back(line);
-  }
+    vector<string> lines;
+    while (getline(ifs, line)) {
+        lines.push_back(line);
+    }
 }
 
 void test(const string& model_file,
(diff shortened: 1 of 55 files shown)


beru:
I've noticed some of layer classes still don't have `serialize` and `load_and_construct` methods implementation. So they cannot be serialized/deserialized

Talking about changes in this PR, declaring static variable for activation function might be a bit cumbersome. Another way I can think of is passing enum value to layer class constructor that selects activation function.

And calling virtual methods of function class inside long loop is not efficient and slow.
But I'd like to improve that later.

beru:
This PR introduces breaking changes so let me explain about them.

tiny-dnn specifies acitivation function type by layer class template parameter, like this.
```cpp
    nn << convolutional_layer<tan_h>(32, 32, 5, 1, 6,  // C1, 1@32x32-in, 6@28x28-out
            padding::valid, true, 1, 1, backend_type)
```

But in that way, the number of the combinations of layers classes and activation function classes gets big at compile-time and that resulted in long compilation and linking time with cereal library.

And making python interface is also hard with layer class template parameter because possible combination of layer classes and activation functions need to be instantiated for bindings.

Now in the PR branch, activation function is specified as a 1st parameter of layer class constructor.
```cpp
    static tan_h tan_h;
    nn << convolutional_layer(tan_h, 32, 32, 5, 1, 6,  // C1, 1@32x32-in, 6@28x28-out
            padding::valid, true, 1, 1, backend_type)
```
But using static variable all the time for activation functions may probably be unfavorable for library users.
So if that's the problem, passing enum type value to select activation function inside constructor would be an alternative way.

---

Talking about serialization, I added a method to get name of activation function to `activation::function` classes and also added `get_function` global function to return an instance of activation function specified by its name.

`CNN_REGISTER_LAYER_WITH_ACTIVATION` macro is not used anymore and `load_and_construct` and `serialize` methods in extended layer classes that have activation function are changed. Activation functions are saved/restored using its name. This change breaks backward compatibility of serialized data.

Please feel free to post comments.

edgarriba:
@beru couple of comments:
1) I'd would like to keep backward compatibility
2) Instead of enums I would remove activations completely from the interface in order to behave like this
```cpp
nn << conv << relu << maxpool << fully << relu;
```

beru:
> couple of comments:
> 1. I'd would like to keep backward compatibility
> 2. Instead of enums I would remove activations completely from the interface in order to behave like this
> ```
> nn << conv << relu << maxpool << fully << relu;
> ```

@edgarriba I'll update serialize/deserilize routines to keep backward compatibility in this branch.

And you mentioned `when there's no cereal in the library root it cannot be compiled even if you define the CNN_NO_SERIALIZATION macro`, if that's really true, is it OK to resolve the problem in this branch? Or should I create another PR?

Talking about the change to define a neural network graph, I have a lot of questions to ask...
- Should activation function classes inherit from `node` in that case?
- What if invalid combination of layers and functions are specified? Should the library throw exceptions?
- Do you want me to add that huge change in this PR??

beru:
@edgarriba Some layer classes still don't have the implementation of `serialize` and `load_and_construct` methods. Do you think it's OK to implement them now? or should I postpone it?

Randl:
> But in that way, the number of the combinations of layers classes and activation function classes gets big at compile-time and that resulted in long compilation and linking time with cereal library.

That's sounds like speculative thing to say, usual usage is only activation function for the whole network

beru:
@Randl Talking about build time issue, please refer #420, #421
If you uncomment calls to `network::load` or `network::save`, or comment out  `CNN_REGISTER_LAYER_WITH_ACTIVATIONS` in `CNN_REGISTER_LAYER_WITH_ACTIVATIONS` and build the programs. You can check the impact.

beru:
> That's sounds like speculative thing to say, usual usage is only activation function for the whole network

@Randl I still don't understand about this sentence. Can you please tell me more specificly?

Randl:
>  I still don't understand about this sentence. Can you please tell me more specificly?

@beru You suppose that since we compiler gets to specialize each layer for each activation, it increases compilation time significantly. I believe, that in real-life app, people will use a single activation function, which means a single specialization for each layer.

beru:
@Randl The problem is when deserializing neural network model from serialized file.
Current library uses `CNN_REGISTER_LAYER_WITH_ACTIVATIONS` macro to instantiate all possible combinations of layers and activation functions for deserialization.

Please compare build time of `tiny-dnn/examples/mnist/test.cpp` before and after commenting out `network::load` call.

beru:
@Randl 
For instance, `examples/mnist/test.cpp` does not define a neural network model at compile time unlike `examples/mnist/train.cpp`. Maybe @nyanp changed the code in that way with the use of `cereal` library.

While that dynamic model loading feature is really good to have, it needs `cereal` library.
Please look at `examples/CMakeLists.txt`. Only when `USE_SERIALIZER` is ON, `example_mnist_train` and `example_mnist_test` and `example_deconv_train` are used.

I think we should relax that limitation. What about moving `construct_net` function into separate file and included the file when `USE_SERIALIZER` is OFF with `examples/mnist/test.cpp`?, so that it can work even without cereal library.

beru:
I will close this PR. This approach (activation function as constructor parameter for layer classes) is abandoned. Future plan would be making activation functions as independent layer extended classes. 
Issue #484 and `feat/decouple_activations`  branch is created for further work.

PR #483, opened 2016-12-28
Improve the interface to import labels

kecsap:


tiny_dnn/nodes.h
```diff
@@ -183,14 +183,26 @@ class nodes {
...
-            vec->back()[t[i]] = target_value_max();
+            vec.emplace_back(outdim, target_value_min());
+            vec.back()[t[i]] = target_value_max();
+        }
+    }
+
+    void label2vec(const std::vector<label_t>& labels, std::vector<vec_t>& vec) const {
+        serial_size_t outdim = out_data_size();
+        size_t num = labels.size();
+
```
beru: By calling overloaded method `label2vec(const label_t* t, serial_size_t num, std::vector<vec_t>& vec)` from here, below duplicated code can be removed.

kecsap:
Oh, you are right. I forgot it is guaranteed that std::vector is continuous in memory.

beru:
@kecsap LGTM!

Issue #484, opened 2016-12-29
Error compilation without cereal


edgarriba:
I open this issue to discuss and coordinate serialization issues.

**PROBLEMS:**
1) Noticed that when third party `cereal` codes are removed from the root we cannot succeed with compilation even after defining `CNN_NO_SERIALIZATION` macro.
2) Compilation time slows down when `CNN_NO_SERIALIZATION` macro is defined.

**SOLUTIONS:**
1) Check which codes do not have guards for this macro.
2) As agreed in an internal discussion, the solution could be to have activations as independent ops

/cc @beru @Randl @bhack

PR #485, opened 2016-12-29
Fix compilation errors without cereal

beru:
Hi, this PR fixes compilation errors when `USE_SERIALIZER` is `OFF` in `CMakeLists.txt` and `cereal` folder is missing. #484 

I wish serialize codes were written in non-intrusive way...

### DIFF
--- a/tiny_dnn/layers/arithmetic_layer.h
+++ b/tiny_dnn/layers/arithmetic_layer.h
@@ -83,6 +83,7 @@ public:
             *in_grad[i] = *out_grad[0];
     }
 
+#ifndef CNN_NO_SERIALIZATION
     template <class Archive>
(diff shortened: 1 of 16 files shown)


beru:
@edgarriba Yes, I use below command to ready VS projects.
```
cmake -G "Visual Studio 14 2015 Win64" -DBUILD_EXAMPLES=ON -DBUILD_TESTS=ON -DUSE_SERIALIZER=OFF
```
And rename root `cereal` folder to `acereal` and built the projects to check compile errors.
Please examine it on your side as well with your toolset.

> BTW, do you think that we can do any trick to move cereal codes inside third_party?

It should be possible (with someone to reimplement the routines... :speak_no_evil:).
Open this page. http://uscilab.github.io/cereal/serialization_functions.html
And search the page with the keyword `external`.

PR #486, opened 2016-12-29
opencv3 version support

notecola:
Fixed crash when using OpenCV version 3.0+

### DIFF
--- a/examples/cifar10/test.cpp
+++ b/examples/cifar10/test.cpp
@@ -5,8 +5,8 @@
 #include <opencv2/highgui.hpp>*/
 #include "tiny_dnn/tiny_dnn.h"
 
-using namespace tiny_cnn;
-using namespace tiny_cnn::activation;
+using namespace tiny_dnn;
+using namespace tiny_dnn::activation;
 using namespace std;
 
 // rescale output to 0-100
--- a/examples/cifar10/test.cpp
+++ b/examples/cifar10/test.cpp
@@ -3,7 +3,7 @@
 /*#include <opencv2/imgproc.hpp>
 #include <opencv2/imgcodecs.hpp>
 #include <opencv2/highgui.hpp>*/
-#include "tiny_cnn/tiny_cnn.h"
+#include "tiny_dnn/tiny_dnn.h"
 
 using namespace tiny_cnn;
 using namespace tiny_cnn::activation;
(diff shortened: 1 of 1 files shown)


examples/cifar10/test.cpp
```diff
@@ -31,7 +31,12 @@ void convert_image(const std::string& imagefilename,
     for(int c = 0; c < resized.channels(); ++c){
         for(int y = 0; y < resized.rows; ++y){
             for(int x = 0; x < resized.cols; ++x){
-                data[c * w * h + y*w + x] = resized.data[y*resized.step + x*resized.step + c];
+#if CV_MAJOR_VERSION == 2
+					data[c * w * h + y*w + x] = resized.data[y*resized.step + x*resized.step + c];
+#elif CV_MAJOR_VERSION == 3
+					data[c * w * h + y*w + x] = resized.data[y*resized.step[0] + x*resized.step[1] + c];
```
Randl: Maybe define step variable before loop? Like 
```cpp
#if CV_MAJOR_VERSION == 2
  int step_x=resized.step, step_y=resized.step;
#elif CV_MAJOR_VERSION == 3
int step_x=resized.step[1], step_y=resized.step[0];
#endif

Randl:
Could same of similar issue influence any other code?

notecola:
I could not find any other place where cv::Mat::step is used so I guess the issue is quite isolated.

examples/cifar10/test.cpp
notecola: Hold on, I just found the opencv version 2.0 has the specs for step.
I believe this ` int step_x=resized.step[1], step_y=resized.step[0];` should work in both versions.
Need to test on cifar data.

notecola:
I found this issue when my tests were crashing in release mode on some images.
Turns out the issue has nothing to do with version of opencv3.
Just using step alone and no step[i] seems to be incorrect way.

The proper usage of cv::Mat::step is shown here:
http://docs.opencv.org/2.4/modules/core/doc/basic_structures.html

edgarriba:
After seeing this I guess that cifar10 test is not enabled by default during compilation. Do you think guys that we can enable it during CMake compiation? It requires OpenCV but we have `stb` backend too to play with

beru:
cifar10 test should be updated to work with/without external OpenCV library.
And it should be listed in `examples/CMakeLists.txt` file.

Also, cifar10 train and test programs still don't use external `cereal` library.
All example programs should be built with or without `cereal` and `OpenCV` libraries.
I opened #491 so please take a look at it.

edgarriba:
@notecola thx for your contribution! If you have time try to implement cifar10 example also with stb

pliptor:
Patch submitted #582 to address the issues above. Thanks.

PR #489, opened 2016-12-30
Fix/standalone compilation

edgarriba:
This PR fixes some errors when tiny-dnn is integrated to other frameworks.
-  Patch to compile without cereal
- Add macro to compile without Image API
- Quick fix to disable Gemmlowp suport


edgarriba:
@beru check this, your last PR was broken

tiny_dnn/network.h
```diff
@@ -601,6 +601,7 @@ class network {
     const_iterator begin() const { return net_.begin(); }
     const_iterator end() const { return net_.end(); }
 
+#ifndef CNN_NO_SERIALIZATION
```
beru: Throwing an exception or `assert(false);` might be good in case `#ifdef CNN_NO_SERIALIZATION`

tiny_dnn/nodes.h
```diff
@@ -43,21 +43,25 @@ namespace cereal {
 
 template <typename Archive>
 void save(Archive & ar, const std::vector<tiny_dnn::layerptr_t>& v) {
+#ifndef CNN_NO_SERIALIZATION
     ar(cereal::make_size_tag((cereal::size_type)v.size()));
```
beru: Throwing an exception or `assert(false);` might be good in case `#ifdef CNN_NO_SERIALIZATION`

tiny_dnn/network.h
beru: Or is it better to nullify `network::save` and `network::load` method definitions when `#ifdef CNN_NO_SERIALIZATION`? So that errors can be detected at compile time.

beru:
> check this, your last PR was broken

@edgarriba Thank you for the fix! My PR was incomplete nevertheless it passed tests.
I'm not sure about the reason why it was built without errors.. maybe not referenced codes weren't checked during the build.

I think there are a lot of hidden unreferenced errneous codes in the library such as
`max_unpooling_layer::forward_propagation` and `max_unpooling_layer::back_propagation`.
There is no `test_max_unpooling_layer.h` file in `test` folder so the class isn't probablly well tested.

Talking about your PR, I cloned and checked out `fix/standalone_compilation` branch and tested on guest Linux OS in Virtual Box on Windows. It passed the tests.

```
$ uname -a
Linux debian 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u2 (2016-10-19) x86_64 GNU/Linux

$ /usr/bin/c++ --version
Debian clang version 3.5.0-10 (tags/RELEASE_350/final) (based on LLVM 3.5.0)
Target: x86_64-pc-linux-gnu

$ /opt/cmake-3.7.1-Linux-x86_64/bin/cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=ON -DUSE_SERIALIZER=OFF .

$ ./tiny_dnn_test 
.......
[==========] 194 tests from 22 test cases ran. (438249 ms total)
[  PASSED  ] 194 tests.

```

During build, I saw a lot of warnings with `backend_avx.h` and `backend_nnp.h`. Probablly, the reason is I didn't remove unreferenced members at #479.

tiny_dnn/nodes.h
```diff
@@ -43,24 +43,32 @@ namespace cereal {
...
     for (auto n : v) {
         tiny_dnn::layer::save_layer(ar, *n);
     }
+#else
+    throw tiny_dnn::nn_error("TinyDNN was not build with Serialization support");
```
beru: The past tense of build is built so, past simple passive form sentence would be
`"TinyDNN wasn't built with Serialization support"`

tiny_dnn/nodes.h
```diff
@@ -472,7 +480,12 @@ class graph : public nodes {
 
         template <typename Archive>
         void serialize(Archive & ar) {
-            ar(CEREAL_NVP(connections), CEREAL_NVP(in_nodes), CEREAL_NVP(out_nodes));
+#ifndef CNN_NO_SERIALIZATION
+            ar(CEREAL_NVP(connections), CEREAL_NVP(in_nodes),
+               CEREAL_NVP(out_nodes));
```
beru: By the way, I prefer this style
```
            ar(CEREAL_NVP(connections),
               CEREAL_NVP(in_nodes),
               CEREAL_NVP(out_nodes));
```

beru:
`build` -> `built` . Other than that LGTM

tiny_dnn/nodes.h
edgarriba: I'm thinking more on this and I think that it will be better to provide that error message via a custom exception `nn_not_build_error'`

beru: How about this one?
```cpp
#ifndef CNN_NO_SERIALIZATION
template <typename Archive>
void save(Archive & ar, const std::vector<tiny_dnn::layerptr_t>& v) {
    ar(cereal::make_size_tag((cereal::size_type)v.size()));
    for (auto n : v) {
        tiny_dnn::layer::save_layer(ar, *n);
    }
}

template <typename Archive>
void load(Archive & ar, std::vector<std::shared_ptr<tiny_dnn::layer>>& v) {
    cereal::size_type size;
    ar(cereal::make_size_tag(size));

    for (size_t i = 0; i < size; i++) {
        v.emplace_back(tiny_dnn::layer::load_layer(ar));
    }
}
#endif // #ifndef CNN_NO_SERIALIZATION
```

edgarriba: @beru well yes also possible. But do we want disable signatures when a "module" is also disabled? (applicable for the whole library) /cc @Randl @bhack @nyanp

beru: Hmm then, how about this one?

```
template <typename Archive>
void save(Archive & ar, const std::vector<tiny_dnn::layerptr_t>& v)
#ifdef CNN_NO_SERIALIZATION
    = delete;
#else
{
    ar(cereal::make_size_tag((cereal::size_type)v.size()));
    for (auto n : v) {
        tiny_dnn::layer::save_layer(ar, *n);
    }
}
#endif

template <typename Archive>
void load(Archive & ar, std::vector<std::shared_ptr<tiny_dnn::layer>>& v)
#ifdef CNN_NO_SERIALIZATION
    = delete;
#else
{
    cereal::size_type size;
    ar(cereal::make_size_tag(size));

    for (size_t i = 0; i < size; i++) {
        v.emplace_back(tiny_dnn::layer::load_layer(ar));
    }
}
#endif
```

beru: Anyways, serialization functions can also be externally implemented with `cereal` library.
So, after moving serialization code outside the library, we don't need to use these preprocessor directives.

tiny_dnn/nodes.h
```diff
@@ -43,24 +43,32 @@ namespace cereal {
 
 template <typename Archive>
 void save(Archive & ar, const std::vector<tiny_dnn::layerptr_t>& v) {
+#ifndef CNN_NO_SERIALIZATION
     ar(cereal::make_size_tag((cereal::size_type)v.size()));
```
bhack: Can we extend classes with serialization methods?

beru: Maybe technically yes but ethically not.

tiny_dnn/nodes.h
edgarriba: ok, so I'll merge like it is. We can improve later. By moving do you mean having an external project in the organization?

beru: I think moving `cereal` folder into `third_party` would be good.
And what about adding `tiny_dnn/util/serialization` folder and moving `deserialization_helper.h`, `serialization_helper.h`, `serialization_layer_list.h` files into the folder.

`external` means we can separate existing `internal` functions that use `cereal` library into a separated file. Please take a look at the explanation on this page. http://uscilab.github.io/cereal/serialization_functions.html

edgarriba: By moving `cereal` to `third_party` I remember from @nyanp that the issue was about the include paths.
Regarding serialization, I'd move codes to `tiny_dnn/core/serialzation` or  `tiny_dnn/core/util/serialzation` instead.

PR #490, opened 2016-12-30
remove nnp and dnn backends

edgarriba:


Randl:
There are definitely some places these files are [included](https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/layers/quantized_convolutional_layer.h)

edgarriba:
@Randl right! However, my compiler I don't know why doesn't detect this things! I'll fix, thx!

edgarriba:
@Randl check again please. All tests passing

PR #493, opened 2017-01-01
Trivial fixes and user request to finish an ongoing training

kecsap:


tiny_dnn/network.h
```diff
@@ -386,6 +386,16 @@ class network {
     }
 
     /**
+     * request to finish an ongoing training
+     *
+     * It is safe to test the current network performance in @a on_batch_enumerate and
+     * @a on_epoch_enumerate callbacks during training,
```
beru: I see a comma at the end of the line instead of a period, please correct it if it's not intended.

tiny_dnn/tiny_dnn.h
```diff
@@ -136,10 +136,4 @@ using slice = tiny_dnn::slice_layer;
 
 using power = tiny_dnn::power_layer;
 
-using batch_norm = tiny_dnn::batch_normalization_layer;
-
-using slice = tiny_dnn::slice_layer;
-
-using power = tiny_dnn::power_layer;
-
```
beru: I'm not sure why they are removed. Mind if I ask for the reason?

Randl: Probably because there is exact duplicate before that

beru: Oh I see, that makes sense. I couldn't see the diff well. Thank you. @Randl

tiny_dnn/network.h
```diff
@@ -768,7 +778,8 @@ class network {
         for (auto n : net_)
             n->set_parallelize(true);
         optimizer.reset();
-        for (int iter = 0; iter < epoch; iter++) {
+        finish_training_ = false;
+        for (int iter = 0; iter < epoch && !finish_training_; iter++) {
             for (size_t i = 0; i < inputs.size(); i += batch_size) {
```
beru: Please also check newly added member variable `finish_training_` at the terminating condition of the inner loop since `finish_ongoing_training` can be called in `on_batch_enumerate`.
```cpp
             for (size_t i = 0; i < inputs.size() && !finish_training_; i += batch_size) {
```

tiny_dnn/network.h
```diff
@@ -1035,6 +1046,7 @@ class network {
 
     std::string name_;
     NetType net_;
+    bool finish_training_;
```
beru: I personally think `stop_training_` or `halt_training_` or `interrupt_training_` is more suitable.

kecsap:
Updated

beru:
LGTM

PR #496, opened 2017-01-02
drop Visual Studio 12 support in Appveyor

edgarriba:


beru:
I guess `vc/vc12` folder won't be needed if Visual Studio 2013 support will be dropped.

edgarriba:
actually, are this files doing something? Because there are still `tiny-cnn` names

beru:
They are eating bytes in computer storages.

PR #497, opened 2017-01-03
Don't include unnecessary includes when CNN_SINGLE_THREAD is defined

kecsap:


beru:
LGTM

edgarriba:

PR #498, opened 2017-01-03
Optional tr1 namespace usage for random generators

kecsap:
I am not sure if this is best way to do, but when I cross-compile tiny-dnn with gcc 5.4 for a mipsel platform, the random generators are available only in tr1 namespace despite using -std=gnu++14.


tiny_dnn/util/random.h
```diff
@@ -25,12 +25,26 @@
     SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 #pragma once
-#include <random>
+#if !defined(CNN_TR1_RANDOM)
+#include <random.h>
+#else
+#include <tr1/random>
+#endif
```
beru: This should be
```cpp
#if !defined(CNN_TR1_RANDOM)
#include <random>
#else
#include <tr1/random>
#endif
```

Please cut `.h` .

pansk:
Can we just drop the support for tr1? It's pre-C++11, and <random> should be supported on all the platforms we're interested in, right now.

kecsap:
I don't mind too much if this PR is dropped. I just try to upstream my changes to minimize my delta.

bhack:
@kecsap Can you identify GCC version that moved the code away from TR1 on MIPS?

kecsap:
This change is not enough generic.

PR #499, opened 2017-01-03
Externalize some serialization functions

beru:
Hi, this PR externalizes some serialization functions in layer classes.

Friend declarations are still needed to access private members so it makes this less appealing.
I never thought implementing external serialization functions would be this difficult... 
There are still other classes that still have serialization methods such as `nodes`, `sequential`, `graph`, `core::connection_table`, `index3d`. So this PR is not a complete work.


edgarriba:
@beru thx for your contribution! :+1:

PR #254, opened 2016-08-02
protobuf 3.0 parser tool used for TensorFlow weights and graph

Wangyida:
Enough for parsing weights and graph from TensorFlow. Especially using GraphDef Class.

PR #501, opened 2017-01-05
Feat/tensorflow parser

edgarriba:


edgarriba:
@wangyida I have some problems with *.proto file path since some of them depends on others. Any tip how we can easily fix it?

tiny_dnn/io/tensorflow/tensorflow/core/framework/proto_parser.h
```diff
@@ -0,0 +1,270 @@
+#include <iostream>
+#include <fstream>
+#include <sstream>
+#include <string>
+#include <map>
+#include "tensorflow/core/framework/graph.pb.h"
```
Randl: This one is definetely doesn't exists

bhack: How could we maintain versioning of all this Tensorflow files?

bhack:
I don't know if the list of TF protobuf

tiny_dnn/io/tensorflow/tensorflow/core/framework/proto_parser.h
edgarriba: it's supposed that's generated during the cmake configuration by the proto compiler

wangyida:
@edgarriba for compiling problems with .proto files, you can have a try for:
```
protoc -I=..  --cpp_out=.. ../tiny_dnn/io/tensorflow/tensorflow/core/framework/*.proto
```
This is similar to my previous shell script for generating headers and sources from .proto files using protoc, you can find it [Here](https://github.com/wangyida/tiny-cnn/blob/8763a0b445fe0f1f359156bbd73533bc8cc67b00/tiny_cnn/io/tensorflow/protoc.sh)
The last merged PR is [there](https://github.com/tiny-dnn/tiny-dnn/pull/254) including all tools for parsing and loading, it is missed in current master branch. Is it suitable to be merged again?

edgarriba:
@wangyida I know the existence of our script to generate the proto files, however I want to simplify things and do it from CMake as it's done with Caffe
https://github.com/tiny-dnn/tiny-dnn/blob/master/examples/caffe_converter/CMakeLists.txt#L8

wangyida:
@edgarriba OK, let me have a check on CMakeList style, I just came back from Xiamen:) The current CMake checking strategy also have bugs.

edgarriba:
@wangyida oks. Feel free to commit in this branch

wangyida:
@edgarriba I had a try on the graph brach with tensorflow model parser and find that there might be some problem generating headers and source files within cmake from subdirectory, it's declared officially there in [cmake doc](https://cmake.org/cmake/help/v3.0/module/FindProtobuf.html)
```
NOTE: The PROTOBUF_GENERATE_CPP macro & add_executable() or add_library()
calls only work properly within the same directory.
```

wangyida:
So I think maybe it's also applicable to call shell command within cmakelist

bhack:
Cannot we add a CmakeList subdirectory file? What is the behaviour that we want on make install?

edgarriba:
I think that there are two possible cases for caffe/tensorflow converters:
1) The user (or library) already has the proto files
2) The user (or library) does not have the proto files

For case 1) we can assume that they at the correct place so there's no problem.
For case 2) I think that we should provide this CMake script in order to generate and point the proto files to the included directories (as is done now with the caffe_converter). I don't think that we have to include (physically) the files inside the project, instead generating them inside the build directory during the compilation could be enough so that we keep the library free of temporal files (and possible bugs like after compiling with different PROTOC version).

Regarding installation, if we behave in this way and we generate the proto files each time during the compilation we can avoid the mentioned problems.

bhack:
Ok so what is the subdirectory problem?

wangyida:
As for the updating strategy of TensorFlow, the case should be case 2, only .proto files should be given into tiny-dnn.
The problem is that PROTOBUY_GRNERATR_CPP doesn't complie proto files at this time, so protoc is not executed there. We should add add_dependency(tensorflow_parser proto_source proto_header) to ensure that our target need the generated c++ files for linking. Here is a solution that we declare it in advance and complie it before linking it with target in cmakelist file, let me have a try on this way.

bhack:
Ok but I don't fully understand the scope. We want a futher tensorflow_converter like caffe one so we have this linking dependencies only there.. or we want also in user apps?

bhack:
So the main issue is we want to use converters to reserialize models in tiny-dnn format?

wangyida:
This converter is used for loading tf models and graphs stored in protobuf files, there is a parser in graph branch used for loading TF parametric model into memory.

There are no API for reserializing Parametric models into tiny-dnn models currently even we have all graphical relationships and parameter values

bhack:
Yes I know but I am talking of the roadmap. Do we want to use for create our standalone tool that produce serialized tiny-dnn models from TF or to work in memory with third-party developed apps?

edgarriba:
closing this since @wangyida has already another PR
https://github.com/tiny-dnn/tiny-dnn/pull/522

PR #502, opened 2017-01-05
Measuring unit of elapsed benchmark time should be second

limitfan:
Measuring unit of elapsed time of AlexNet forward benchmark should be second instead of millisecond.


beru:
LGTM

Randl:
Agreed, good to merge

edgarriba:
@limitfan thx for your contribution! :+1:

PR #505, opened 2017-01-08
Fix the OpenCL/CUDA build (libgreentea_dnn)

kecsap:


Randl:
LGTM. 

@kecsap We are still not working on GPU implementation, since we want first to integrate Tensor into tiny-dnn and only then implement GPU support. Thus, GPU functional is buggy, poor and untested as for now. Thanks for contribution!

edgarriba:

PR #506, opened 2017-01-08
Fix clang compiler warning

kecsap:
Parenthesis in the stbi__SOS macro causes an extra-parenthesis warning with clang 3.9.


beru:
#440 is related.

edgarriba:
@beru I think so. So I vote to remove all stb warnings

edgarriba:

PR #507, opened 2017-01-08
Extend the 1-dimensional input support for all pooling layer types

kecsap:


Randl:
Should we extend it to any case where filter size is bigger than input size?

edgarriba:
@Randl why not, and should be done at layer level?

edgarriba:

PR #510, opened 2017-01-12
fix examples/main.cpp and update data path

NiklasRosenstein:
So that the example can be run from the examples/ folder.


edgarriba:
LGTM

Issue #429, opened 2016-11-29
Caffe Converter computes wrong results. Simply not working.


rmbarbosa:
Hello,
I compiled successfully the protobuf, and tiny-cnn and caffe_converter using VStudio 2015 64 bit versions.
But i have a problema with the results of the network.
Apparently it loads well the network and converts it successfully but the results are completly diferent.

using tiny-cnn I run the command:
**_caffe_converter.exe deploy.prototxt bvlc_alexnet.caffemodel imagenet_mean.binaryproto synset_words.txt cat.jpg_**
and the results are:
Elapsed time(s): 2.113
n02090721 Irish wolfhound,0.934235
n03814906 necklace,0.0107539
n03976657 pole,0.00875462
n03124043 cowboy boot,0.00467396
n03063689 coffeepot,0.00278851

but when i run the caffe version using the same data, i get:

**_classification.exe deploy.prototxt bvlc_alexnet.caffemodel imagenet_mean.binaryproto synset_words.txt cat.jpg_**
---------- Prediction for cat.jpg ----------
0.3094 - "n02124075 Egyptian cat"
0.1761 - "n02123159 tiger cat"
0.1221 - "n02123045 tabby, tabby cat"
0.1132 - "n02119022 red fox, Vulpes vulpes"
0.0421 - "n02085620 Chihuahua"

I've tried to see if the problema was related with the imagetype but its defined as image_type::bgr and it should be ok.
i've also tried with some samples from the mnist exemple and changed the image_type to grayscale but the problem is the same: the results are wrong when using the tiny-cnn converter.

Can anyone help? i'm i doing something wrong? did anybody got the same results?

thanks in advance.

zjulujia:
I've got similar results as you, do you figure it out?

`./example_caffe_converter ~/works/caffe/models/bvlc_reference_caffenet/deploy.prototxt ~/works/caffe/models/bvlc_reference_caffenet/bvlc_reference_caffenet.caffemodel ~/works/caffe/data/ilsvrc12/imagenet_mean.binaryproto ~/works/caffe/data/ilsvrc12/synset_words.txt ~/works/caffe/examples/images/cat.jpg
[INFO] convert Convolution => N8tiny_dnn19convolutional_layerINS_10activation8identityEEE
[INFO] shape:55x55x96
[INFO] convert ReLU => N8tiny_dnn12linear_layerINS_10activation4reluEEE
[INFO] shape:55x55x96
[...]

PR #515, opened 2017-01-14
fix broken caffe_converter

edgarriba:
In this PR the broken Caffe converter is fixed due to the fact that some users reported in https://github.com/tiny-dnn/tiny-dnn/issues/429 the bug in 5c2ab5b99abe4c283b290d9338aea080430261d7.

### DIFF
--- a/tiny_dnn/io/caffe/layer_factory.h
+++ b/tiny_dnn/io/caffe/layer_factory.h
@@ -150,14 +150,20 @@ inline void reload_weight_from_caffe_net(const caffe::NetParameter& layer,
     for (size_t caffe_layer_idx = 0; caffe_layer_idx < src_net.size(); caffe_layer_idx++) {
         auto type = src_net[caffe_layer_idx].type();
 
-        size_t next_idx = tiny_layer_idx + 1;
+	if ( detail::layer_skipped(type) ||
+	    !detail::layer_has_weights(type)) {
+            continue;
+	}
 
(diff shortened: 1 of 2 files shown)


edgarriba:
/cc @nyanp Could you explain why did you remove https://github.com/tiny-dnn/tiny-dnn/commit/5c2ab5b99abe4c283b290d9338aea080430261d7#diff-1201f4bcc000cb58d71ee51947e42774L149. ?
I realized that this is what makes the converter give wrong results.

Randl:
LGTM.

Should we add simple real life test (one of reference caffe networks?) to make sure convertor works?

edgarriba:
Probably, but which one? Usually big models require large space amount (not sure about our quota in Github)

Randl:
The simplest one should be enough to find most critical bugs. And if caffe managed to fit it into repo, so can we I think

Randl:
OK, the model is >200MB, we might add downloading to travis script and check for file existence in test

edgarriba:
download 200Mb each time might slow down the testing process. Probably we could setup a small network e.g a simple mnist recognizer with few layers

bhack:
I don't think will be hard to have a gitlfs for testing in the internal github network. What do you think?

animetrics:
A couple comments. 

1.  Just a reminder to use git-lfs for binary files so that you don't end up with every version of the file when you clone.  Also, if you don't want to pull lfs files you can always temporarily disable the smudge filter when you clone. 

2.  Per my comment in #429 this commit does not seem to completely fix the problem.  Just thought I'd also mention it here.

PR #516, opened 2017-01-15
Speed optimization

beru:
Hi, this PR optimizes various parts of the library.

### DIFF
--- a/tiny_dnn/activations/activation_function.h
+++ b/tiny_dnn/activations/activation_function.h
@@ -35,6 +35,14 @@ class function {
 
   // dfi/dyi
   virtual float_t df(float_t y) const = 0;
+  void itedf(vec_t &cur,
+             const vec_t &prev,
+             const vec_t &out,
+             size_t cnt) const {
+    for (size_t i = 0; i < cnt; i++) {
+      cur[i] = prev[i] * df(out[i]);
+    }
+  }
 
   // dfi/dyk (k=0,1,..n)
   virtual vec_t df(const vec_t &y, size_t i) const {
@@ -53,8 +61,21 @@ class function {
 class identity : public function {
  public:
   using function::df;
-  float_t f(const vec_t &v, size_t i) const override { return v[i]; }
-  float_t df(float_t /*y*/) const override { return float_t{1}; }
+  inline float_t f(const vec_t &v, size_t i) const override { return v[i]; }
+  inline void itef(vec_t &out, const vec_t &in, size_t cnt) const {
+    CNN_UNREFERENCED_PARAMETER(cnt);
+    out = in;
+  }
+
+  inline float_t df(float_t /*y*/) const override { return float_t(1.0); }
+  void itedf(vec_t &cur,
+             const vec_t &prev,
+             const vec_t &out,
+             size_t cnt) const {
+    CNN_UNREFERENCED_PARAMETER(out);
+    CNN_UNREFERENCED_PARAMETER(cnt);
+    cur = prev;
+  }
   std::pair<float_t, float_t> scale() const override {
     return std::make_pair(float_t(0.1), float_t(0.9));
   }
@@ -125,14 +146,22 @@ class softmax : public function {
     for (auto x : v) denom += std::exp(x - alpha);
     return numer / denom;
   }
+  void itef(vec_t &out, const vec_t &in, size_t cnt) const {
+    float_t alpha = *std::max_element(in.begin(), in.end());
+    for (size_t i = 0; i < cnt; i++) {
+      float_t numer = std::exp(in[i] - alpha);
+      float_t denom{0};
+      for (auto x : in) denom += std::exp(x - alpha);
+      out[i] = numer / denom;
+    }
+  }
 
   float_t df(float_t y) const override { return y * (float_t(1) - y); }
 
   vec_t df(const vec_t &y, size_t index) const override {
     vec_t v(y.size(), 0);
-    for (size_t i = 0; i < y.size(); i++) {
+    for (size_t i = 0; i < y.size(); i++)
       v[i] = (i == index) ? df(y[index]) : -y[i] * y[index];
-    }
 
     return v;
   }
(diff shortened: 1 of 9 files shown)


tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -31,6 +31,201 @@
...
+template <typename Allocator>
+inline void
+avx_fully_connected_forward_kernel(
+    const std::vector<std::vector<double, Allocator>>& in_data,
+    const std::vector<double, Allocator>&              W,
+    const std::vector<double, Allocator>&              bias,
+    std::vector<std::vector<double, Allocator>>&       out_data,
+    const fully_params &params,
+    const bool      layer_parallelize) {
+    fully_connected_op_internal(
```
edgarriba: Is this really needed?

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -31,6 +31,201 @@
...
+avx_fully_connected_back_kernel(
+    const std::vector<std::vector<double, Allocator>>& prev_out,
+    const std::vector<double, Allocator>&              W,
+    std::vector<std::vector<double, Allocator>>&       dW,
+    std::vector<std::vector<double, Allocator>>&       db,
+    std::vector<std::vector<double, Allocator>>&       curr_delta,
+    std::vector<std::vector<double, Allocator>>&       prev_delta,
+    const fully_params &params,
+    const bool      layer_parallelize) {
+    fully_connected_op_internal(
```
edgarriba: Why not directly call AVX method?

tiny_dnn/core/kernels/fully_connected_op_avx.h
beru: I think this is needed because I only optimized single float path.

tiny_dnn/core/kernels/fully_connected_op_avx.h
beru: I only optimized single float path. What do you mean calling AVX method?

beru:
@edgarriba Can you do something to appveyor job queue? Looks like it has stalled.
https://ci.appveyor.com/project/tinydnn/tiny-dnn/history

edgarriba:
@beru I don't have owner permissions

beru:
@edgarriba Who has the permission then? @nyanp? But he's been away for a while...

I guess it's about time to drop Visual Studio and Windows support from this library.
The majority of active developers don't use nor test them so it's become a burden to continue supporting it.

edgarriba:
@beru Although it's a pain to maintain, I thint that Windows support is an attractive to the library. Do you know any alternative to Appveyour?

beru:
@edgarriba It looks like there's no alternative. Maybe we can locally run the tests and submit logs until someone (I guess it's nyanp) who has rights to access the AppVeyor queue appeared...

edgarriba:
LGTM

Any report about the improvements?

edgarriba:
@beru please rebase

beru:
@edgarriba @Randl rebased and force-pushed.
Applied clang-format manually on local Windows and I noticed
`DerivePointerAlignment: true` nullifies `PointerAlignment: Right` so I had to set it false temporarily...
and I still see diffs around long comment lines.

edgarriba:
@Randl I think we should some output where it crashed during the clang test

Randl:
As for now it only says that file doesn't respect formatting, but we definitely can make it more informative

tiny_dnn/activations/activation_function.h
```diff
@@ -161,7 +161,7 @@ class softmax : public function {
   vec_t df(const vec_t &y, size_t index) const override {
     vec_t v(y.size(), 0);
     for (size_t i = 0; i < y.size(); i++)
-      v[i] = (i == index) ? df(y[index]) : -y[i] * y[index];
+      v[i]        = (i == index) ? df(y[index]) : -y[i] * y[index];
```
beru: stupid clang-format

Randl: Put braces around for body. That's both fix the issue and is considered good code style

edgarriba: @Randl can we fix it somehow via the clang file?

Randl: I believe best fix here is to put body of the loop into curly braces. But if you think it's inappropriate, I'll think of different solution

tiny_dnn/activations/activation_function.h
```diff
@@ -165,7 +196,16 @@ class tan_h : public function {
...
   }*/
 
-  float_t df(float_t y) const override { return float_t(1) - sqr(y); }
+  inline float_t df(float_t y) const override { return float_t(1) - sqr(y); }
+  void itedf(vec_t &cur,
+             const vec_t &prev,
+             const vec_t &out,
+             size_t cnt) const {
+    for (size_t i = 0; i < cnt; i++) {
+      cur[i] = prev[i] * df(out[i]);
```
edgarriba: @Randl For the same rule it should do the same thing here Could you check it?
Besides I think that some original codes were like this http://stackoverflow.com/questions/32682426/clang-format-align-function-parameter-names

Randl: Since there is a curly bracket, it doesn't align cur[i], since it's not consequentive assignmentsanymore. As I said, putting curly brackets seems like right thing to do and has nothing to do with clang-tool. Yet aligning of assignments after `for` is probably a bug

tiny_dnn/activations/activation_function.h
beru: @edgarriba @Randl I will put curly braces from next time.

I used clang-format for different program today and saw a similar problem. I ended up using `AlignConsecutiveAssignments: false` to prevent it.

Some of the other options I changed from tiny-dnn's `.clang-format` are like these.
```
ReflowComments:  false
BreakStringLiterals: false
SortIncludes:    false
```

Though clang-format is pretty useful, I will not post an essay titled "How I learned to stop worrying and love the clang-format". :persevere:

PR #518, opened 2017-01-16
Remove dead code and add missing refs for loops

kecsap:
These things were found by Clang Static Analyzer and Clazy. After removing these codes, I see 2-5 % speedup with 1D convolution layer, depending on the network topology.

### DIFF
--- a/tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
+++ b/tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
@@ -282,13 +282,6 @@ inline void tiny_quantized_conv2d_back_kernel(const conv_params &params,
     &min_prev_delta_requantized, &max_prev_delta_requantized,
     &prev_delta_requantized);
 
-  // dequantize to flaot, this could be removed within concatenated quantized
-  // network
-  vec_t prev_delta_vec = quantized_tensor_to_float<uint8_t>(
-    prev_delta_requantized, min_prev_delta_requantized,
-    max_prev_delta_requantized);
-  prev_delta = &prev_delta_vec;
-
   // Accumulate dw
   for_i(params.in.depth_, [&](int inc) {
     for (serial_size_t outc = 0; outc < params.out.depth_; outc++) {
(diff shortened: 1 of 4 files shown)


edgarriba:
@kecsap @beru @Randl Are we sure that this is not needed anymore for quantization? Notice that @wangyida is refactoring this codes https://github.com/tiny-dnn/tiny-dnn/pull/443

tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
```diff
@@ -281,10 +281,6 @@ inline void tiny_quantized_conv2d_back_kernel(const conv_params& params,
     quantize_down_and_shrink_range<int32_t, uint8_t>(prev_delta_quantized, min_prev_delta_value, max_prev_delta_value,
     &min_prev_delta_requantized, &max_prev_delta_requantized, &prev_delta_requantized);
 
-    // dequantize to flaot, this could be removed within concatenated quantized network
-    vec_t prev_delta_vec = quantized_tensor_to_float<uint8_t>(prev_delta_requantized, min_prev_delta_requantized, max_prev_delta_requantized);
-    prev_delta = &prev_delta_vec;
```
beru: This removed operation
```
prev_delta = &prev_delta_vec;
```
seems to be strange. Shouldn't it be like
```
*prev_delta = prev_delta_vec;
```

edgarriba:
@kecsap please rebase from master

beru:
> @kecsap @beru @Randl Are we sure that this is not needed anymore for quantization? Notice that @wangyida is refactoring this codes #443

@edgarriba I'm not so sure if removing dead code in below files is OK
```
tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
tiny_dnn/core/kernels/tiny_quantized_deconv2d_kernel.h
```
Maybe not. There might be s a bug in the original quantized code.
Please check my comment. https://github.com/tiny-dnn/tiny-dnn/pull/518#discussion_r96675153 
The original code overwrites the argument pointer `vec_t *prev_delta`, but that is meaningless operation.
After checking the reference implementation (`core/kernels/conv2d_op_internal.h`), it looks to me that the content of `prev_delta` should be written.

beru:
@edgarriba @kecsap Have you read my comment? Please don't forget to respond otherwise this PR will remain...

 I think adding missing reference to `range_declaration` of `Range-based for loop` should be merged. But I supposedly think @kecsap removed dead code in files `tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h` and `tiny_dnn/core/kernels/tiny_quantized_deconv2d_kernel.h` without caring its origins.

Unfortunately, @wangyida who wrote those files does not give any response to this PR.
So it's hard to get the answer... how about to revert removing dead code and adopting only the missing refs? So we can merge this PR.

edgarriba:
@wangyida can you comment this PR?

tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h
```diff
@@ -313,69 +306,6 @@ inline void tiny_quantized_conv2d_back_kernel(const conv_params &params,
...
-      db[outc] += std::accumulate(delta, deltaa, float_t{0});
-    }
-  }
-}
-
-inline void tiny_quantized_conv2d_kernel(const conv_params &params,
-                                         const vec_t &in,
-                                         const vec_t &W,
-                                         const vec_t &bias,
-                                         const vec_t &in_r,
```
Randl: The variable doesn't exists anymore but is used in line 314 https://github.com/tiny-dnn/tiny-dnn/pull/518/files#diff-a3667d0eb9568c179f54aa188939ec3aR314

Same for `bias`, `a`, `layer_parallelize`, etc.

tiny_dnn/core/kernels/tiny_quantized_deconv2d_kernel.h
```diff
@@ -279,13 +279,6 @@ inline void tiny_quantized_deconv2d_back_kernel(const deconv_params &params,
...
-  // dequantize to flaot, this could be removed within concatenated quantized
-  // network
-  vec_t prev_delta_vec = quantized_tensor_to_float<uint8_t>(
-    prev_delta_requantized, min_prev_delta_requantized,
-    max_prev_delta_requantized);
-  prev_delta = &prev_delta_vec;
-
   // Accumulate dw
   for_i(params.in.depth_, [&](int inc) {
     for (serial_size_t outc = 0; outc < params.out.depth_; outc++) {
```
Randl: Unbalanced braces. I ahve a feeling that you just mistakenly removed random part of the code

kecsap:
There are many changes underway. I will check the sources again in the fall with Clazy and Clang Static Analyzer after GSoC.

Issue #519, opened 2017-01-16
invalid use of incomplete type std::ifstream


noonv:
Try to build example under Qt 5.2.1 (mingw48_32)
and got errors:

> tiny-dnn-master\tiny_dnn\network.h:628: error: invalid use of incomplete type 'std::ifstream {aka class std::basic_ifstream<char>}' if (ifs.fail() || ifs.bad()) ^

> tiny-dnn-master\tiny_dnn\network.h:657: error: invalid use of incomplete type 'std::ofstream {aka class std::basic_ofstream<char>}'
>          if (ofs.fail() || ofs.bad())
>                 ^

and so on.

Add string 
`#include <fstream>`
to file **network.h** and it's solved the problem.

Randl:
I suppose you to send this as PR

bhack:
@edgarriba Can you close this?

PR #521, opened 2017-01-16
Add a missing include

kecsap:
Fixes issue "invalid use of incomplete type std::ifstream"
https://github.com/tiny-dnn/tiny-dnn/issues/519

### DIFF
--- a/tiny_dnn/network.h
+++ b/tiny_dnn/network.h
@@ -7,6 +7,9 @@
 */
 #pragma once
 
+#ifndef CNN_NO_SERIALIZATION
+#include <fstream>
+#endif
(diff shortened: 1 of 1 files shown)


edgarriba:
@kecsap thx! Not sure if we have to put a macro for this since it's needed only for serialization.
Check this: https://github.com/kecsap/tiny-dnn/blob/9e93dfdfcb35c1e812d52d1eeb7cd09331560118/tiny_dnn/network.h#L626

/cc @Randl @beru

beru:
LGTM

edgarriba:
@kecsap please rebase from master

PR #522, opened 2017-01-18
modified tensorflow parser from graph branch

wangyida:
This is a modified version of TensorFlow parser from graph branch, there are some techniques for compiling .proto files in cmake style

### DIFF
--- a/tensorflow/core/framework/allocation_description.proto
+++ b/tensorflow/core/framework/allocation_description.proto
@@ -1,0 +1,27 @@
+syntax = "proto3";
+
+package tensorflow;
+option cc_enable_arenas = true;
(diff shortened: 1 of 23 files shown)


wangyida:
Hi, here I started a new PR for TF parser regarding to graph branch and current caffe converter.

tensorflow/core/framework/allocation_description.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
```
edgarriba: @wangyida this files should go to `tiny_dnn/io/tensorflow`

bhack:
Can we mark this as stalled?

bhack:
What we want to do with this? Close? Recover for the Gsoc?

PR #529, opened 2017-01-22
Initial playground: activation_layer base class with relu_layer inheriting it.

karandesai-96:
This PR sets up an initial playground for performing the decoupling of layers and activations. Passing tests on this PR give a green signal to build upwards on this piece of code.

### Scope of this PR:

1. `tiny_dnn::activation_layer` inheriting from `tiny_dnn::layer`. This class resides in **tiny_dnn/activations/activation_layer.h**.
2. `tiny_dnn::relu_layer` which inherits `tiny_dnn::activation_layer`.
3. Usage of this `relu_layer` in atleast one of the existing `gradient_check` tests of network, to see whether it fits in properly with other layers.

### About the design of activations as layers:

I have made a `class activation_layer : public layer`, which will be abstract base class for other activation classes - it has three virtual methods to override:
1. `layer_type` - that's the usual layer name.
2. `forward_activation` - this one takes in `vec_t` of input data (x) and `vec_t` of output data (y) as pass by references and fills in output data (y) by applying element wise activation function.
3. `backward_activation` - Same as `forward_activation`, but this one takes four `vec_t`s - input and output data, and their gradients. All will be of same length. This method uses gradients of output data (dy) and output data (y) to fill in gradients of input data (dx).

#### Note:
* `forward_activation` is similar to `tiny_dnn::activation::function::itef` method.
* `backward_activation` is similar to `tiny_dnn::activation::function::itedf` method.

After decoupling is complete, current classes in **tiny_dnn/activations/activation_function.h** can be removed safely.

Child classes (such as sigmoid, relu layers...) must override these three functions. Constructors, `in_shape`, `out_shape`, `forward_propagation` and `backward_propagation` are common for all kinds of "activation" layers only.

### After this PR:

1. Drop the activation templating from `fully_connected_layer` and make the tests pass with decoupled layers. (Might get tricky in terms of serialization code and backward compatibility.)
2. Later complete making all the current activations as layers.
3. Gradually start dropping templating from other existing layers. (Should be doable properly if (1) gave no pain!)

### DIFF
--- a/tiny_dnn/activations/activation_layer.h
+++ b/tiny_dnn/activations/activation_layer.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "tiny_dnn/layer/layer.h"
+#include "tiny_dnn/layers/layer.h"
 #include "tiny_dnn/util/util.h"
 
 namespace tiny_dnn {
@@ -36,7 +36,7 @@ public:
 
     void back_propagation(const std::vector<tensor_t *> &in_data,
                           const std::vector<tensor_t *> &out_data,
-                          const std::vector<tensor_t *> &out_grad,
+                          std::vector<tensor_t *> &out_grad,
                           std::vector<tensor_t *> &in_grad) override {
         tensor_t &dx       = *in_grad[0];
         const tensor_t &dy = *out_grad[0];
@@ -45,7 +45,7 @@ public:
 
         for (serial_size_t i = 0; i < x.size(); i++) {
             for (serial_size_t j = 0; j < x[i].size(); j++) {
-                backward_activation(x, y, dx, dy);
+                backward_activation(x[i][j], y[i][j], dx[i][j], dy[i][j]);
             }
         }
     }
@@ -56,8 +56,8 @@ public:
                                     float_t &y) = 0;
 
     virtual void backward_activation(const float_t x,
-                                     float_t &dx,
                                      const float_t &y,
+                                     float_t &dx,
                                      const float_t &dy) = 0;
 
 private:
--- a/tiny_dnn/activations/activation_layer.h
+++ b/tiny_dnn/activations/activation_layer.h
@@ -20,14 +20,29 @@ class activation_layer : public layer {
   activation_layer(const shape3d &in_shape)
     : layer({vector_type::data}, {vector_type::data}), in_shape_(in_shape) {}
 
+  /**
+   * This constructor is suitable for adding an activation layer after
+   * 3D layers such as convolution / pooling layers.
+   *
+   * @param in_width    [in] number of input elements along width
+   * @param in_height   [in] number of input elements along height
+   * @param in_channels [in] number of channels (input elements along depth)
+   */
   activation_layer(serial_size_t in_width,
                    serial_size_t in_height,
                    serial_size_t in_channels)
     : layer({vector_type::data}, {vector_type::data}),
       in_shape_(in_width, in_height, in_channels) {}
 
-  activation_layer(serial_size_t dim)
-    : layer({vector_type::data}, {vector_type::data}), in_shape_(dim, 1, 1) {}
+  /**
+   * This constructor is suitable for adding an activation layer after
+   * 1D layers such as fully connected layers.
+   *
+   * @param in_dim      [in] number of elements of the input
+   */
+  activation_layer(serial_size_t in_dim)
+    : layer({vector_type::data}, {vector_type::data}),
+      in_shape_(in_dim, 1, 1) {}
 
   activation_layer(const layer &prev_layer)
     : layer({vector_type::data}, {vector_type::data}),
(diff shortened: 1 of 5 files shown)


beru:
@karandesai-96 You have to modify a lot of things other than adding the base activation layer. If you want to make progress step by step, that's fine.

karandesai-96:
@beru Yes I understand, I am more of a Python guy, so I'm moving slowly and steadily.

test/test.cpp
```diff
@@ -17,14 +17,15 @@ using namespace tiny_dnn::activation;
...
 // TODO(yida): fix broken test
-//#include "test_average_unpooling_layer.h"
+#include "test_average_unpooling_layer.h"
 #include "test_batch_norm_layer.h"
 #include "test_concat_layer.h"
 #include "test_convolutional_layer.h"
 #include "test_core.h"
 #include "test_deconvolutional_layer.h"
 #include "test_dropout_layer.h"
 #include "test_fully_connected_layer.h"
```
Randl: Probably should remove this. Conflicting test names and no point in testing old code.

karandesai-96: I don't remember when I uncommented this stuff, I'm rewriting my recent commit.

tiny_dnn/activations/relu_layer.h
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  void backward_activation(const vec_t x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) {
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      float_t relu_grad = y[j] > float_t(0) ? float_t(1) : float_t(0);
+      dx[j]             = dy[j] * relu_grad;
```
Randl: Maybe `dx[j]  =  y[j] > float_t(0) ? dy[j] : float_t(0);`

Randl:
LGTM. @beru @bhack please review.

tiny_dnn/activations/relu_layer.h
karandesai-96: Okay I'll do that directly.

edgarriba: For readability (and not sure about performance), I'd set `relu_grad` as `const` or declare and initial ize outside the for loop block

tiny_dnn/activations/relu_layer.h
```diff
@@ -0,0 +1,36 @@
...
+
+class relu_layer : public activation_layer {
+ public:
+  using activation_layer::activation_layer;
+
+  std::string layer_type() const override { return "relu-activation"; }
+
+  void forward_activation(const vec_t &x, vec_t &y) {
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] = std::max(float_t(0), x[j]);
```
edgarriba: Are we still inheriting `serial_size_t` ?

edgarriba:
@karandesai-96 I've added a couple of comments but looks good overall

LGTD

tiny_dnn/activations/relu_layer.h
Randl: This is going to be serealized, isn't it? For example, here https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/layers/convolutional_layer.h#L268 `serial_size_t` is used

tiny_dnn/activations/relu_layer.h
Randl: I believe the direct assignment isn't less readable (maybe requires comment that gradient of ReLU is either 1 or 0) and theoretically gives more space for optimizations.

karandesai-96: I agree with @edgarriba in terms of readability, but am unsure about performance.

karandesai-96: This change is done :white_check_mark:

tiny_dnn/activations/relu_layer.h
```diff
@@ -27,9 +27,10 @@ class relu_layer : public activation_layer {
                            const vec_t &y,
                            vec_t &dx,
                            const vec_t &dy) {
+    float_t relu_grad;
```
edgarriba: @karandesai-96 I guess then that you don't need anymore this variable

karandesai-96: Oh silly fault. Amended my previous commit. :white_check_mark:

bhack:
I think it is mergiable. Any other check?

tiny_dnn/activations/activation_layer.h
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * Populate vec_t of elements 'dx' according to gradient of activation.
+   *
+   * @param x  input vector of current layer (same as forward_activation)
+   * @param y  output vector of current layer (same as forward_activation)
+   * @param dx gradient of input vectors (i-th element correspond with x[i])
+   * @param dy gradient of output vectors (i-th element correspond with y[i])
+   **/
+  virtual void backward_activation(const vec_t x,
```
beru: @karandesai-96 You should make 1st argument `x` to a reference `&` type. `const vec_t &x`

tiny_dnn/activations/relu_layer.h
beru: @edgarriba @Randl @nyanp 
I don't think implementation in layer classes should use `serial_size_t` type inside its code.
Serialization classes should take care of data sizes when they work.

tiny_dnn/activations/relu_layer.h
```diff
@@ -0,0 +1,36 @@
...
+
+  std::string layer_type() const override { return "relu-activation"; }
+
+  void forward_activation(const vec_t &x, vec_t &y) {
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] = std::max(float_t(0), x[j]);
+    }
+  }
+
+  void backward_activation(const vec_t x,
```
beru: @karandesai-96 You should make 1st argument `x` to a reference `&` type. `const vec_t &x`

PR #536, opened 2017-01-28
use ThreadPool

beru:
Hi, this PR changes `parallel_for` function to use ThreadPool library written by Jakob Progsch, Václav Zeman. https://github.com/progschj/ThreadPool

This will improve execution speed on Linux.


Randl:
What are speedups? Would it give speedup on windows too?

beru:
I'm not sure about the speed difference on Windows.
I noticed Debugging with Visual Studio 2015 is broken after using Windows Insider Preview.
Then I installed Ubuntu 16.04 on my notebook, so I can't use Windows right now.
I will check with another computer that has Windows installed.

beru:
Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz

program | master(33052fe26030a841fa729a8fa60744d0f46abee2) | thread_pool(5cf904e0bff6db0da7723a74d060d2d8868c50aa)
---|---------------|---------------
example_mnist_train | 19.5 (sec) | 14.5
example_cifar_train | 87.1 | 80.5

[log.txt](https://github.com/tiny-dnn/tiny-dnn/files/737012/log.txt)

Randl:
Is the choice of thread pool implementation critical? For example, there is also https://github.com/vit-vit/CTPL/ which seems to be pretty popular

beru:
I don't think the choice is critical. I just prefer simpler one.
https://github.com/progschj/ThreadPool/blob/master/ThreadPool.h
https://github.com/vit-vit/CTPL/blob/master/ctpl_stl.h

Randl:
Agreed, of course if simplicity doesn't affects performance, which was my concern.

edgarriba:
@beru should we move the lib to third_party?

beru:
If we seriously pursuit performance, code will be verry ugly. I can't stand such crudeness.

edgarriba:
I mean only this file https://github.com/beru/tiny-cnn/blob/thread_pool/tiny_dnn/util/ThreadPool.h

beru:
Is moving the foreign header file to `third_party` folder necessary?
If so, adding an user option `USE_THREAD_POOL` and a definition `-DCNN_USE_THREAD_POOL` to `CMakeLists.txt` and using preprocessor statements might be needed. Otherwise, users will not be able to use tiny-dnn without `third_party` folder.

edgarriba:
you're right, macros are needed. The policy is to be as modular as possible in terms of third party lilbs. What's true is that in this case they just play around with std.

beru:
Intel(R) Core(TM) i5-4670 CPU @ 3.40GHz
Microsoft Windows 10 (10.0) Professional 64-bit   (Build 14393) 

average time per epoch

program | master(33052fe26030a841fa729a8fa60744d0f46abee2) | thread_pool(5cf904e0bff6db0da7723a74d060d2d8868c50aa)
---|---------------|---------------
example_mnist_train | 7.45 (sec) | 7.2
example_cifar_train | 70.7 | 70.6

Not so effective on Windows.

beru:
Intel(R) Core(TM) i5-4670 CPU @ 3.40GHz
Microsoft Windows 10 (10.0) Professional 64-bit   (Build 14393) 

average time per epoch
c573e3e3f7a23140ce00442ef15c35a5f03653fc

program | -DUSE_THREADPOOL=ON | -DUSE_THREADPOOL=OFF
---|---------------|---------------
example_mnist_train | 7.5 (sec) | 6.83
example_cifar_train | 71 | 70.8

This time a bit effective with example_mnist_train on Windows.
Maybe I didn't measure well the last time...

edgarriba:
LGMT

nyanp:
@beru
Nice, LGTM 👍🏻

beru:
I think I need to revise this PR.
I'd like to implement custom thread pool class inside `parallel_for.h` file.
Because there are some restrictions with using third_party implementation.
And there's no contineuous integration test that checks this feature since `USE_THREAD_POOL` is OFF by default.

tiny_dnn/util/parallel_for.h
```diff
@@ -96,61 +91,113 @@ void parallel_for(int begin, int end, const Func &f, int /*grainsize*/) {
...
-  int total_size = end - start;
-  if (total_size <= nthreads) {
-    for (int i = 0; i < total_size; i++) {
-      int block_start = start + i;
-      int block_end   = start + i + 1;
-      futures.emplace_back(pool.enqueue([block_start, block_end, &f] {
-        f(blocked_range(block_start, block_end));
-      }));
+struct thread_pool {
+  thread_pool(size_t nthreads = std::thread::hardware_concurrency())
```
edgarriba: @beru just to keep things organized, could you move it in a separated file?

beru: Then how about moving `struct thread_pool` into new `tiny_dnn/util/thread_pool.h` file?
And how about `singleton` template class? Do I need to move it into separated file?

beru: Just to note that this `thread_pool` class isn't for generic purpose.

edgarriba: mmm, well and then what about just having a whole dir for multiprocess stuff?

beru: What about moving `thread_pool` and `singleton` into `tiny_dnn::detail` namespace?

beru: Having new dir for multiprocess stuff is alright. But this `thread_pool` is solely made for `parallel_for` function. So, separating it to a new file might be an overkill.

beru:
osx jobs somehow canceled...

beru:
decrease in performance with self-made thread_pool :sob:

Randl:
So
* implementation of threadpool is important, we might want to test couple of them
* I believe we should make an exception for threadpool, especially if it'll be a singleheader, and include it in main sources and bit as third party lib.

beru:
@Randl @edgarriba @nyanp @pansk 
> So
> - implementation of threadpool is important, we might want to test couple of them
> - I believe we should make an exception for threadpool, especially if it'll be a singleheader, and include it in main sources and bit as third party lib.

Testing other (threadpool) libraries is fine. I wish to see more performant one with concise code.
And I agree that threadpool implementation is one of the most important mechanisms of the library when it comes to speed. So I ask you to review and find bugs and improve the code in this PR.

To further enhance thread pool performance, we need to take into consideration Simultaneous Multithreading. I'd like to take care of it in future.
http://eli.thegreenplace.net/2016/c11-threads-affinity-and-hyperthreading/

Talking about third party libs, I think the policy of placing those files under `tiny_dnn/../third_party` folder is a bit harsh. But since this library should be tiny. Maybe it's one of burdens developers have to endure...

kecsap:
Ok, but can we merge this very important PR now, eventually? :)

beru:
@kecsap It usually takes a decade of scrutinization to review a 100 lines of code.. if possible.

bhack:
@beru Decade? Are you are aware of a really [decade long review process](http://ghtorrent.org/pullreq-perf/BVLC-caffe/) :)

beru:
@kecsap @bhack I'd expected technical code review to this PR but it looks like abundant free time is not available for anyone.... so you guys didn't read my stupid comment, all right?

Randl:
@beru imho instead of rewriting something implemented 

https://github.com/bloomen/cxxpool
https://github.com/inkooboo/thread-pool-cpp
https://github.com/nbsdx/ThreadPool

Here are some more threadpool libraries. All are header only. I think we shouldn't care about what's inside, as far as it's header only and works on all platforms. So we should just test couple of those and choose the fastest one.

Even though we introduce hard dependency, it's justified IMHO. @edgarriba @nyanp @bhack what do you think?

beru:
@Randl Among those you have introduced, I personally think the middle one (https://github.com/inkooboo/thread-pool-cpp) is not suitable because of several reasons.
- It's `C++14` header pool, not `C++11`.
- It is comprised of several header files. Not a single header file library. (but being so may not be a problem, it's just a matter of choice I supoose..)
- its README.md says `All code except MPMCBoundedQueue is under MIT license.` so using it will complicate the license of this library.

beru:
@Randl 
> Here are some more threadpool libraries. All are header only. I think we shouldn't care about what's inside, as far as it's header only and works on all platforms. So we should just test couple of those and choose the fastest one.

I can understand choosing a black box approach as it reduces the cost of learning. But well... anyways.

Another concern is if putting third-party library files into `tiny_dnn` folder is really not allowed, adding a cmake option and preprocessor macros are needed and we need to increase the amount of Continuous Integration tests. That's something I don't want to choose..

bhack:
What is the actual status of  std::async implementations?

beru:
@bhack I don't understand your question. Can you make it more precisely? Which implementation? GNU's libstdc++? llvm libc++? Your statement is so sparse...

`std::async` is used in original implementation.

This page (http://en.cppreference.com/w/cpp/thread/async) says
> The template function async runs the function f asynchronously (potentially in a separate thread **which may be part of a thread pool**)

But it seems still not.

bhack:
Yes I mean if implementations (in all compiler we support) under the line use efficient threading allocation.

beru:
@bhack So do you suggest to test those or wait for future std libs?

bhack:
Probably we could try to benchmark it but I was also reading [this analysis](http://eli.thegreenplace.net/2016/the-promises-and-challenges-of-stdasync-task-based-parallelism-in-c11/)

Randl:
> Another concern is if putting third-party library files into tiny_dnn folder is really not allowed, adding a cmake option and preprocessor macros are needed and we need to increase the amount of Continuous Integration tests. That's something I don't want to choose..

That's exactly what I meant - making it hard dependency, i.e. not have an option to compile tiny-dnn without it. Anyways we are supposed to be hardly dependent on some thread pool implementation, so does it matter if it's written by us or someone else, as far as it's header only, supports all the platforms and is provided by our repository?

beru:
@Randl So are you saying including third party library/ies inside `tiny_dnn` folder should be permitted?
If it's not allowed, I think third_party libs should be optional.

Randl:
@beru I say that we should consider allowing hard dependency (on header-only library supporting all the platforms) in some very limited and specific cases like this one or automatic differentiation.  Where the library will be located in source tree doesn't really bothers me. 

I'm looking from the following point of view: there will be header file for thread pooling, so does it really matters who wrote it? But chances are there is a library written by people who are specializing on this stuff, so it'll be faster and less buggy.

bhack:
I think that we could include something, better if BSD cause it will not be included [forever](https://www.khronos.org/assets/uploads/developers/library/2016-supercomputing/SC16-compareSYCL-Michael-Wong_Nov16.pdf).

beru:
@Randl I wished I simply could.

The advantage of using in-house made library is we won't have to worry about additional license.

I'd like to put thread pool and singleton implementation inside existing `parallel_for.h`. So additional header file won't always be necessary in this way. I prefer to choose this way. I don't want to broaden the scope of this PR.

By the way, my understanding is a singleton should be used with/without using third party thread pool libraries. Otherwise it would probablly difficult to prevent multiple instatntiation of thread pool inside `parallel_for` template function. So, Is it OK for you to use singleton impl I wrote for thread pool? Or should we also choose from supposedly high quality third party libraries somewhere around github?

bhack:
It is ok for me after we evaluate [1.3](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ri-singleton)

Randl:
> I think that we could include something, better if BSD cause it will not be included forever.

C++20 is some 5-6 years from us if we keep current compiler support scheme.

bhack:
Yes also 5-6 year is not forever if this library will be still alive. So I don't think that we cannot maintain a simply Thread Pool in the meantime.

beru:
@Randl 

I tested 6 thread pool libraries mentioned in here with `example_mnist_train` program.
However, I couldn't make 2 of them work property. The cause is still unknown.

Tested env:
Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
Ubuntu 16.04.1 LTS
gcc 5.4.0

Number | Library Name | mnist 1 epoch time | comment
---|---------------|---------------|-------------
1 | beru_thread_pool_4_parallel_for | 12 to 14 sec | the fastest, I self-recommend this
2 | bloomen_cxxpool | 14 to 15 sec | 
3 | inkooboo_thread-pool-cpp | too slow so stopped measuring | my usage isn't correct?
4 | nbsdx_ThreadPool | halted | didn't work at all. my usage was wrong?
5 | progschj_ThreadPool | 15 to 16 sec | 
6 | vit-vit_CTPL | 16.5 to 16.7 sec | 

With 1839f05, thread pool library is selectable with command line parameter.
`$ cmake -DBUILD_EXAMPLES=ON -DTHREAD_POOL_KIND=1`

beru:
Time to finish this PR.

bhack:
?

Randl:
Not sure why you closed it, but if actually in-house realization is the fastest, then just revert to https://github.com/tiny-dnn/tiny-dnn/pull/536/commits/50cec2438b283bf2361bd3d669d1edf5d66f781d and lets merge it.

beru:
@Randl According to the report from @kecsap, Intel TBB is faster than puny thread-pool, so I'll test it privately and I don't think merging will happen soon, as they aren't active at all.

beru:
@Randl @kecsap 

Just installed Intel TBB and measured performance.

Tested env:
  Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
  Ubuntu 16.04.1 LTS
  gcc 5.4.0

  Installed TBB with package manager
  $ sudo apt install libtbb-dev

Tested revision:
  5c93b2d0b2873bb97db4278abfc43f408900f137

average time per epoch
program | USE_TBB | in-house
---|---------------|---------------
example_mnist_train | 10.6 (sec) | 12.2
example_cifar_train | 82.3 | 78.2

beru:
@Randl Do you know how to make these third-party thread pool libraries work with tiny-dnn?
https://github.com/inkooboo/thread-pool-cpp
https://github.com/nbsdx/ThreadPool

I tried to use them but failed.

Randl:
@beru if you're not in hurry, I ll try to make them work next week.

beru:
@Randl I'm not in a hurry. I will check why TBB isn't very performant with example_cifar_train.

kecsap:
My test results for Cifar-10 (SSE+AVX):

Cifar-10

TBB                               199
beru_thread_pool_4_parallel_for   246
5c93b2                            247
bloomen_cxxpool	                  254
progschj_ThreadPool	          250
vit-vit_CTPL	                  254

It seems that it really depends on the platform when TBB or others are the best.

edgarriba:
@beru is this ready to be merged? in that case do we really need all those files in third_party?

beru:
@edgarriba Not yet. We're still testing several thread pool libraries.

Randl:
@beru I've run some test and they aren't promising at all. As for me, `beru_thread_pool_4_parallel_for` should be merged.

beru:
@Randl Could you share more details? Did you test these thread pool libraries?
https://github.com/inkooboo/thread-pool-cpp
https://github.com/nbsdx/ThreadPool

PR #554, opened 2017-02-09
Use pass-by-reference instead of pass-by-value in backward_activation.

karandesai-96:
Quickfix PR following #529. I passed a `vec_t` by value instead of reference in `activation_layer::backward_activation`. That hurts performance, hence rectified it.


beru:
LGTM

Randl:
LGTM too

edgarriba:
LGTM

bhack:
:+1:

PR #556, opened 2017-02-09
Add an optional implementation of parallel_for using Grand Central Dispatch

rolandpersson:
This change adds a configurable option CNN_USE_GCD which if defined enables a version of parallel_for implemented by using Grand Central Dispatch. This runs a lot faster than the default implementation of parallel_for that uses std::async. 

I spend a lot of time training and experimenting with tiny-dnn on a Mac and being able to do the training more than twice as fast is quite valuable.

Some numbers from my machine (MacBook Pro 15-inch, 2016 2,9 GHz Intel Core i7)

examples/example_mnist_train data
GCD 		5 minutes 19 seconds
std::async      13 minutes 49 seconds

test/tiny_dnn_test
GCD		14.5 seconds
std::async	2 minutes 47 seconds

### DIFF
--- a/tiny_dnn/config.h
+++ b/tiny_dnn/config.h
@@ -31,6 +31,11 @@
 // #define CNN_USE_OMP
 
 /**
+ * define to enable Grand Central Dispatch parallelization
+ */
+//#define CNN_USE_GCD
+
+/**
  * define to use exceptions
  */
 #define CNN_USE_EXCEPTIONS
--- a/tiny_dnn/util/parallel_for.h
+++ b/tiny_dnn/util/parallel_for.h
@@ -31,6 +31,10 @@
 #include <thread>
 #endif
 
+#if defined(CNN_USE_GCD) && !defined(CNN_SINGLE_THREAD)
+#include <dispatch/dispatch.h>
+#endif
+
 namespace tiny_dnn {
 
 #ifdef CNN_USE_TBB
@@ -80,6 +84,30 @@ template <typename Func>
 void parallel_for(int begin, int end, const Func &f, int /*grainsize*/) {
 #pragma omp parallel for
   for (int i = begin; i < end; ++i) f(blocked_range(i, i + 1));
+}
+
+#elif defined(CNN_USE_GCD)
+
+template<typename Func>
+void parallel_for(int begin, int end, const Func& f, int grainsize) {
+    int count = end - begin;
+    int blockSize = grainsize;
+    if (count < blockSize || blockSize == 0) {
+        blockSize = 1;
+    }
+    int blockCount = (count + blockSize - 1) / blockSize;
+    assert(blockCount > 0);
+    
+    dispatch_apply(blockCount, dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^(size_t block) {
+        int blockStart = static_cast<int>(block*blockSize);
+        int blockEnd = blockStart + blockSize;
+        if (blockEnd > end) {
+            blockEnd = end;
+        }
+        assert(blockStart < blockEnd);
+        
+        f(blocked_range(blockStart,blockEnd));
+    });
 }
 
 #elif defined(CNN_SINGLE_THREAD)
--- a/tiny_dnn/util/parallel_for.h
+++ b/tiny_dnn/util/parallel_for.h
@@ -88,26 +88,27 @@ void parallel_for(int begin, int end, const Func &f, int /*grainsize*/) {
 
 #elif defined(CNN_USE_GCD)
 
-template<typename Func>
-void parallel_for(int begin, int end, const Func& f, int grainsize) {
-    int count = end - begin;
-    int blockSize = grainsize;
-    if (count < blockSize || blockSize == 0) {
-        blockSize = 1;
-    }
-    int blockCount = (count + blockSize - 1) / blockSize;
-    assert(blockCount > 0);
-    
-    dispatch_apply(blockCount, dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^(size_t block) {
-        int blockStart = static_cast<int>(block*blockSize);
-        int blockEnd = blockStart + blockSize;
-        if (blockEnd > end) {
-            blockEnd = end;
-        }
-        assert(blockStart < blockEnd);
-        
-        f(blocked_range(blockStart,blockEnd));
-    });
+template <typename Func>
+void parallel_for(int begin, int end, const Func &f, int grainsize) {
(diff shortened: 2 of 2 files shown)


Randl:
Please run clang-format on code

edgarriba:
@rolandpersson very impressive speed performance! Have you tried in Ubuntu?
Besides, would you mind to add it as a CMake option? Maybe in another PR

edgarriba:
overall LGTM

bhack:
@edgarriba Why ubuntu?

edgarriba:
@bhack with Ubuntu I mean Linux in general
https://github.com/nickhutchinson/libdispatch

bhack:
@edgarriba Yes there is also multipaltform [xdispatch](http://opensource.mlba-team.de/xdispatch/docs/current/index.html)  and we talked about this on gitter. But there isn't any solution header only and also if libdispatch it is a new dependency it is considered quite a system library on OSX.

rolandpersson:
I have only used this on macOS and iOS where using libdispatch feels about the same as using <vector> etc in terms of dependencies.

It should be easy for someone to use on any system that happens to have libdispatch though.

bhack:
@rolandpersson Is this PR require linking?

rolandpersson:
@bhack On macOS and iOS libdispatch is like a standard library that gets linked into basically everything by default.( I'm not sure it is even possible to not use it).

So, defining CNN_USE_GCD is enough, there is no need (on macOS and iOS) to explicitly link against libdispatch. The tiny-dnn cmake files etc build and link everything without changes on macOS.

tiny_dnn/config.h
```diff
@@ -31,6 +31,11 @@
 // #define CNN_USE_OMP
 
 /**
+ * define to enable Grand Central Dispatch parallelization
+ */
+//#define CNN_USE_GCD
```
Randl: Why this line is commented out?

Randl:
LGTM

edgarriba:
@rolandpersson Add a couple of lines to CMake in order to define USE_GCD
https://github.com/tiny-dnn/tiny-dnn/blob/master/CMakeLists.txt#L59
```
option(USE_GCD "Build tiny-dnn with GCD support" OFF)

---

if(USE_GCD AND APPLE)  # not sure how to check ios systems
    add_definitions(-DCNN_USE_GCD)
endif()
```

tiny_dnn/config.h
rolandpersson: This PR is only meant to add another optional flavour of parallel_for that uses GCD, much like the other already existing flavours like CNN_USE_OMP and CNN_USE_TBB above in the same file. I just added this in the same style. 

Maybe there is a better way?

Adding it but not commenting it out would enable the GCD version for all platforms and that wouldn't work.

It could be argued that CNN_USE_GCD would be a good default option on macOS so that people who just try tiny-dnn out on a Mac don't get the slower default implementation.

Randl: @rolandpersson See comment by @edgarriba , you need to add couple of lines to CMake

bhack:
Ok "automatically linked in when you build your program with Xcode or cc and ld." We need cmake linking only if we want to let user experiment on other platform trought third party.

tiny_dnn/util/parallel_for.h
```diff
@@ -31,6 +31,10 @@
 #include <thread>
 #endif
 
+#if defined(CNN_USE_GCD) && !defined(CNN_SINGLE_THREAD)
+#include <dispatch/dispatch.h>
+#endif
```
beru: `#if defined(CNN_USE_TBB)` then GCD version of `parallel_for` function will not appear after cpp preprocessing. That's alright but even in that case `<dispatch/dispatch.h>` header will be included and it will help consume tiny amount of electricity.

tiny_dnn/util/parallel_for.h
```diff
@@ -82,6 +86,31 @@ void parallel_for(int begin, int end, const Func &f, int /*grainsize*/) {
...
+  int count     = end - begin;
+  int blockSize = grainsize;
+  if (count < blockSize || blockSize == 0) {
+    blockSize = 1;
+  }
+  int blockCount = (count + blockSize - 1) / blockSize;
+  assert(blockCount > 0);
+
+  dispatch_apply(blockCount, dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0),
+                 ^(size_t block) {
```
beru: Not all compilers support the block syntax `^`, so I thought using `__OBJC__` macro to check if Objective C compiler is being used might be good.

Someone may want to use GCD with other compilers, but that's a minority so supporting it can be postponed.

beru:
LGTM, other people can edit `CMakeLists.txt` and `cmake/summary.cmake` with subsequent pull request.

tiny_dnn/util/parallel_for.h
rolandpersson: Yes, that's a good point. __OBJC__ won't work though  as it is only defined when actually compiling Objective-C or Objective-C++ and here we are compiling plain C++ (with blocks).

But, there is a non-block variant called dispatch_apply_f that can be used instead. It will look a bit uglier but is better for other platforms.

beru: I see, thank you for the info. I wish GCD will accept C++ lambda expression.

bhack: @beru lambda like http://opensource.mlba-team.de/xdispatch/docs/current/lambda__blocks_8h_source.html?

beru: @bhack I meant officiall GCD. But thank you for the info. I'll check XDispatch someday.

bhack: Ok see ["Use of Blocks" section](http://opensource.mlba-team.de/xdispatch/docs/current/group__xdispatch.html)

beru:
@rolandpersson Measured performance with my machine.

Tested env:
  Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
  System Version: OS X 10.11.6 (15G1217)
  Kernel Version: Darwin 15.6.0
  Apple LLVM version 8.0.0 (clang-800.0.42.1)

Tested revision:
  dd80632d7e807ca71991186e9951a93d441c2fc4

average time per epoch
program | #define CNN_USE_GCD | //#define CNN_USE_GCD
---|---------------|---------------
example_mnist_train | 10.9 (sec) | 29.3
example_cifar_train | 87 | 132

Splendid performance improvement!
I Should've done it myself earlier. Kudos to you.

Issue #459, opened 2016-12-15
Can't import DetectNet model from DIGITS


reunanen:
I have trained a [DetectNet](https://devblogs.nvidia.com/parallelforall/detectnet-deep-neural-network-object-detection-digits/) model in DIGITS, and I'm trying to import it to tiny-dnn.

Unfortunately, I can't import it – I get `nn_error("unsupported padding type")` [here](https://github.com/tiny-dnn/tiny-dnn/blob/341da33a62d9bac97f782420d4438e0e43137cb4/tiny_dnn/io/caffe/layer_factory_impl.h#L273). (`w_pad` is 1, `pool_size_w` is 3.)

Will try to find the time to study the problem and hopefully create a PR – but not entirely sure what is the correct way to fix this, so any tips would be greatly appreciated.

Also, not sure, but guessing the same might apply even to vanilla GoogLeNet trained in DIGITS (DetectNet is based on GoogLeNet).

reunanen:
For what it's worth, this comes from a max-pooling layer.

reunanen:
[This change](https://github.com/reunanen/tiny-dnn/commit/38adeae05f9bacc8f921ae712f20fa2ac5dddcdd) at least makes the import apparently succeed. But quite unsure about it, so not creating a PR for now.

bhack:
Any news? Can you propose a draft PR?

reunanen:
@bhack No news. As [I wrote](https://github.com/tiny-dnn/tiny-dnn/issues/459#issuecomment-267341135), [this change](https://github.com/reunanen/tiny-dnn/commit/38adeae05f9bacc8f921ae712f20fa2ac5dddcdd) made the import apparently succeed, but I have no clue if it breaks something else.

bhack:
Try to propose it as PR. We will put in the PR review queue.

Randl:
@reunanen After merge of https://github.com/tiny-dnn/tiny-dnn/pull/557 , this one can be closed, right?

bhack:
Can we close also https://github.com/tiny-dnn/tiny-dnn/issues/146?

reunanen:
@Randl Yes, I think so. Closing.

PR #557, opened 2017-02-10
Problem: Can't import DetectNet model from DIGITS. (#459)

reunanen:
Solution: Relax the checks for padding::same.

No clue if this creates more problems than it fixes. That said, now the model can at least be imported.

I didn't really want to create a PR when I have barely tested at all, but then @bhack [encouraged](https://github.com/tiny-dnn/tiny-dnn/issues/459#issuecomment-278607710) me to do it nevertheless, so here we go.

### DIFF
--- a/tiny_dnn/io/caffe/layer_factory_impl.h
+++ b/tiny_dnn/io/caffe/layer_factory_impl.h
@@ -244,19 +244,23 @@ inline std::shared_ptr<layer> create_pooling(const caffe::LayerParameter &layer,
   }
 
   if (w_pad != 0) {
-    if (w_pad == pool_size_w - 1 || w_pad == (pool_size_w - 1) / 2) {
+    if (w_pad == (pool_size_w - 1) / 2) {
       pad_type = padding::same;
     } else {
       throw nn_error("unsupported padding type");
     }
+    // NB: w_pad == pool_size_w - 1 could lead to pad_type = padding::full,
+    //     if such a type existed
   }
 
   if (h_pad != 0) {
-    if (h_pad == pool_size_h - 1 || h_pad == (pool_size_h - 1) / 2) {
+    if (h_pad == (pool_size_h - 1) / 2) {
       pad_type = padding::same;
     } else {
       throw nn_error("unsupported padding type");
     }
+    // NB: h_pad == pool_size_h - 1 could lead to pad_type = padding::full,
+    //     if such a type existed
   }
 
   if (pool_param.has_pool()) {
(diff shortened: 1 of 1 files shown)


Randl:
@edgarriba I haven't check it deeply, but it seems like `h_pad == pool_size_h - 1` is `full` padding, not `same`. Correct me if I'm mistaken.

cc @pansk @nyanp

tiny_dnn/io/caffe/layer_factory_impl.h
```diff
@@ -244,15 +244,15 @@ inline std::shared_ptr<layer> create_pooling(const caffe::LayerParameter &layer,
   }
 
   if (w_pad != 0) {
-    if (w_pad == pool_size_w - 1) {
+    if (w_pad == pool_size_w - 1 || w_pad == (pool_size_w - 1) / 2) {
```
Randl: I believe you should replace it with `if (w_pad == (pool_size_w - 1) / 2)` (same for `h_pad`)? As far as I understand, this would exactly check for `same` padding, while the current version checks for `full` padding. cc @edgarriba @beru @bhack @wangyida

reunanen: Sure, I can change it, if everyone agrees. Or you can.

Should `w_pad == pool_size_w - 1` lead to `padding::full` then?

bhack:
Yes

tiny_dnn/io/caffe/layer_factory_impl.h
Randl: It should, but it appears we dont have one https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/util.h#L62

bhack: You can add in another PR if you need that type.

reunanen: Right. Well, I don't need it.

bhack:
Why avx tests are failing on VS?

Randl:
Looks like a glitch

bhack:
Is it an hint for https://github.com/tiny-dnn/tiny-dnn/issues/570?

Randl:
@edgarriba Is it possible to rerun tests on AppVeyor? Pretty sure that the code is ok

edgarriba:
@Randl I don't have such permissions. /cc @nyanp

Randl:
@reunanen I'm really sorry that we have to bother you, but can you now revert the change to respect code style?

edgarriba:
@reunanen Code seems to work. Coding style doesn't accept this silly change.

reunanen:
Reverted.

Issue #545, opened 2017-02-07
Adopt semantic versioning


Randl:
[Semantic Versioning](http://semver.org/) or semver is a simple set of rules and requirements that dictate how version numbers are assigned and incremented. It makes versioning more intuitive and easier to understand.

bhack:
It is ok for me

edgarriba:
is it enough through CMake ?
https://github.com/edgarriba/tiny-cnn/blob/master/CMakeLists.txt#L11

bhack:
https://github.com/fakenmc/cmake-git-semver

bhack:
And https://github.com/toeb/cmakepp/blob/master/cmake/semver/README.md

Issue #457, opened 2016-12-14
proposal:use using instead of typedef


nocotan:
In C++11, we can now use  ```using``` to write type alias instead of ```typedef``` .
```using``` is easy to read than ```typedef```, and we can use template alias.
I think better to rewrite will be better.

Randl:
Disagreed. Changes for changes.

Randl:
cc @edgarriba @nyanp @beru @bhack

bhack:
What I see is https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rt-using

bhack:
Ops.. the right ancohor is https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#t43-prefer-using-over-typedef-for-defining-aliases

edgarriba:
@Randl add it to tasks list
https://github.com/tiny-dnn/tiny-dnn/projects/2

Randl:
@edgarriba It's in https://github.com/tiny-dnn/tiny-dnn/projects/1#card-1205550

beru:
I didn't know much about this `using` feature.
@bhack Thank you for the link.

Issue #449, opened 2016-12-09
Parametric activation funcions


jpauwels:
It would be nice if you could specify the parameters to certain activation functions. For example, the alpha parameter of leaky_relu is now fixed to 0.01, but I'd like to experiment with changing it. It's not trivial to change this, as the activation functions are types and floats can't be used as template parameters. I came up with the following hack: 
```
template <cnn_size_t Numerator=1, cnn_size_t Denominator=100>
class leaky_relu : public function {
public:
   using function::df;
   float_t f(const vec_t& v, cnn_size_t i) const override { return (v[i] > float_t(0)) ? v[i] : float_t(Numerator) / float_t(Denominator) * v[i]; }
   float_t df(float_t y) const override { return y > float_t(0) ? float_t(1) : float_t(Numerator) / float_t(Denominator); }
   std::pair<float_t, float_t> scale() const override { return std::make_pair(float_t(0.1), float_t(0.9)); }
 };
```
If you want, I can submit this as a PR, but I wanted to check first whether there are better solutions.

Randl:
Why not `template <float_t Alpha = 0.01>`?

jpauwels:
@Randl Try it :-) If that was valid C++, the whole issue would be non-existant. See also http://stackoverflow.com/questions/2183087/why-cant-i-use-float-value-as-a-template-parameter

Randl:
As I told [here](https://gitter.im/tiny-dnn/developers?at=5865190d058ca96737a42784), I have a bit more convenient idea.
What if we create very simplistic analogue of `std::ratio` with implicit constructor from `float_t`? And use it as a parameter for template. Then user can use float as template parameter I think.
@edgarriba @beru @wangyida

beru:
@Randl 
I'm not so sure if giving constant parameter as template parameter is a good idea.
Since it would be hard to dynamically instantiate and deserialize activation function in that way.

edgarriba:
I vote to have activation functions as independent ops. We can just make them to inherit from `layer` so that they can also have internal parameters.

Randl:
I see, this gets us back to serialization problem. Would adding parameters to activations result in significant slowdown? If so, maybe we want to introduce separate parametric functions?

Randl:
This could be done after decoupling of activations ( https://github.com/tiny-dnn/tiny-dnn/pull/529 ) will be finished

PR #564, opened 2017-02-10
Child classes of activation_layer for all currently supported activations.

karandesai-96:
This PR is a succession to #529 which takes forward the task of decoupling layers and activations. Prior to this PR, a new base class `activation_layer` was created, and here several child classes for other activations have been made and used in tests. 

### Scope of this PR:

1. Following new layers inheriting `tiny_dnn::activation_layer` have been created in `tiny_dnn` namespace, supporting all current activation functions of the library:
    * `sigmoid_layer`
    * `tanh_layer`
    * `tanh_p1m2_layer`
    * `relu_layer` (Created before in #529 )
    * `softmax_layer`
    * `leaky_relu_layer`
    * `elu_layer`

2. Refactor `make_mlp` method to construct an MLP with `fully_connected` layers templated with identity activation, alongwith desired activation as a layer. This will make dropping of template later easy.

3. Use these newly created layers in atleast one of the tests written in **test_network.h** to see if they work properly.

### After this PR:

1. Refactor complete codebase where occurrences of layers with templated activations are replaced by layers with identity activation as template and an attached activation layer after the original layer. I could have done it here, but it would make the PR diff huge and hard to review.

2. After this point, since all the existing layer instances will have identity activation as template parameter, the `fully_connected` layer can be experimented with and template can be removed. Some changes in fully connected OpKernels will have to be made for all backends to properly undertake this, as we will remove `vector_type::aux` from output data in absence of activation template.


test/test_network.h
```diff
@@ -453,14 +454,15 @@ TEST(network, gradient_check4) {  // sigmoid - cross-entropy
 
 TEST(network, gradient_check5) {  // softmax - cross-entropy
   typedef cross_entropy loss_func;
-  typedef softmax activation;
```
beru: typedef can still be used to switch activation function layer type.

tiny_dnn/activations/elu_layer.h
```diff
@@ -0,0 +1,36 @@
...
+#include "tiny_dnn/activations/activation_layer.h"
+#include "tiny_dnn/layers/layer.h"
+
+namespace tiny_dnn {
+
+class elu_layer : public activation_layer {
+ public:
+  using activation_layer::activation_layer;
+
+  std::string layer_type() const override { return "elu-activation"; }
```
beru: I personally think return type of `layer::layer_type()` should be `const char*` to prevent an instance creation.
But that's different topic.

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,49 @@
...
+
+  void forward_activation(const vec_t &x, vec_t &y) {
+    float_t alpha = *std::max_element(x.begin(), x.end());
+    float_t denominator(0);
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] = std::exp(x[j] - alpha);
+      denominator += y[j];
+    }
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] /= denominator;
```
beru: This loop should be written as
```cpp
    float_t recip_denominator = float_t(1) / denominator;
    for (serial_size_t j = 0; j < x.size(); j++) {
      y[j] *= recip_denominator;
    }
```
to reduce a number of expensive division operation.

tiny_dnn/activations/tanh_p1m2_layer.h
```diff
@@ -0,0 +1,38 @@
...
+      ep   = std::exp(x[j]);
+      y[j] = ep / (ep + (float_t(1) / ep));
+    }
+  }
+
+  void backward_activation(const vec_t &x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) {
+    for (serial_size_t j = 0; j < x.size(); j++) {
```
beru: Why the name of loop counter variable is `j`? `i` seems more better to me.

tiny_dnn/activations/tanh_p1m2_layer.h
```diff
@@ -0,0 +1,38 @@
...
+      // dx = dy * (gradient of tanh-scaled)
+      dx[j] = dy[j] * (2 * y[j] * (float_t(1) - y[j]));
```
beru: `y[j]` is repeatedly used in here. While compilers will optimize it, how about using a local variable?

beru:
@karandesai-96 I've noticed you put your name into file header copyright section.

To simplify file headers, I think it would be good only put project first author's name. (nyanp)
Your name will eventually be listed in https://github.com/tiny-dnn/tiny-dnn/blob/master/AUTHORS file.
And LICENSE file also mentions about other contributors copyright. After all, git log contains every contributors activities.

It is just my personal opinion but I don't want to see loads of names of contributors splattered to every header files.

beru:
@karandesai-96 I've noticed original `activation::function` class has this method.
```cpp
  // target value range for learning
  virtual std::pair<float_t, float_t> scale() const = 0;
```
And inherited classes override it.

It looks to me that one is missing in new activation function layer classes.
I guess adding virtual `scale` method to `activation_layer` class and overriding it might be needed.
Also, all existing code that calls `scale` method of activation function needs to be taken care of.

tiny_dnn/activations/elu_layer.h
```diff
@@ -0,0 +1,36 @@
...
+
+  void forward_activation(const vec_t &x, vec_t &y) {
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] = x[j] < float_t(0) ? (std::exp(x[j]) - float_t(1)) : x[j];
```
beru: `x[j]` appears 3 times to access same value. I would use a local variable in this case.

tiny_dnn/activations/elu_layer.h
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  void backward_activation(const vec_t &x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) {
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      // dx = dy * (gradient of elu)
+      dx[j] = dy[j] * (y[j] > float_t(0) ? float_t(1) : (float_t(1) + y[j]));
```
beru: `y[j]` appears 2 times here, I'd use a local variable to hold it.

beru: Oh well, but it is same in original code..

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  void backward_activation(const vec_t &x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) {
+    const serial_size_t len = static_cast<serial_size_t>(dy.size());
```
beru: I think `cnt` parameter should be added to `forward_activation` and `backward_activation` methods to represent length of vec_t type parameters.

Each of `vec_t` instances content length has to be same. I think it is a bit inappropriate to use lengths from several instances `size()` method as it makes ambiguousness.

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,49 @@
...
+
+    // auxilliary vector to store element wise softmax gradients of all elements
+    vec_t df(len, 0);
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      for (serial_size_t k = 0; k < x.size(); k++) {
+        df[k] = (k == j) ? y[j] * (float_t(1) - y[j]) : -y[k] * y[j];
```
beru: `y[j]` is used 3 times in here, it should be stored to local variable before inner loop starts.

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  void backward_activation(const vec_t &x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) {
+    const serial_size_t len = static_cast<serial_size_t>(dy.size());
+
+    // auxilliary vector to store element wise softmax gradients of all elements
+    vec_t df(len, 0);
```
beru: Adding `thread_local` specifier to `vec_t df` local variable and resizing it later might improve performance as it reduces memory allocation/deallocation. This method will frequently be called during training.

tiny_dnn/activations/tanh_p1m2_layer.h
```diff
@@ -0,0 +1,38 @@
...
+ public:
+  using activation_layer::activation_layer;
+
+  std::string layer_type() const override { return "tanh-scaled-activation"; }
+
+  void forward_activation(const vec_t &x, vec_t &y) {
+    float_t ep;
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      ep   = std::exp(x[j]);
+      y[j] = ep / (ep + (float_t(1) / ep));
```
beru: :+1: :ok_hand: :clap:

tiny_dnn/activations/softmax_layer.h
karandesai-96: @beru `y[i] /= denominator;` what is the original value of `y[i]` ?

karandesai-96: Actually the value is `std::exp(x[i] - alpha) / denominator`, I wrote the implementation as it is right now to prevent costly calls to `std::exp` again in second for loop.

tiny_dnn/activations/tanh_p1m2_layer.h
karandesai-96: @bery I named it `j` just because it iterates over a `vec_t` and the outer loop which iterates over `tensor_t` in `forward_propagation` or `backward_propagation` have a counter named `i`. It won't cause any errors, but I thouht it might be just better this way. Does it help though ? If not much, then I'll simply use `i`.

karandesai-96:
@beru:

> @karandesai-96 I've noticed you put your name into file header copyright section.

I saw @edgarriba 's name at a few places in the project in this manner, so I assumed that it might be mandatory to specify name if an author is the one who first created the file to keep a track. 

> It is just my personal opinion but I don't want to see loads of names of contributors splattered to every header files.

Now that you clarified that not having the names is preferred, I'll remove them because I personally found it cumbersome to keep on mentioning at places.

tiny_dnn/activations/tanh_p1m2_layer.h
beru: I see. I just thought it is somewhat peculiar to see `j` counter with single loop.

beru:
@karandesai-96 
> I saw @edgarriba 's name at a few places in the project in this manner, so I assumed that it might be mandatory to specify name if an author is the one who first created the file to keep a track.

If it is mandatory, I'd like to put my name into several places. But let's not.

> Now that you clarified that not having the names is preferred, I'll remove them because I personally found it cumbersome to keep on mentioning at places.

It's just my personal opinion. So I can't force anyone to follow it.
I've recently started regarding 5 letters nyanp as a group mind generated from avitivities of tiny-dnn contributors just to have peace of mind.

tiny_dnn/activations/softmax_layer.h
beru: Maybe you are right or not. I initially thought using reciprocal multiplication would reduce computing cost.

karandesai-96:
@beru I just added a `scale` virtual method and made all child classes override it.

edgarriba:
@beru Regarding License you're right, at that time I thought the same as @karandesai-96 and nobody was against that. Said that, license should be simplified in every file and refer it to the one in the root directory.

tiny_dnn/activations/softmax_layer.h
```diff
@@ -24,8 +24,10 @@ class softmax_layer : public activation_layer {
       y[j] = std::exp(x[j] - alpha);
       denominator += y[j];
     }
+
+    float_t recip_denominator = float_t(1) / denominator;
```
edgarriba: When values will not be changed I tend to mark as const. Speed compilation and benefits readability

examples/main.cpp
```diff
@@ -131,9 +131,9 @@ void sample2_mlp(const string& data_dir) {
 #if defined(_MSC_VER) && _MSC_VER < 1800
   // initializer-list is not supported
   int num_units[] = {28 * 28, num_hidden_units, 10};
-  auto nn         = make_mlp<tan_h>(num_units, num_units + 3);
+  auto nn         = make_mlp<tanh_layer>(num_units, num_units + 3);
```
edgarriba: Does it mean that we are already dropping current template approach. No backward compatibility?

test/test_network.h
```diff
@@ -485,7 +487,7 @@ TEST(network, gradient_check6) {  // sigmoid - cross-entropy
 
 TEST(network, gradient_check7) {  // leaky-relu - mse
   typedef mse loss_func;
-  typedef leaky_relu activation;
+  typedef leaky_relu_layer activation;
```
edgarriba: You might want to use `using` feature here
proposal:use using instead of typedef

tiny_dnn/activations/activation_layer.h
```diff
@@ -101,6 +101,11 @@ class activation_layer : public layer {
                                    vec_t &dx,
                                    const vec_t &dy) = 0;
 
+  /**
+   * Target value range for learning.
+   */
+  virtual std::pair<float_t, float_t> scale() const = 0;
```
edgarriba: Do all the activations have scale? Is it really needed a pure virtual method?

tiny_dnn/activations/elu_layer.h
edgarriba: Too C using `const char*` no? :smile:

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,58 @@
...
+namespace tiny_dnn {
+
+class softmax_layer : public activation_layer {
+ public:
+  using activation_layer::activation_layer;
+
+  std::string layer_type() const override { return "softmax-activation"; }
+
+  void forward_activation(const vec_t &x, vec_t &y) override {
+    float_t alpha = *std::max_element(x.begin(), x.end());
```
edgarriba: Use const

examples/main.cpp
Randl: @edgarriba I think we need to drop backward compatibility. Anyway that's gonna happen (once again) on switch to `Tensor`. BTW, this means we better drop back to 0.*.* if we adopt https://github.com/tiny-dnn/tiny-dnn/issues/545

test/test_network.h
Randl: Agreed. See https://github.com/tiny-dnn/tiny-dnn/issues/457

tiny_dnn/activations/elu_layer.h
Randl: `std::string` is OK I think. Overhead is negligible.

tiny_dnn/activations/elu_layer.h
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  void backward_activation(const vec_t &x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) override {
+    float_t yj;
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      yj = y[j];
```
Randl: Why do you assign `y[j]` to separate variable?

tiny_dnn/activations/elu_layer.h
```diff
@@ -0,0 +1,44 @@
...
+      // dx = dy * (gradient of elu)
+      dx[j] = dy[j] * (yj > float_t(0) ? float_t(1) : (float_t(1) + yj));
```
Randl: `dx[j] = dy[j] + (yj > float_t(0) ? float_t(0) : (dy[j] * y[j]));`

What about this one?

tiny_dnn/activations/leaky_relu_layer.h
```diff
@@ -0,0 +1,42 @@
...
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
+#pragma once
+#include "tiny_dnn/activations/activation_layer.h"
+#include "tiny_dnn/layers/layer.h"
+
+namespace tiny_dnn {
+
+class leaky_relu_layer : public activation_layer {
```
Randl: Can we add parameters to activations, as described here https://github.com/tiny-dnn/tiny-dnn/issues/449 ?

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,58 @@
...
+    float_t denominator(0);
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] = std::exp(x[j] - alpha);
+      denominator += y[j];
+    }
+
+    float_t recip_denominator = float_t(1) / denominator;
+    for (serial_size_t j = 0; j < x.size(); j++) {
+      y[j] *= recip_denominator;
```
Randl: Why not divide by `denominator`?

tiny_dnn/activations/softmax_layer.h
```diff
@@ -0,0 +1,58 @@
...
+
+  void backward_activation(const vec_t &x,
+                           const vec_t &y,
+                           vec_t &dx,
+                           const vec_t &dy) override {
+    const serial_size_t len = static_cast<serial_size_t>(dy.size());
+    float_t yj;
+
+    // auxilliary vector to store element wise softmax gradients of all elements
+    thread_local vec_t df;
```
Randl: This seems to cause compilation error in OS X. See https://stackoverflow.com/questions/23791060/c-thread-local-storage-clang-503-0-40-mac-osx

tiny_dnn/activations/elu_layer.h
bhack: I think that isocpp guidelines has some good performance principles. One of this is "Don't optimize not performance critical parts". I also think that here we could be involved in RVO or not?

tiny_dnn/activations/elu_layer.h
karandesai-96: @beru suggested that way, if a value is referenced through index multiple times.

tiny_dnn/activations/elu_layer.h
karandesai-96: This is less calculation, I would do that but my doubt is whether it looks intuitive ? I mean `dx = dy * (gradient of elu)`. The equation format is maintained throughout other classes as well. Looks perfectly like applying chain rule while carrying out derivative.

tiny_dnn/activations/leaky_relu_layer.h
karandesai-96: I will do that in separate PR. Current PR just replicated classes from **tiny_dnn/activations/activation_function.h** as layers.

karandesai-96: ^ Same reply to @edgarriba, I just replicated a virtual method `scale` and gave it to all child classes, like earlier. We can incrementally improve for sure.

tiny_dnn/activations/softmax_layer.h
karandesai-96: @beru says division is costlier than multiplication. I had it multiplied just a few commits ago.

tiny_dnn/activations/softmax_layer.h
karandesai-96: I guess then I will have to revert back

tiny_dnn/activations/elu_layer.h
Randl: I see. Could you check what is difference in speed and we'll make a decision based on it?

tiny_dnn/activations/softmax_layer.h
Randl: @beru This seems like a trivial optimization for compiler, I'm not sure we should do that. Same partially applies for https://github.com/karandesai-96/tiny-dnn/blob/ea3b0da23619fe8b0c8a560fb25fca247389bf85/tiny_dnn/activations/elu_layer.h#L34

tiny_dnn/activations/softmax_layer.h
Randl: Check solutions suggested at stackoverflow, for example http://stackoverflow.com/a/25393790/1894478

tiny_dnn/activations/elu_layer.h
bhack: We can use a general rule to prefer readbility vs microptimization if it is not critical after we get timing with optimized compiler flags.

tiny_dnn/activations/softmax_layer.h
bhack: It is supported since the clang compiler included with the Xcode 8 Beta

tiny_dnn/activations/elu_layer.h
bhack: Are we sure that we need this microptimizzation?

examples/main.cpp
beru: @edgarriba I also would like to drop backward compatibility to improve the library with ease.

But before then, there are several kinds of backward compatibilities, such as API, binary file format.
How much are they important? Will we need to announce a change in release notes?

tiny_dnn/activations/activation_layer.h
beru: At least, original `tiny_dnn::activation::function::scale` is a pure virtual method.

tiny_dnn/activations/softmax_layer.h
beru: > Too C using const char* no?
if you really want to hide a naked pointer, adding `string_ref` class can be a solution.
But I think `const char*` is simple enough in this case.

tiny_dnn/activations/elu_layer.h
beru: @bhack It depends on which compiler you use. I don't think using a local variable severely hurts code readability in this case. If we really prefer concise code, we wouldn't use the batch interface in the first place, would we?

tiny_dnn/activations/softmax_layer.h
beru: @Randl @karandesai-96 Doing reciprocal multiplication slightly decrease floating point precision so compilers may not do it automatically. (It depends on cases though)
http://stackoverflow.com/questions/35506226/will-the-compiler-optimize-division-into-multiplication/35506382#35506382

tiny_dnn/activations/elu_layer.h
beru: For me, it's too old-C++.

examples/main.cpp
edgarriba: I'm fine with dropping templated layers but yes, as you said we should make a pre-release with it

tiny_dnn/activations/activation_layer.h
edgarriba: yeah, but most of them returns same values. Why not just override the ones that are really needed

beru: I can understand that.

tiny_dnn/activations/softmax_layer.h
beru: Once travis-ci OSX jobs has passed, LGTM

beru: If local variable is not static, using below macro can be a solution to support OSX.
```cpp
#if HAS_CXX11_THREAD_LOCAL
#define ATTRIBUTE_TLS thread_local
#else
#define ATTRIBUTE_TLS 
#endif
```

It might be confusing though.... I wish we can drop OSX and Windows support.

bhack: Do you to remove OSX and Windows support? It is not the first time that I see this kind of comment. Use emoticons if you want to be ironic.

beru: I do want to remove Apple on Earth. This damn MacBook keyboard I'm using right now is hard to use. :tired_face:

tiny_dnn/activations/activation_layer.h
beru: @edgarriba An advantage of making base method pure virtual is new activation function layer class writer would pay attention to important scale value. (Comment says it is target value range for learning, but it's all Greek to me...)
Contrary, an advantage of making base method non-pure virtual is implementors can have the luxury of object-oriented programming language.

tiny_dnn/activations/softmax_layer.h
edgarriba: @beru there's no plan to drop support for OSX and Windows. Actually, I think it's a valuable feature of this library. Please, don't confuse possible external readers. If you think the contrary, open a ticket so that we can discuss there.

tiny_dnn/activations/softmax_layer.h
Randl: Asserting equal lengths might be a solution

tiny_dnn/activations/activation_layer.h
karandesai-96: > (Comment says it is target value range for learning, but it's all Greek to me...)

I would also like to understand this, just kept the docstring because it was _there_ before.

tiny_dnn/activations/softmax_layer.h
```diff
@@ -35,8 +35,15 @@ class softmax_layer : public activation_layer {
...
-    vec_t df(len, 0);
+// auxilliary vector to store element wise softmax gradients of all elements
+
+#if HAS_CXX11_THREAD_LOCAL
+#define ATTRIBUTE_TLS thread_local
+#else
+#define ATTRIBUTE_TLS
+#endif
+
+    ATTRIBUTE_TLS vec_t df(len, 0);
```
beru: Since this is library is header only, it's a bit difficult to isolate preprocessor simbols defined in library header files.
So, using `#undef ATTRIBUTE_TLS` or simply writing in this can be preferrable.

```cpp
#if HAS_CXX11_THREAD_LOCAL
    thread_local
#endif
    vec_t df(len, 0);
```

Or putting preprocessor macros in `util/macro.h` would be good. But this `thread_local` matter is a bit complicated as discussed earlier. Why Apple didn't release C++11 conformant compiler sooner!

karandesai-96: @beru I'm using this snippet for now. We can discuss the possibilities of using `thread_local` at other places - probably in master, instead of this branch too.

beru: @karandesai-96  I see. Then LGTM.

edgarriba:
LGTM, let's merge

Issue #538, opened 2017-02-01
change batch size in mnist example got wrong answer


fastalgo:
I changed the batch size of mnist.cpp example from 10 to 64.

https://github.com/tiny-dnn/tiny-dnn/blob/master/examples/mnist/train.cpp

Then I got a wrong answer. Why?

pliptor:
@Randl Could you point me out where to read on the relation between dropout and batch sizes? Thanks!

Randl:
@pliptor As for this bug, I suppose the proble is in batch and not in dropout, just with droupout it causes crash and with different settings it causes wrong results.

Generally I believe dropout and batch size are not very related.

pliptor:
Thank you. It seems both issues are very important to be solved and I wanted to learn more about them. Let me know if there's anything I could help with.

Randl:
We need to understand what causes the bug and to localize it first. If you have any insights, feel free to share them.

pliptor:
I tried to look at the dropout code earlier but no clue so far. Will keep trying.

bhack:
@Randl I think it is safer to have a test on the batch size dimension

pliptor:
The learning rate of the optimizer gets multiplied by sqrt(batch dimension) in the mnist code. For a batch size of 64 that's 8. It might be too fast. Let me see.

fastalgo:
@pliptor Thanks so much!

pliptor:
@fastalgo  My machine is slow so I can't simulate fast I'm getting good results with batch size of 64 by limiting the multiplying factor with std::max(float_t(4),...). The 4 came from the fact that a size of 16 was working fine for me. But this is a hack. There might be better things to do here.

fastalgo:
@pliptor I got it, have a nice day :-)

Randl:
@pliptor Well those issues might be not connected after all.

As for limiting learning rate, I don't see something much better than that.

pliptor:
It's a hack but I can submit a patch for mnist if you want.

Randl:
If you can open PR, we'll discuss it there.

Randl:
@pliptor Probably, just add it to https://github.com/tiny-dnn/tiny-dnn/pull/574

pliptor:
Good idea.

Randl:
@fastalgo Can you check with latest master?

fastalgo:
@Randl Yes, it is working, thanks!

PR #574, opened 2017-02-12
MNIST enhanced performance. Faster convergence, better score

pliptor:
This patch credit goes to @beru in my experiments with the activation function in the last layer of the network. This is just a one line patch in the training file for the MNIST example that does not touch the network configuration. It only changes the batch size from 10 to 16. 

Convergence is at least 10% faster in my AMD Quad-Core A6 with SSE and TBB enabled.
Final classification accuracy is 99.11%. Results attached.

It appears that AVX must be disabled in order to achieve the above classification  accuracy. Thanks.

[mninst-train-batch16.txt](https://github.com/tiny-dnn/tiny-dnn/files/769672/mninst-train-batch16.txt)

Added one more patch to address the issue of users testing larger mini-batch sizes #538 
The current training code for MNIST example scales the learning rate of the optimizer by the
square root of the mini-batch size. This negatively impacts the convergence if the batch size is large. The first patch in this pull changed the batch size to 16 with positive results. Hence 
the limiting factor has been changed to the minimum of sqrt(16)=4 and the square root of the mini-batch size. Thanks.

### DIFF
--- a/examples/mnist/train.cpp
+++ b/examples/mnist/train.cpp
@@ -79,7 +79,7 @@ static void train_lenet(const std::string& data_dir_path) {
 
   progress_display disp(static_cast<unsigned long>(train_images.size()));
   timer t;
-  int minibatch_size = 10;
+  int minibatch_size = 16;
   int num_epochs     = 30;
 
   optimizer.alpha *= static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size));
--- a/examples/mnist/train.cpp
+++ b/examples/mnist/train.cpp
@@ -82,7 +82,9 @@ static void train_lenet(const std::string& data_dir_path) {
   int minibatch_size = 16;
   int num_epochs     = 30;
 
-  optimizer.alpha *= static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size));
+  optimizer.alpha *=
+    std::min(tiny_dnn::float_t(4),
+             static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size)));
 
   // create callback
   auto on_enumerate_epoch = [&]() {


examples/mnist/train.cpp
```diff
@@ -79,10 +79,11 @@ static void train_lenet(const std::string& data_dir_path) {
 
   progress_display disp(static_cast<unsigned long>(train_images.size()));
   timer t;
-  int minibatch_size = 10;
+  int minibatch_size = 16;
   int num_epochs     = 30;
 
-  optimizer.alpha *= static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size));
+  optimizer.alpha *= std::min(
+    float_t(4), static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size)));
```
Randl: @edgarriba @bhack Any ideas for better solution? If not, LGTM

Randl: `std::min(tiny_dnn::float_t(4), static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size)));`

pliptor: Yep. What's the best way to fix the repo? Reset back one commit and recommit or add another one?

Randl: Whatever is convenient for you. Since we're squashing PRs in one commit, there is no much difference for us

pliptor:
@Randl Thanks for your patience. I tried to consolidate them for cleanness but they got split. I think I can still rebase it but it this is okay I'll leave it as is.

pliptor:
Thank you @Randl

examples/mnist/train.cpp
```diff
@@ -79,10 +79,12 @@ static void train_lenet(const std::string& data_dir_path) {
...
   progress_display disp(static_cast<unsigned long>(train_images.size()));
   timer t;
-  int minibatch_size = 10;
+  int minibatch_size = 16;
   int num_epochs     = 30;
 
-  optimizer.alpha *= static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size));
+  optimizer.alpha *=
+    std::min(tiny_dnn::float_t(4),
+             static_cast<tiny_dnn::float_t>(std::sqrt(minibatch_size)));
```
beru: I'm not sure how this change is needed. Aside from that, `std::sqrt` function has some overloads.
http://en.cppreference.com/w/cpp/numeric/math/sqrt, so rather than casting returned value, casting int argument to `tiny_dnn::float_t` might be another way to suppress a warning.

Randl: For bigger batches, learning rate becomes too high, so we need to limit it somehow

bhack: For small batch  https://github.com/tensorflow/tensorflow/issues/7476

edgarriba:
it's okay to have good examples showing good performance, but I don't think we should care about optimizing time with this ... as a library (like others) in training mode the learning rate policy is up to the user.

edgarriba:
LGTM

pliptor:
@edgarriba Thanks. You have a good point. I forked the library to play instead with optimization over there.  This library is however very awesome for a learning platform for those that want C++. I'm already getting better than 99.2% with half the complexity of the example by just playing with it for a couple of days.

edgarriba:
@pliptor thx for the feedback. In case you get more improvements don't hesitate to report them. Besides, could you add short comments in source showing where this formula comes from, or just cite if you got it from a paper. At first glance, it looks a magic number.

examples/mnist/train.cpp
pliptor: @bhack The conclusions of the paper end with "We have demonstrated that Batch Normalization, while effective, is not well suited to small or non-i.i.d. training minibatches."

I tried to track down with blame what was the rationale for line 85 but maybe only the original author knows it. It seems mini-batch processing reduces the learning rate by O(sqrt(batch_size)) (I'll try to find the paper). I believe that's why that multiplying constant was added to it. I mentioned @Randl this fix is a hack to address a user that was having trouble with the code by playing with the mini-batch size.

pliptor:
@edgarriba No, unfortunately it appears to me line 95 was a hack fix for a hack :) There is a paper that I did find trying to find the reason behind original line. I'll try to find it.

bhack:
https://github.com/NVIDIA/DIGITS/issues/227

pliptor:
@bhack No, that was not the paper I saw. The one I saw had analytical derivation. The one you are pointing out seem empirical. 

The empirical :) observation that I had while playing with the code is that it is clear that mini-bathing 
reduces convergence rate. So I found line 95 justified. But I find it strange that sqrt(mini-batch) is already bad for batches larger than 16 or so with this setup. multiplying by mini-batch would make it even worse. But note there is another important variable (or two) the optimization algorithm itself and the initial learning rate value. So it doesn't seem to me there is a simple answer here if you take these all into consideration.

pliptor:
Here's the paper.

http://www.cs.cmu.edu/~muli/file/minibatch_sgd.pdf

One really weird thing in the tests was the jump at epoch 16 for batch size 64. So initially batch size 64 looked very bad but eventually it caught up with batch 16.

examples/mnist/train.cpp
pliptor: @beru In numbers, say we pick minibatch_size = 16. The default start learning rate set by tiny-dnn in adagrad is alpha=0.01. We end up with alpha = 0.04 with the HEAD and it the algorithm takes it well: convergence speeds up without stalling it. But alpha bigger than that is making the algorithm to stall. 

Regarding the casting, yes, either way suppresses the warning although the scaling value would be different. The way you suggest would make the alpha scaling more conservative because the sqrt would be floored.  Either way is fine with me. Let me know.

beru: @Randl Thank you for the answer! :bow: I haven't studied about neural networks so there're a lot to learn. 

@pliptor Than you for the detailed explanation. :joy: It's pretty helpful. I'm not sure if maximum of alpha should be adjusted per network. Talking about the casting, I need to check its consequences.

pliptor: @beru This summary looks nice http://sebastianruder.com/optimizing-gradient-descent/

I agree with @edgarriba  it's not worth spending too much time in the optimization for the sake of writing the library. Maybe the easiest way to close this subject would be to run two simulations with each of the casting and pick the one that gives better result?

edgarriba:
@beru give green light so that I can merge

examples/mnist/train.cpp
beru: @pliptor Thank you for introducing the page. Actually, I searched about the optimization and found several informative pages such as these and also the one you introduced.
http://cs231n.github.io/optimization-1/
http://int8.io/comparison-of-optimization-techniques-stochastic-gradient-descent-momentum-adagrad-and-adadelta/

I wish I can understand basics of deep learning someday...

pliptor: @beru  I like those two links you are referencing. Time to study them!

PR #573, opened 2017-02-12
Drop template completely from fully_connected_layer.

karandesai-96:
This PR is a succession to #564 and aims to drop templating from `fully_connected_layer`. We inherit directly from `layer` instead of `feedforward_layer`. The `out_shape` no longer needs `{vector_type::aux}` anymore as the activations are no longer a part of this layer.

The usages of this layer in tests have been modified accordingly.

PR #577, opened 2017-02-14
Serialization of activation layers.

karandesai-96:
This PR is a part of series of PRs made for decoupling activation templates and layers. It is a succession of unmerged #573 but is meant to be merged before #573 for that to work smoothly.

### Scope of this PR:

* Serialize existing activation layers. An example of JSON output of an activation layer looks like this:
```json
"type": "tanh"
"in_size": {
    "width": 256,
    "height": 256,
    "depth": 3
}
```

* Appropriate functions have been added in a friend struct named `serialization_buddy` and these new activation layers have been registered in **serialization_layer_list.h** using `CNN_REGISTER_LAYER` macro.
* Corresponding tests have been added.

### Further Work:

* Cereal docs are a little bit complicated, but they mention on serializing objects from classes having supporting polymorphic relationship. This may help in reducing the code duplication, but it time consuming so can be taken up after decoupling, as the serialization code is yet to be changed a lot.

* Once this is in place, get #573 merged.

### DIFF
--- a/tiny_dnn/util/serialization_functions.h
+++ b/tiny_dnn/util/serialization_functions.h
@@ -243,10 +243,82 @@ struct LoadAndConstruct<tiny_dnn::slice_layer> {
 };
 
 template <>
+struct LoadAndConstruct<tiny_dnn::sigmoid_layer> {
+  template <class Archive>
+  static void load_and_construct(
+    Archive& ar, cereal::construct<tiny_dnn::sigmoid_layer>& construct) {
+    tiny_dnn::shape3d in_shape;
+
+    ar(cereal::make_nvp("in_size", in_shape));
+    construct(in_shape);
+  }
+};
+
+template <>
+struct LoadAndConstruct<tiny_dnn::tanh_layer> {
+  template <class Archive>
+  static void load_and_construct(
+    Archive& ar, cereal::construct<tiny_dnn::tanh_layer>& construct) {
+    tiny_dnn::shape3d in_shape;
+
+    ar(cereal::make_nvp("in_size", in_shape));
+    construct(in_shape);
+  }
+};
+
+template <>
 struct LoadAndConstruct<tiny_dnn::relu_layer> {
   template <class Archive>
   static void load_and_construct(
     Archive& ar, cereal::construct<tiny_dnn::relu_layer>& construct) {
+    tiny_dnn::shape3d in_shape;
+
+    ar(cereal::make_nvp("in_size", in_shape));
+    construct(in_shape);
+  }
+};
+
+template <>
+struct LoadAndConstruct<tiny_dnn::softmax_layer> {
+  template <class Archive>
+  static void load_and_construct(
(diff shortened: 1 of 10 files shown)


karandesai-96:
Hi @Randl @beru, I am ready for reviews on this PR.

tiny_dnn/util/serialization_functions.h
```diff
@@ -391,41 +391,6 @@ struct specialize<Archive,
                   tiny_dnn::slice_layer,
                   cereal::specialization::non_member_serialize> {};
 
-template <class Archive>
-struct specialize<Archive,
-                  tiny_dnn::sigmoid_layer,
-                  cereal::specialization::non_member_serialize> {};
-
```
edgarriba: @karandesai-96 why do you remove this? I propose to adopt a templated routine instead
```cpp
template <class Archive, typename L>
  void serialize(Archive& ar, L& layer) {
    serialization_buddy::serialize(ar, layer);
}
```

karandesai-96: @edgarriba Serialization works perfectly without these lines. They weren't hit even once, as per coverage report as well.

karandesai-96:
@beru @edgarriba I removed those structs and methods because coveralls report stated that they weren't used. Moreover, I tested locally, both serialization and deserialization works perfectly without them. I am unable to understand why is keeping them necessary.

edgarriba:
@karandesai-96 good catch. @beru why is that? you implemented those routines, right?

beru:
@karandesai-96 I think you need to add test cases to use new layers in `test/test_serialization.h` to verify validity of those structs and methods. My guess is those are not instantiated becase they aren't referenced at all.

edgarriba:
LGTM

@beru give approval so that I can merge

beru:
LGTM

tiny_dnn/util/serialization_functions.h
```diff
@@ -242,6 +242,90 @@ struct LoadAndConstruct<tiny_dnn::slice_layer> {
   }
 };
 
+template <>
+struct LoadAndConstruct<tiny_dnn::sigmoid_layer> {
+  template <class Archive>
+  static void load_and_construct(
```
edgarriba: @karandesai-96 @beru could we reuse code with templated parameters here and below files?

karandesai-96: @edgarriba I am unsure about this. I don't understand yet how cereal wants the design of these functions.

beru: @edgarriba @karandesai-96 Yeees, we can always use preprocessor macros to reduce redundant template c++ code and dig deeper into nightmarish unreadable code hell. :rage1:

beru: Please read
https://github.com/tiny-dnn/tiny-dnn/blob/master/cereal/access.hpp#L42
to understand why `LoadAndConstruct<serialized_type>::load_construct` methods need to be writen.

PR #580, opened 2017-02-15
Non contiguous tensor

edgarriba:
- Removed kDim templated parameter
- Added strided tensor

### DIFF
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -75,7 +75,6 @@ namespace tiny_dnn {
  * Unmutable if kConst == true
  */
 template <typename U         = float_t,
-          size_t kDimensions = 4,
           bool kConst        = false,
           typename Allocator = aligned_allocator<U, 64>>
 class Tensor {
@@ -98,22 +97,8 @@ class Tensor {
    * @return
    */
   Tensor() {
-    offset_ = size_ = size_t(0);
+    offset_ = size_ = stride_ = size_t(0);
     storage_ptr_    = std::make_shared<TensorStorageType>();
-  }
-
-  /**
-   * Constructor that assepts an array of shape and create a Tensor with that
-   * shape. For example, given shape = {2,3,4,5,6}, tensor
-   * will be of size 2x3x4x5x6
-   * @param shape array containing N integers, sizes of dimensions
-   * @return
-   */
-  explicit Tensor(const std::array<size_t, kDimensions> &shape) {
-    offset_      = size_t(0);
-    shape_       = shape;
-    size_        = product(shape);
-    storage_ptr_ = std::make_shared<TensorStorageType>(shape);
   }
 
   /**
@@ -125,22 +110,9 @@ class Tensor {
    */
   explicit Tensor(const std::vector<size_t> &shape) {
     offset_ = size_t(0);
+    stride_ = size_t(0);
     size_   = product(shape);
-    std::copy(shape.begin(), shape.end(), shape_.begin());
-    storage_ptr_ = std::make_shared<TensorStorageType>(shape);
-  }
-
-  /**
-   * Constructor that assepts an initializer list of shape and create a
-   * Tensor with that shape. For example, given shape = {2,3,4,5,6}, tensor
-   * will be of size 2x3x4x5x6
-   * @param shape array containing N integers, sizes of dimensions
-   * @return
-   */
-  explicit Tensor(std::initializer_list<size_t> const &shape) {
-    offset_ = size_t(0);
-    size_   = product(shape);
-    std::copy(shape.begin(), shape.end(), shape_.begin());
+    shape_  = shape;
     storage_ptr_ = std::make_shared<TensorStorageType>(shape);
   }
 
@@ -174,7 +146,7 @@ class Tensor {
    *
    * @return the tensor shape
    */
-  const std::array<size_t, kDimensions> &shape() const { return shape_; }
+  const std::vector<size_t> &shape() const { return shape_; }
 
   /**
    * Checked version of access to indexes in tensor (throw exceptions
@@ -222,16 +194,20 @@ class Tensor {
    */
   template <typename... Args>
   size_t host_pos(const size_t d, const Args... args) const {
-    static_assert(sizeof...(args) < kDimensions, "Wrong number of dimensions");
-    size_t dim = kDimensions - sizeof...(args) - 1;
+    const size_t dim = dims() - sizeof...(args) - 1;
     if (d >= shape_[dim]) {
       throw nn_error("Access tensor out of range.");
     }
+    
     size_t shift = 1;
-    for (size_t i = dim + 1; i < kDimensions; ++i)
+    for (size_t i = dim + 1; i < dims(); ++i) {
       shift *= shape_[i];  // TODO(Randl): optimize. Reverse argumets?
+    }
 
-    return (d * shift + host_pos(args...));
+    // apply stride in case that we are in the last two dimensions
+    const size_t total_shift = (dim > 2) ? shift : shift + stride_;
+    
+    return (d * total_shift + host_pos(args...));
   }
 
   template <typename... Args>
@@ -242,7 +218,6 @@ class Tensor {
   template <typename... Args>
   StorageIterator host_iter(const Args... args) const {
     static_assert(!kConst, "Non-constant operation on constant Tensor");
-    static_assert(sizeof...(args) == kDimensions, "Wrong number of dimensions");
     return storage_ptr_->host_data(offset_) + host_pos(args...);
   }
 
@@ -250,10 +225,10 @@ class Tensor {
     return storage_ptr_->host_data(offset_);
   }
 
-  StorageIterator host_data() const {
+  /*StorageIterator host_data() const {
     // fromDevice();
     return storage_ptr_->host_data(offset_);
-  }
+  }*/
 
 // TODO: should we enable this again?
 #if 0
@@ -290,30 +265,34 @@ class Tensor {
 
   // TODO(Randl): variadic template version of reshape
   // TODO(Randl): non-checked version
-  void reshape(const std::array<size_t, kDimensions> &sz) {
+  /*void reshape(const std::array<size_t, kDimensions> &sz) {
     static_assert(!kConst, "Non-constant operation on constant Tensor");
     // No size change for reshape
     if (calcSize() != product(sz)) {
       throw nn_error("Reshape to Tensor of different size.");
     }
     shape_ = sz;
-  }
+  }*/
 
-  void resize(const std::array<size_t, kDimensions> &sz) {
+  void resize(const std::vector<size_t> &shape) {
     if (offset_ != 0 || size_ != storage_ptr_->size()) {
       throw nn_error("Resize of partial view is impossible.");
     }
-    shape_ = sz;
-    storage_ptr_->resize(std::vector<size_t>(sz.begin(), sz.end()));
+    shape_ = shape;
+    storage_ptr_->resize(std::vector<size_t>(shape.begin(), shape.end()));
   }
 
   size_t size() const { return size_; }
 
-  Tensor operator[](size_t index) {
+  size_t stride() const { return stride_; }
+
+  size_t dims() const { return shape_.size(); }
+
+  /*Tensor operator[](size_t index) {
     return Tensor(
       storage_ptr_, offset_ + index * size_ / shape_[0],
       std::array<size_t, kDimensions - 1>(shape_.begin() + 1, shape_.end()));
-  }
+  }*/
 
   /**
    * @brief Returns a sub view from the current tensor with a given size.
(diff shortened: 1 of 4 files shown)


Randl:
LGTM

edgarriba:
not sure how deal with view iterator

tiny_dnn/core/framework/tensor.h
```diff
@@ -392,21 +375,23 @@ class Tensor {
...
     // the new view.
-    // TODO(edgarriba/randl): add proper tests to this
     const size_t new_offset = offset_ + compute_offset(start, shape_);
     if (new_offset + product(new_shape) > size_) {
       throw nn_error("Cannot create a view from this tensor");
     }
 
-    return Tensor(storage_ptr_, new_offset, new_shape);
+    // computes the stride
+    const size_t new_stride = std::abs(shape_[dims() - 1] - new_shape[1]);
```
beru: This is a very trivial thing.Since definition of `Tensor::dims` method is `size_t dims() const { return shape_.size(); }`,
This line can also be written in this way.
```cpp
const size_t new_stride = std::abs(shape_.back() - new_shape[1]);
```

tiny_dnn/core/framework/tensor.h
```diff
@@ -415,12 +400,15 @@ class Tensor {
...
    */
-  std::array<size_t, kDimensions> shape_;
+  std::vector<size_t> shape_;
 
   /* Offset from the beginning of TensorStorage */
   size_t offset_;
   size_t size_;
 
+  /* Offset in case of non-contiguous memory */
+  size_t stride_;
```
beru: offset_ can be negative in case of bottom up data. so `ptrdiff_t` might be good to use.

beru:
Removing `kDimention` tensor dimention template parameter looks OK as it makes Tensor more flexible.
Usage of `stride` is  different from image processing programming so this needs to be debated.

tiny_dnn/core/framework/tensor.h
Randl: Agreed, more readable

edgarriba:
@beru not sure yet but the meaning of `stride` in Tensor might be different as in image processing. I don't understand the part of the branches. What do you mean with that?
http://nd4j.org/tensor

beru:
@edgarriba

For the branches, please take a look at [this line](https://github.com/tiny-dnn/tiny-dnn/pull/580/files#diff-54658142cc96675fb83229795a17df30R208).
> // apply stride in case that we are in the last two dimensions
> const size_t total_shift = (dim > 2) ? shift : shift + stride_;

In [here](https://github.com/tiny-dnn/tiny-dnn/pull/580/files#diff-54658142cc96675fb83229795a17df30R409), it says

>  /* Offset in case of non-contiguous memory */
>  size_t stride_;

So, I recognize in this PR, `stride` is used as an additional offset

I checked the page you introduced. -> http://nd4j.org/tensor
> It is the number of elements you skip in the buffer to get to the next channel or row or column. There’s a stride for each dimension.

Also checked several websites.
https://petewarden.com/2017/01/03/why-deep-learning-needs-assembler-hackers/
From looking the example code in below article, it looks to me `stride` variables are used to refer the number of elements to next different dimension with same indexes in other dimensions.

http://torch5.sourceforge.net/manual/torch/index-6-1-4.html
> The optional LongStorage strides gives the jump necessary to go from one element to the next one in the each dimension.

beru:
I still don't clearly understand how `stride` will be used in upcoming `Tensor` class.
But I wonder why it is only used for the last 2 dimensions.

My understanding is, regardless of whether tensor is continuous or not, we can use strides (pitches or offsets...you migh say) for the addressing elements. If a tensor is continuous, the stride will be equal to the size of dimension.

tiny_dnn/core/framework/tensor_utils.h
```diff
@@ -68,55 +68,39 @@ void print_pack(std::ostream &out, T t, U u, Args... args) {
   print_pack(out, u, args...);
```
Randl: Looks like now it calls itself recursively with same arguments

edgarriba:
@beru you are right, strides should be per dimension. In the way that @Randl designed accessors, I don't think that that stride has to equal to the dimension size. Learning on the fly how to manage everything.

As for use case, getting proper tensor views should speed up things and avoid unnecessary copies.

tiny_dnn/core/framework/tensor_utils.h
edgarriba: any idea how to fix it? still not confident with variadic arguments

Randl: @edgarriba I believe I've fixed it *and* got rid of most of part of variadic templates at the same time

Randl:
`C:\projects\tiny-dnn\tiny_dnn/core/framework/tensor.h(428): error C2668: 'abs': ambiguous call to overloaded function (compiling source file C:\projects\tiny-dnn\test\test.cpp) [C:\projects\build\test\tiny_dnn_test.vcxproj]`

What does that mean?

beru:
This code
```cpp
const size_t new_stride = std::abs(shape_.back() - new_shape[1]);
```

The type of `shape_.back()` is `size_t` and the type of `new_shape[1]` is also `size_t`.
So the type of the result is also `shape_.back() - new_shape[1]`.

But there aren't any `abs` overloaded function that argument is `size_t` so that must be the reason why it's `ambiguous`. 

http://en.cppreference.com/w/cpp/numeric/math/abs

edgarriba:
@beru thanks. I'll cast it

Randl:
After a discussion in a gitter group, I want to explain my vision of the view and subtensor topic, since I probably won't have an opportunity to implement it this week. 

I believe that the simplest way to implement arbitrary subtensor is just to store "parent" tensor dimensions, just calculating the position in an actual tensor. The problem might be in a iterator implementation. 

More complicated way is to store "distance" to the next element in each dimension (which I believe is called stride) and use it to calculate position of the next element. I personally prefer this method.

Randl:
Also note that although all test are passed, stride functionality is missing in new `host_at` functions

bhack:
@Randl See subarray section of https://github.com/kokkos/array_ref/blob/master/proposals/P0331.rst

beru:
@Randl Thank you for clearly explaining your ideas. Unfortunately, I can't understand the first one without seeing actual code... The second `strides` method seems very popular and I strongly recommend to follow this idiom.

## OpenCV
http://docs.opencv.org/trunk/d3/d63/classcv_1_1Mat.html#details
> The data layout of the array M is defined by the array M.step[], 
> ...
> So, the data layout in Mat is fully compatible with CvMat, IplImage, and CvMatND types from OpenCV 1.x. It is also compatible with the majority of dense array types from the standard toolkits and SDKs, such as Numpy (ndarray), Win32 (independent device bitmaps), and others, that is, with any array that uses steps (or **strides**) to compute the position of a pixel. Due to this compatibility, it is possible to make a Mat header for user-allocated data and process it in-place using OpenCV functions.

## NumPy
https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.strides.html
**numpy.ndarray.strides**
> Tuple of bytes to step in each dimension when traversing an array.

## Torch7
https://github.com/torch/torch7/blob/master/doc/tensor.md#number-stridedim
**[number] stride(dim)**
> Returns the jump necessary to go from one element to the next one in the specified dimension dim. Example:

## ND4J
http://nd4j.org/tensor#linear-buffer
> Stride tells the computation layer how to interpret the flattened representation. It is the number of elements you skip in the buffer to get to the next channel or row or column. There’s a stride for each dimension.

The stride of an array is pretty popular in computer vision (image processing) programming. 
https://en.wikipedia.org/wiki/Stride_of_an_array

It can be used for both of `continuous` and `non-continuous` tensors. We can just use the member variable in both cases.

beru:
@edgarriba 
>@beru the data structure models in image processing and in this case Tensors are a bit different. I think we discussed this somewhere some time ago. We should follow modern frameworks such torch or TF which support non contiguous memory tensors

Please explain about this specifically.

beru:
> Also note that although all test are passed, stride functionality is missing in new `host_at` functions
@Randl In that case, the test cases should be revised.

edgarriba:
@beru I will recheck all this resources and all possible tensor implementations in other frameworks to answer back to your questions

edgarriba:
@Randl @beru @pansk @bhack

https://github.com/QuantStack/xtensor

we should start a discussion about introducing xtensor as a backend for our tensor class. Clean API and covers almost all of our needs. No GPU but it shouldn't be difficult to wrap the raw data. As for CPU maybe @beru could give a deep insight to see how might affect to our current performance

Randl:
I'll look into it, sounds like a good idea.

beru:
@edgarriba @Randl I'm not against using xtensor but if we're going to use third-party library everywhere, I guess we need to place the library files inside tiny_dnn folder.

edgarriba:
@beru right, but I think that we can do an exception for this since it will be part of the library core. Personally I think that it can a lot of efforts and probably we should fork the library itself to our org account. We need to evaluate may things like state of the code, upstream workflow, performance, etc.

Try to take a look at those things for a further evaluation.

bhack:
There is no problem by the license cause it has the same tiny-dnn license. The question is generally how to [maintain upstream](http://stackoverflow.com/questions/31769820/differences-between-git-submodule-and-subtree).

beru:
By the way, when classifiers are used for object detection, people use sliding window / grid algorithm.
And I thought copying portions of a image for samples would be inefficient. So I guess we need iterator classes to ready tensor objects during training/inference.

edgarriba:
That's the whole point of having a good implementation for views

edgarriba:
https://github.com/QuantStack/xtensor/pull/136?

beru:
@edgarriba @bhack In terms of license, xtensor librarys uses BSD 3-Clause license, and I guess it's pretty much similar to the MIT lincense this library uses, so I suppose it's OK to incorporate it.

If I may speak about my opinion, I have a strong Not invented here syndrome, so I prefer to finish in-house one.

bhack:
This library is under BSD

beru:
@bhack Oh you're right. This library also uses The BSD 3-Clause License.
https://github.com/tiny-dnn/tiny-dnn/blob/master/LICENSE

Randl:
I'm for including third party tensor lib, as far as it's easy to maintain and easy to use. Unless we can provide simpler interface or significantly better performance, I don't see a point to reinvent the bicycle.

bhack:
@SylvainCorlay is subscribed to this topic If we want to discuss something specific about xtensors and tiny-dnn.

JohanMabille:
@edgarriba @beru @bhack We are currently refactoring parts of xtensor so that it is easy to plug other libraries in it and benefit from its expression system without the need to copy data from already existing structures. The documentation will be added a section which explains how to do that.

edgarriba:
@JohanMabille thanks for your comments. Is there any release date for this refactoring?
And BTW, how often do you release versions? Just to figure out what could be the best approach for integration: submodule, fork under org account, etc.

SylvainCorlay:
At the moment, we have a package for the conda package manager (linux, windows, OS X). Some people are looking at making a debian package too.

Are you guys be interested in using a package manager rather than vending a copy of your dependencies.

JohanMabille:
This refactoring should be released today, and the tutorial should land in the next couple of days.
Currently we release a new version every two or three weeks, or so.

About integration, we provide a install target in our cmake configuration so any project depending on xtensor can find it via the ``find_package()`` command. The ``xtensor`` conda package is built with cmake and includes the ``xtensorConfig.cmake`` file, so ``find_package(xtensor)`` works too (and I think it is the way debian packagers will go).

JohanMabille:
We just release xtensor 0.7.3 with further improvements on this. You can find [here](http://xtensor.readthedocs.io/en/latest/external-structures.html) a tutorial on how to plug external data structures into xtensor.

From my point of view, adapting one-dimensional containers with a strided index scheme is quite easy, but adapting a multi-dimensional container still requires too much code from authors of third-party libraries.

Let me know what you think. And do not hesitate to point to structures you may want to plug into xtensor.

edgarriba:
@JohanMabille thx to point this out! We'll look into it

beru:
@edgarriba We need to update `Tensor` class to use xarray's strides argument.

edgarriba:
@beru I guess we don't need to implement this anymore. xtensor view routines takes care of it

beru:
@edgarriba I can't grasp the idea... that assumption can be true if kernel implmentations would internally use xtensor view routines. But I don't think it would always be the case.

And .... the `Tensor` class internally holds `xarray` instance as its storage. I wonder how users let the library use the buffer that comes from outside of the library...

There can be many requests and corner-cases for this Tensor thing which makes it difficult to implement.

SylvainCorlay:
@beru xtensor strided containers provide a `raw_data` method that returns a pointer to the data buffer.

Views and strided views on expressions that expose the raw-data API also expose the `raw_data` API.

```cpp
template <class T = xexpression_type>
std::enable_if_t<has_raw_data_interface<T>::value, const value_type*>
raw_data() const;
```

beru:
I wonder how we can control line stride without strides parameter in Tensor class constructor... it seems automatic and cannot change it.

@SylvainCorlay Thank you for the info.

`xarray_container` extends `xstrided_container` and it has `raw_data` method so we can retrieve an address of tensor storage for crude operations.

PR #581, opened 2017-02-16
Fixes Cifar-10 readme.md dimensionality.

pliptor:
The input size for this problem are 32x32 pixels frames of depth 3 (RGB).
Fixes the calculation of dimensions in this problem from 1024 to 32x32x3=3072.
The formula 32x32x3 is correct but the result is not.

Thanks.


edgarriba:
LGTM

Issue #491, opened 2016-12-31
Example programs should be built without external libraries


beru:
I think example programs should be built without external libraries.
Some of them used to work without `cereal` library so it's possible to do so.
(`example_mnist_train`, `example_mnist_test`, `example_mnist_quantized_train`, `example_deconv_train`)

Talking about `examples/cifar10/test.cpp`, it isn't listed in `examples/CMakeLists.txt` and it depends on OpenCV library so I'd like to update it.

And I don't see `example_deconv_test` and `example_mnist_quantized_test`...

PR #582, opened 2017-02-16
Cifar-10 example. Removes OpenCV req. Add entry in CMakeLists…(#491)

pliptor:
This patch addresses #491 

* It removes the requirement of OpenCV to build the test example for cifar10.
* It adds an entry in CMakesList.txt so that the executable is actually built (it wasn't before, at least on ubuntu).

OpenCV is used in the version prior this patch to convert images from the user input to tiny-dnn vector. The same functionality is present in MNIST test, which uses tiny-dnn native image converter. Therefore OpenCV was replaced by the native version. 

In addition:

* The input range is properly normalized. The previous version was discarding the maxv value. 
* Minor refactoring to make it consistent with MNIST test variables.

The resulting binary was tested and functional with arbitrary images given to it as input. <del> Needs additional tests after RGB deinterleaving.</del>

The attached image shows the output in the first layer when an arbitrary image of a horse from the web is given to the input. No typical signs of interlacing such as aliasing or other issues related to RGB deinterleaving is present.

layer_0

Thanks.

### DIFF
--- a/examples/cifar10/test.cpp
+++ b/examples/cifar10/test.cpp
@@ -20,12 +20,12 @@ void convert_image(const std::string &imagefilename,
                    vec_t &data) {
   image<> img(imagefilename, image_type::rgb);
   image<> resized = resize_image(img, w, h);
-  data.resize(w * h * 3);
-  for (int c = 0; c < 3; ++c) {
-    for (int y = 0; y < w; ++y) {
-      for (int x = 0; x < h; ++x) {
-        data[c * w * h + y * w + x] =
-          (maxv - minv) * (resized[y * w + x + c]) / 255.0 + minv;
+  data.resize(resized.width() * resized.height() * resized.depth());
+  for (size_t c = 0; c < resized.depth(); ++c) {
+    for (size_t y = 0; y < resized.height(); ++y) {
+      for (size_t x = 0; x < resized.width(); ++x) {
+        data[c * resized.width() * resized.height() + y * resized.width() + x] =
+          (maxv - minv) * (resized[y * resized.width() + x + c]) / 255.0 + minv;
       }
     }
   }
--- a/examples/cifar10/test.cpp
+++ b/examples/cifar10/test.cpp
@@ -24,8 +24,8 @@ void convert_image(const std::string &imagefilename,
   for (int c = 0; c < 3; ++c) {
     for (int y = 0; y < w; ++y) {
       for (int x = 0; x < h; ++x) {
-        data[c * w * h + y * w + x] +=
-          (maxv - minv) * (data[y * w + x + c]) / 255.0 + minv;
+        data[c * w * h + y * w + x] =
+          (maxv - minv) * (resized[y * w + x + c]) / 255.0 + minv;
       }
     }
   }
--- a/examples/cifar10/test.cpp
+++ b/examples/cifar10/test.cpp
@@ -20,8 +20,15 @@ void convert_image(const std::string &imagefilename,
                    vec_t &data) {
   image<> img(imagefilename, image_type::rgb);
   image<> resized = resize_image(img, w, h);
-  std::transform(resized.begin(), resized.end(), std::back_inserter(data),
-                 [=](uint8_t c) { return c * (maxv - minv) / 255.0 + minv; });
+  data.resize(w * h * 3);
+  for (int c = 0; c < 3; ++c) {
+    for (int y = 0; y < w; ++y) {
+      for (int x = 0; x < h; ++x) {
+        data[c * w * h + y * w + x] +=
+          (maxv - minv) * (data[y * w + x + c]) / 255.0 + minv;
+      }
+    }
+  }
 }
 
 template <typename N>
--- a/examples/cifar10/test.cpp
+++ b/examples/cifar10/test.cpp
@@ -1,8 +1,4 @@
 #include <iostream>
-#include <opencv2/opencv.hpp>
-/*#include <opencv2/imgproc.hpp>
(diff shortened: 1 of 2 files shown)


MasazI:
As for the test of Cifar10, since the channel is fixed to 3, I think that it is better to be dynamic.

pliptor:
@MasazI Yes. I agree.

pliptor:
@MasazI 
I reviewed the tiny-dnn code that wraps stb and the stb library. It seems the code as it is now (including the patches) handles it all automatically. If you mean to make the hard-coded value of 3 in line 24 to be dynamic, it is guaranteed it will be always 3 channels because the stb library is being requested to convert the input to RGB via image_type::rgb. 

Does it make sense?

The stb interface is the details.
<details>

> 

// Basic usage (see HDR discussion below for HDR usage):
//    int x,y,n;
//    unsigned char *data = stbi_load(filename, &x, &y, &n, 0);
//    // ... process data if not NULL ...
//    // ... x = width, y = height, n = # 8-bit components per pixel ...
//    // ... replace '0' with '1'..'4' to force that many components per pixel
//    // ... but 'n' will always be the number that it would have been if you said 0
//    stbi_image_free(data)
//
// Standard parameters:
//    int *x       -- outputs image width in pixels
//    int *y       -- outputs image height in pixels
//    int *comp    -- outputs # of image components in image file
//    int req_comp -- if non-zero, # of image components requested in result
//
// The return value from an image loader is an 'unsigned char *' which points
// to the pixel data, or NULL on an allocation failure or if the image is
// corrupt or invalid. The pixel data consists of *y scanlines of *x pixels,
// with each pixel consisting of N interleaved 8-bit components; the first
// pixel pointed to is top-left-most in the image. There is no padding between
// image scanlines or between pixels, regardless of format. The number of
// components N is 'req_comp' if req_comp is non-zero, or *comp otherwise.
// If req_comp is non-zero, *comp has the number of components that _would_
// have been output otherwise. E.g. if you set req_comp to 4, you will always
// get RGBA output, but you can check *comp to see if it's trivially opaque
// because e.g. there were only 3 channels in the source image.
//
// An output image with N components has the following components interleaved
// in this order in each pixel:
//
//     N=#comp     components
//       1           grey
//       2           grey, alpha
//       3           red, green, blue
//       4           red, green, blue, alpha
`
</details>

examples/cifar10/test.cpp
```diff
@@ -22,16 +18,14 @@ void convert_image(const std::string &imagefilename,
...
-  if (img.data == nullptr) return;  // cannot open, or it's not an image
-  cv::Mat resized;
-  cv::resize(img, resized, cv::Size(w, h), .0, .0);
-  data.resize(w * h * resized.channels(), minv);
-  for (int c = 0; c < resized.channels(); ++c) {
-    for (int y = 0; y < resized.rows; ++y) {
-      for (int x = 0; x < resized.cols; ++x) {
+  image<> img(imagefilename, image_type::rgb);
+  image<> resized = resize_image(img, w, h);
+  data.resize(w * h * 3);
```
edgarriba: use this
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/image.h#L271

examples/cifar10/test.cpp
```diff
@@ -22,16 +18,14 @@ void convert_image(const std::string &imagefilename,
...
+  for (int c = 0; c < 3; ++c) {
+    for (int y = 0; y < w; ++y) {
+      for (int x = 0; x < h; ++x) {
         data[c * w * h + y * w + x] =
-          resized.data[y * resized.step[0] + x * resized.step[1] + c];
+          (maxv - minv) * (resized[y * w + x + c]) / 255.0 + minv;
```
edgarriba: make sure that this indexing work as expected.

Notice that we are introducing a Tensor class which have proper accessors and you might want to you use it here. Probably Image API can be refactored with Tensor API too.
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/framework/tensor.h

examples/cifar10/test.cpp
pliptor: Yeah, that 3 is ugly but I was't sure what's the best approach to handle it. 
1. Define a macro? #define N_CHANNELS 3 // number of channels
2. Define int depth = 3; and plug 3 in there? // number of channels
I did look at the line you referenced earlier but it was a private variable. We could define a get_channel() method perhaps?

edgarriba: right, add whatever change you need. Better, `channel()` for getters

examples/cifar10/test.cpp
pliptor: I inspected the images that get created inside the layers and they looked fine but I'll double check. The original OpenCV code was not scaling the range and that has been added above as you see.  Okay. Let me take at the Tensor API. That's new for me.

edgarriba: We are still testing it. So just be aware of it by now.

examples/cifar10/test.cpp
pliptor: Alright. I'll do it tomorrow. I was trying to stay away from the library code itself but here we go.

Randl: Maybe use `image_type::rgb` to determine number of channels?

edgarriba:
@Randl `image_type::rgb` it's just an enum

pliptor:
In a closer inspection, the image.h library does provide the depth() method so we won't need to touch the library itself. I replaced the w and h values by the width() and height() methods as well since they are also provided. Thanks.

PR #585, opened 2017-02-18
Caffe converter: support ELU activation, skip HDF5Data layers

festlv:
This PR adds ELU activation support and skips HDF5Data layers when importing caffe model.
It also fixes a build failure (using image functions required DNN_USE_IMAGE_API).


edgarriba:
LGTM

Issue #588, opened 2017-02-19
Question: Does anyone have a cifar10 weight pre-computed for sharing?


pliptor:
My computer seems way too slow to train it with the default setup.

Thank you!

Randl:
I can run one for you. Do you want default parameters?

pliptor:
@Randl Sweet. Yes, it would be useful for the newcomers if the run into the same problem. How long does it take to train in your box?

Randl:
@pliptor I've put it right now with all defaults and `time`. One epoch takes ~140 secs, with TBB and AVX (note it's 4 socket server).

pliptor:
That's roughly that it takes for me to train one MNIST epoch :( Thanks again.

Randl:
Here are the weights: [cifar-weights.zip](https://github.com/tiny-dnn/tiny-dnn/files/786674/cifar-weights.zip). I'll send PR adding them and couple of other changes later today

pliptor:
Thank you! Now I can fully test the test code too.

PR #590, opened 2017-02-20
 Add pre-trained weights to CIFAR-10

Randl:
Since CIFAR-10 example can take pretty much time to train, putting pre-traing weights into repository might help newcomers (see https://github.com/tiny-dnn/tiny-dnn/issues/588 ). Also I added epoch number as a run-time parameter.


pliptor:
That's a nice parameter parser you added. Here's a couple of minor suggestions.

[cifar.patch.txt](https://github.com/tiny-dnn/tiny-dnn/files/788405/cifar.patch.txt)

beru:
Personally, I prefer to use nonnamed arguments. But not a big issue.
And I wish to start training with pre-trained weights...

Randl:
@bedu done

pliptor:
I think this got dropped too. --data_path is getting glued to the command name.
[a.patch.txt](https://github.com/tiny-dnn/tiny-dnn/files/792502/a.patch.txt)

examples/cifar10/readme.md
```diff
@@ -52,7 +53,8 @@ parse_cifar10(data_dir_path + "/test_batch.bin",
 
 # Grid Search
 One of the most important hyperparameter in deep learning is learning rate. 
-To get stable and better result, let's try [grid search](https://en.wikipedia.org/wiki/Hyperparameter_optimization#Grid_search)
+To get stable and better result, let's try
+ [grid search](https://en.wikipedia.org/wiki/Hyperparameter_optimization#Grid_search)
```
edgarriba: @Randl grid search is not a good idea for DL. Use random search instead. We should mention this paper here
http://www.jmlr.org/papers/volume13/bergstra12a/bergstra12a.pdf

Randl: It's a bit out of scope of PR, but I can try to fix it

pliptor: @Randl Which hyper-parameters were you planning to search for?

Randl: That was there before

edgarriba: @Randl oh, I thought that you inserted this in this PR. It's just the formatter. Up to you, in this or next PR is okay.

examples/cifar10/train.cpp
```diff
@@ -104,21 +119,58 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
-         << " learning rate (example:0.01)" << endl;
+  double learning_rate         = 0.1;
+  int epochs                   = 30;
+  std::string data_path        = "";
+  int minibatch_size           = 10;
+  core::backend_t backend_type = core::default_engine();
+  for (int count = 1; count + 1 < argc; count += 2) {
+    std::string argname(argv[count]);
+    if (argname == "--learning_rate")
+      learning_rate = atof(argv[count + 1]);
```
beru: I wonder if invalid value should be checked .... `learning_rate <= 0`

examples/cifar10/train.cpp
```diff
@@ -104,21 +119,58 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
+    else if (argname == "--epochs")
+      epochs = atoi(argv[count + 1]);
```
beru: The number of epochs must be greater than 0.

examples/cifar10/train.cpp
```diff
@@ -104,21 +119,58 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
+    else if (argname == "--minibatch_size")
+      minibatch_size = atoi(argv[count + 1]);
```
beru: `minibatch_size` must be greater than 0 and should not be greater than the number of CIFAR-10 training images (50000).

examples/cifar10/train.cpp
```diff
@@ -104,21 +119,58 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
+              << argv[0]
+              << " --data_path ../data --learning_rate 0.01 --epochs 30 "
+              << "--minibatch_size 10 --backend_type internal" << std::endl;
     return -1;
   }
-  train_cifar10(argv[1], stod(argv[2]), cout);
-}
+  std::cout << "Running with learning rate " << learning_rate
+            << " and minibatch of size " << minibatch_size << " for " << epochs
+            << " epochs." << std::endl;
```
beru: I personally prefer hyperparameters to be printed on separate lines so that they can easily parsed by programs.

examples/cifar10/train.cpp
```diff
@@ -77,21 +90,23 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
                 -1.0, 1.0, 0, 0);
 
-  cout << "start learning" << endl;
+  std::cout << "start learning" << std::endl;
 
   progress_display disp(train_images.size());
   timer t;
-  const int n_minibatch    = 10;  ///< minibatch size
-  const int n_train_epochs = 30;  ///< training duration
+  // const int n_minibatch = 10;  ///< minibatch size
```
beru: I guess this comment line isn't needed.

examples/cifar10/train.cpp
```diff
@@ -45,29 +49,38 @@ void construct_net(N &nn) {
...
+     << pool(32, 32, n_fmaps, 2, backend_type)
+     << conv(16, 16, 5, n_fmaps, n_fmaps, padding::same, true, 1, 1,
+             backend_type)
+     << pool(16, 16, n_fmaps, 2, backend_type)
+     << conv(8, 8, 5, n_fmaps, n_fmaps2, padding::same, true, 1, 1,
+             backend_type)
+     << pool(8, 8, n_fmaps2, 2, backend_type)
+     << fully_connected_layer<activation::identity>(4 * 4 * n_fmaps2, n_fc,
+                                                    true, backend_type)
+     << fully_connected_layer<softmax>(n_fc, 10, true, backend_type);
```
beru: Thank you for the hard work!

examples/cifar10/readme.md
```diff
@@ -140,33 +142,49 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream& log) {
...
-Usually it implies too large learning rate.
+Usually it implies too large learning rate:
 
 ```
 [Warning]Detected infinite value in weight. stop learning.
 ```
 
-You will get about 70% accuracy in learning rate=0.01.
+You will get about 70% accuracy in learning rate=0.01. 
+There are pre-trained weights in file `cifar-weights` in this folder.
```
beru: There's a pre-trained weights file named `cifar-weights` in this folder.

examples/cifar10/train.cpp
```diff
@@ -26,16 +26,20 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
+#include <cstdlib>
 #include <iostream>
 #include <vector>
 #include "tiny_dnn/tiny_dnn.h"
 
+#ifdef _MSC_VER
+#define strcasecmp _stricmp
+#endif
+
 using namespace tiny_dnn;
```
bhack: also WIN_32?

Randl: No, thats MSVC specific, http://stackoverflow.com/questions/3694723/error-c3861-strcasecmp-identifier-not-found-in-visual-studio-2008

examples/cifar10/train.cpp
```diff
@@ -26,16 +26,20 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
 using namespace tiny_dnn::activation;
-using namespace std;
 
 template <typename N>
-void construct_net(N &nn) {
+void construct_net(N &nn, core::backend_t backend_type) {
```
edgarriba: the template here means that it also can handle graph nets? don't think so

examples/cifar10/train.cpp
```diff
@@ -104,21 +116,75 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
-
+  nn.test(test_images, test_labels).print_detail(std::cout);
   // save networks
-  ofstream ofs("cifar-weights");
+  std::ofstream ofs("cifar-weights");
   ofs << nn;
 }
 
+static core::backend_t parse_backend_name(const char *name) {
+  static const char *names[] = {
```
edgarriba: why cannot use c++ vector here instead pointer, sizeof etc? just to be coherent with the library itself

examples/cifar10/train.cpp
```diff
@@ -104,21 +116,75 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
-    cerr << "Usage : " << argv[0] << " path_to_data (example:../data)"
-         << " learning rate (example:0.01)" << endl;
+  double learning_rate         = 0.1;
+  int epochs                   = 30;
+  std::string data_path        = "";
+  int minibatch_size           = 10;
+  core::backend_t backend_type = core::default_engine();
+  for (int count = 1; count + 1 < argc; count += 2) {
+    std::string argname(argv[count]);
+    if (argname == "--learning_rate")
```
edgarriba: for readability I would put curly brackets between statements

examples/cifar10/train.cpp
```diff
@@ -168,9 +165,26 @@ int main(int argc, char **argv) {
...
+  if (minibatch_size <= 0 || minibatch_size > 50000) {
+    std::cerr << "Invalid minibatch size. Minibatch rate must be greater than 0"
+                 " and less than dataset size (50000)"
+              << std::endl;
+    return -1;
+  }
+  std::cout << "Running with following parameters:" << std::endl
+            << "Learning rate: " << learning_rate << std::endl
+            << "Minibatch size: " << minibatch_size << std::endl
+            << "Epochs: " << epochs << std::endl;
```
beru: Please also print out `backend_type`. If it is printed in a log file, we can know which backend was used during training.

examples/cifar10/train.cpp
Randl: I've taken the piece of code from @beru . Using C++ vectors with C strings seems weird for me.

examples/cifar10/train.cpp
Randl: It probably can't, but I'm not sure. Once again, it was here before, I believe @nyanp is the author of it.

examples/cifar10/train.cpp
beru: @edgarriba I would use `std::array` from that perspective.
http://en.cppreference.com/w/cpp/container/array

But I wouldn't use `std::vector` and `std::string` in here. They aren't fundamentally needed in this routine. `auto` specifier keyword can also be used instead of explicit type declaratios.

if you really want your programs not-succint and resource hogging, you may carelessly persist to use STL containers. I wouldn't be able stop you in that case.

edgarriba: @Randl  and what about this? http://cpp.sh/873dp
```c++
static core::backend_t parse_backend_name(const std::string& name) {
    static const std::vector<std::string> names = {
        "internal", "nnpack", "libdnn", "avx", "opencl"
    };
      
  for (size_t i = 0; i < names.size(); ++i) {
    if (name.compare(names[i]) == 0) {
        return static_cast<core::backend_t>(i);
    }
  }
  return core::backend_t::internal;
}

```

edgarriba: @beru sure, we can use array too. Saw that after posting.

edgarriba: Also note that we are not optimizing performance here, this is just an example for standard users which probably will use plane c++11 and since the library itself it's targeted in that, I think it's a good idea to have high level interfaces without mixing low level C.

beru: @edgarriba `std::basic_string::cpmpare` is case sensitive and `strcasecmp` function ignores case so that I used that function.

You don't need to (actually, you shouldn't) use `std::string` just to wrap string literals.
I hope something similar to `boost::string_ref` would appear in future C++ standard though.

beru: @edgarriba The reason I don't want to use `std::vector` and `std::string` classes is they allocate and copy memory for their instances. In this function, C string literal array is perfectly fine for the task. What is needed is nothing more or less.

edgarriba: ok, but as said this is just an example. Nothing important to optimize here. We could directly solve this by using integer values. So as you prefer guys, just gave my opinion.

beru: @edgarriba If you don't like this idiom `sizeof(arr) / sizeof(arr[0])` to get array length, adding a nifty `countof` function is another way.
http://www.g-truc.net/post-0708.html
http://brnz.org/hbr/?p=1491

beru: @edgarriba We don't need to use flamethrowers to cook fried chicken.

edgarriba: this last looks better. However, we should provide simple examples with simple code and use "modern" interfaces as much as possible in order to let new users be confident with the library as fast as possible. That was my point.

@Randl up to you to update the code after seeing this discussion. Ping once you are done.

examples/cifar10/train.cpp
```diff
@@ -104,21 +116,77 @@ void train_cifar10(string data_dir_path, double learning_rate, ostream &log) {
...
+  const std::array<const std::string, 5> names = {
+    "internal", "nnpack", "libdnn", "avx", "opencl",
+  };
+  for (size_t i = 0; i < names.size(); ++i) {
+    if (name.compare(names[i]) == 0) {
+      return static_cast<core::backend_t>(i);
+    }
+  }
+  return core::default_engine();
+}
```
Randl: @edgarriba @beru I went with C++-style code at the end.

pliptor:
@Randl  A lot of work from a request for pre-trained weights. Thank you very much.

examples/cifar10/train.cpp
```diff
@@ -126,11 +126,11 @@ void train_cifar10(std::string data_dir_path,
...
 static core::backend_t parse_backend_name(const char *name) {
-  static const char *names[] = {
+  std::array<std::string, 5> names = {
     "internal", "nnpack", "libdnn", "avx", "opencl",
   };
-  for (size_t i = 0; i < sizeof(names) / sizeof(names[0]); ++i) {
-    if (strcasecmp(name, names[i]) == 0) {
+  for (size_t i = 0; i < names.size(); ++i) {
+    if (strcasecmp(name, names[i].c_str()) == 0) {
       return (core::backend_t)i;
```
edgarriba: safe type check here

edgarriba:
@Randl code LGTM. One last thing: could you please write a small comment helper with script usage or ideally a short README to be placed into the official docs?

edgarriba:
@Randl revise the whole README, the codes showed there are not updated

examples/cifar10/train.cpp
```diff
@@ -156,13 +156,14 @@ int main(int argc, char **argv) {
...
-          << "--minibatch_size 10 --backend_type internal" << std::endl;
+      std::cout << "Example of usage :\n"
+                << argv[0]
+                << " --data_path ../data --learning_rate 0.01 --epochs 30 "
+                << "--minibatch_size 10 --backend_type internal" << std::endl;
       return 0;
     } else {
       std::cerr << "argument " << argname << " isn't supported. Use --help to "
-          << "get usage example";
+                << "get usage example";
```
beru: If invalid command line argument is specified, printing out the error message is good but how about ending the program with `return -1;` after the err msg?

Randl:
@edgarriba @beru Can we merge?

beru:
@Randl Yes, we can. I've approved the changes. Other reviewers should update the status.

PR #600, opened 2017-03-01
Drop activation template from conv and deconv layers.

karandesai-96:
The `Activation` template parameter has been completely decoupled from the following layers:

1. `convolutional_layer`
2. `deconvolutional_layer`
3. ~`quantized_convolutional_layer`~ (next PR)
4. ~`quantized_deconvolutional_layer`~ (next PR)
5. ~`quantized_fully_connected_layer`~ (next PR)

Changes were reflected in the class definition, examples, tests, kernels and serialization helpers, if any.

### DIFF
--- a/examples/cifar10/readme.md
+++ b/examples/cifar10/readme.md
@@ -18,22 +18,27 @@ This means network architecture for Cifar-10 tends to be larger (or/and deeper) 
 // specify loss-function and learning strategy
 network<cross_entropy, adam> nn;
 
-typedef convolutional_layer<activation::identity> conv;
-typedef max_pooling_layer<relu> pool;
+using conv    = convolutional_layer;
+using pool    = max_pooling_layer<identity>;
+using fc      = fully_connected_layer;
+using relu    = relu_layer;
+using softmax = softmax_layer;
 
-const int n_fmaps = 32; ///< number of feature maps for upper layer
-const int n_fmaps2 = 64; ///< number of feature maps for lower layer
-const int n_fc = 64; ///< number of hidden units in fully-connected layer
+const int n_fmaps  = 32;  ///< number of feature maps for upper layer
+const int n_fmaps2 = 64;  ///< number of feature maps for lower layer
+const int n_fc     = 64;  ///< number of hidden units in fully-connected layer
 
-nn << conv(32, 32, 5, 3, n_fmaps, padding::same)
-    << pool(32, 32, n_fmaps, 2)
-    << conv(16, 16, 5, n_fmaps, n_fmaps, padding::same)
-    << pool(16, 16, n_fmaps, 2)
-    << conv(8, 8, 5, n_fmaps, n_fmaps2, padding::same)
-    << pool(8, 8, n_fmaps2, 2)
-    << fully_connected_layer(4 * 4 * n_fmaps2, n_fc)
-    << fully_connected_layer(n_fc, 10)
-    << softmax_layer(10);
+nn << conv(32, 32, 5, 3, n_fmaps, padding::same)          // C1
+   << pool(32, 32, n_fmaps, 2)                            // P2
+   << relu(16, 16, n_fmaps)                               // activation
+   << conv(16, 16, 5, n_fmaps, n_fmaps, padding::same)    // C3
+   << pool(16, 16, n_fmaps, 2)                            // P4
+   << relu(8, 8, n_fmaps)                                 // activation
+   << conv(8, 8, 5, n_fmaps, n_fmaps2, padding::same)     // C5
+   << pool(8, 8, n_fmaps2, 2)                             // P6
+   << relu(4, 4, n_fmaps2)                                // activation
+   << fc(4 * 4 * n_fmaps2, n_fc)                          // FC7
+   << fc(n_fc, 10) << softmax_layer(10);                  // FC10
 
 ```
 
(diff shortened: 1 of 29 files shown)


tiny_dnn/layers/deconvolutional_layer.h
```diff
@@ -158,66 +156,21 @@ class deconvolutional_layer : public feedforward_layer<Activation> {
...
     deconv_set_params(shape3d(in_width, in_height, in_channels), window_size,
                       window_size, out_channels, pad_type, has_bias, w_stride,
                       h_stride, connection_table);
     init_backend(backend_type);
   }
 
-  /**
-  * constructing deconvolutional layer
-  *
-  * @param in_width         [in] input image width
```
edgarriba: Why do you remove this?

test/test_convolutional_layer.h
```diff
@@ -154,15 +154,15 @@ TEST(convolutional, fprop) {
...
-    EXPECT_NEAR(0.6899745, out[8], 1E-5);
+    EXPECT_NEAR(float_t(-0.05), out[0], 1E-5);
+    EXPECT_NEAR(float_t(1.65), out[1], 1E-5);
+    EXPECT_NEAR(float_t(1.45), out[2], 1E-5);
+    EXPECT_NEAR(float_t(1.05), out[3], 1E-5);
+    EXPECT_NEAR(float_t(0.00), out[4], 1E-5);
+    EXPECT_NEAR(float_t(-2.0), out[5], 1E-5);
+    EXPECT_NEAR(float_t(0.40), out[6], 1E-5);
+    EXPECT_NEAR(float_t(1.15), out[7], 1E-5);
+    EXPECT_NEAR(float_t(0.80), out[8], 1E-5);
```
karandesai-96: It can do well without this `float_t` right ?

examples/cifar10/readme.md
```diff
@@ -65,22 +65,28 @@ using namespace tiny_dnn::activation;
...
+     << relu(16, 16, n_fmaps)                             // activation
+     << conv(16, 16, 5, n_fmaps, n_fmaps, padding::same)  // C3
+     << pool(16, 16, n_fmaps, 2)                          // P4
+     << relu(8, 8, n_fmaps)                               // activation
+     << conv(8, 8, 5, n_fmaps, n_fmaps2, padding::same)   // C5
+     << pool(8, 8, n_fmaps2, 2)                           // P6
+     << relu(4, 4, n_fmaps)                               // activation
+     << fc(4 * 4 * n_fmaps2, n_fc)                        // FC7
+     << relu(n_fc)                                        // activation
+     << fc(n_fc, 10) << softmax(10);                      // softmax
```
Randl: FC10

tiny_dnn/util/serialization_layer_list.h
```diff
@@ -1,14 +1,14 @@
 // Copyright (c) 2017, Taiga Nomi
 #ifndef CNN_NO_SERIALIZATION
 
-CNN_REGISTER_LAYER_WITH_ACTIVATIONS(convolutional_layer, conv);
 CNN_REGISTER_LAYER_WITH_ACTIVATIONS(average_pooling_layer, avepool);
 CNN_REGISTER_LAYER_WITH_ACTIVATIONS(max_pooling_layer, maxpool);
 CNN_REGISTER_LAYER_WITH_ACTIVATIONS(linear_layer, linear);
```
edgarriba: Do we need this one anymore?

tiny_dnn/layers/deconvolutional_layer.h
karandesai-96: @edgarriba: The error stated that this constructor definition doesn't match with anyone from `layer`. I thought it was derived from `feedforward_layer`

examples/deconv/train.cpp
```diff
@@ -37,17 +37,25 @@ void deconv_lanet(network<graph> &nn,
...
   // declare nodes
   input_layer i1(shape3d(32, 32, 1));
-  convolutional_layer<tan_h> c1(32, 32, 5, 1, 6);
+  convolutional_layer c1(32, 32, 5, 1, 6);
+  tanh_layer c1th(28, 28, 6);
   average_pooling_layer<tan_h> p1(28, 28, 6, 2);
-  deconvolutional_layer<tan_h> d1(14, 14, 5, 6, 16,
-                                  connection_table(tbl, 6, 16));
+  deconvolutional_layer d1(14, 14, 5, 6, 16, connection_table(tbl, 6, 16));
+  tanh_layer d1th(18, 18, 16);
```
Randl: maybe `d1_tanh`?

test/test_convolutional_layer.h
```diff
@@ -13,11 +13,11 @@
 namespace tiny_dnn {
 
 TEST(convolutional, setup_internal) {
-  convolutional_layer<sigmoid> l(5, 5, 3, 1, 2);
+  convolutional_layer l(5, 5, 3, 1, 2);
 
   EXPECT_EQ(l.parallelize(), true);              // if layer can be parallelized
   EXPECT_EQ(l.in_channels(), serial_size_t(3));  // num of input tensors
-  EXPECT_EQ(l.out_channels(), serial_size_t(2));    // num of output tensors
+  EXPECT_EQ(l.out_channels(), serial_size_t(1));    // num of output tensors
```
Randl: Why has that changed?

test/test_convolutional_layer.h
Randl: Better to have it. BTW, are those exact values?

test/test_quantized_convolutional_layer.h
```diff
@@ -97,97 +93,141 @@ TEST(quantized_convolutional, fprop) {
...
-        EXPECT_NEAR(0.8388910, out[1], 2E-2);
-        EXPECT_NEAR(0.8099984, out[2], 2E-2);
-        EXPECT_NEAR(0.7407749, out[3], 2E-2);
-        EXPECT_NEAR(0.5000000, out[4], 2E-2);
-        EXPECT_NEAR(0.1192029, out[5], 2E-2);
-        EXPECT_NEAR(0.5986877, out[6], 2E-2);
-        EXPECT_NEAR(0.7595109, out[7], 2E-2);
-        EXPECT_NEAR(0.6899745, out[8], 2E-2);
-    }
+  weight[0] = 0.3;
```
Randl: Please, change to compact formattion, and add guard comments ` // clang-format off `

test/test_convolutional_layer.h
karandesai-96: `out_channels` returns the number of output vectors of tensors. decoupled layers don't need `vector_type::aux` anymore, so it reduced.

Although I think `out_channels` name is misleading. @edgarriba @Randl thoughts ?

test/test_convolutional_layer.h
karandesai-96: @Randl yes. There was sigmoid earlier so the values were non recurring decimals, but since it is elementwise product of simple rational numbers, these values are rational.

test/test_quantized_convolutional_layer.h
karandesai-96: Yes, even I'll prefer that way. I'm changing it now.

tiny_dnn/util/serialization_layer_list.h
edgarriba: great

examples/cifar10/readme.md
edgarriba: I also propose to add an additional alias for fc -> `linear` in order to be aligned with other framewroks. @karandesai-96 maybe fro next PR

examples/deconv/train.cpp
edgarriba: also activation layers I think that can be renamed at some point and remove the `_layer` part

test/test_convolutional_layer.h
edgarriba: right, just `outputs()` could avoid confusions

examples/cifar10/readme.md
Randl: `fully_connected` is more intuitive Imhotep

tiny_dnn/layers/deconvolutional_layer.h
edgarriba: mmm, I don't think so. I'd say that this ctor supports non-squared kernels.

examples/cifar10/readme.md
edgarriba: fully_connected is a high level interface name, but the underneath operation it's just a linear matrix multiplication.

karandesai-96:
@edgarriba I've shifted code related quantized layers in a separate branch for maintaining smaller diffs, it was hard to inspect. Also I have rewritten history and kept only 2 commits, each containing all the changes required to drop template from one of the layers. To keep a proper track, I think one commit per layer is suitable, could you use "Rebase and Merge" for this case ? Thanks.

tiny_dnn/models/alexnet.h
```diff
@@ -19,14 +19,21 @@ namespace models {
...
-    *this << max_pool<identity>(54, 54, 64, 2);
-    *this << conv<relu>(27, 27, 5, 5, 64, 192, padding::valid, true, 1, 1);
-    *this << max_pool<identity>(23, 23, 192, 1);
-    *this << conv<relu>(23, 23, 3, 3, 192, 384, padding::valid, true, 1, 1);
-    *this << conv<relu>(21, 21, 3, 3, 384, 256, padding::valid, true, 1, 1);
-    *this << conv<relu>(19, 19, 3, 3, 256, 256, padding::valid, true, 1, 1);
-    *this << max_pool<identity>(17, 17, 256, 1);
+    // todo: (karandesai) shift this to tiny_dnn::activation
+    using relu = relu_layer;
+    *this << conv(224, 224, 11, 11, 3, 64, padding::valid, true, 4, 4);
```
Randl: Out of scope of PR, but can't we just do `*this << conv(224, 224, 11, 11, 3, 64, padding::valid, true, 4, 4)<< relu(54, 54, 64)` and so on?

karandesai-96: @Randl: I had the same question when I first saw this :smile:

edgarriba:
LGTM

edgarriba:
@karandesai-96 could you try to rebase from upstream master?

PR #601, opened 2017-03-02
Drop activation template from max_pooling_layer.

karandesai-96:
The `Activation` template parameter has been completely decoupled from `max_pooling_layer`. Changes are similar to those done in #600.


edgarriba:
LGTM

PR #604, opened 2017-03-03
Drop activation template from average pool, unpool layers.

karandesai-96:
The `Activation` template parameter has been completely decoupled from `average_pooling_layer` and `average_unpooling_layer`.

First of all the base class `partial_connected_layer` is made to inherit directly from layer, followed y refactor in pooling and unpooling layer classes.

### DIFF
--- a/examples/mnist/quantized.cpp
+++ b/examples/mnist/quantized.cpp
@@ -33,17 +33,18 @@ void construct_net(network<sequential> &nn) {
   nn << quantized_convolutional_layer<tan_h>(
           32, 32, 5, 1, 6, padding::valid, true, 1, 1,
           backend_type)  // C1, 1@32x32-in, 6@28x28-out
-     << average_pooling_layer<tan_h>(28, 28, 6,
-                                     2)  // S2, 6@28x28-in, 6@14x14-out
+     << average_pooling_layer(28, 28, 6,
+                              2)  // S2, 6@28x28-in, 6@14x14-out
+     << tanh_layer(14, 14, 6)
      << quantized_convolutional_layer<tan_h>(
           14, 14, 5, 6, 16, connection_table(tbl, 6, 16), padding::valid, true,
           1, 1,
           backend_type)  // C3, 6@14x14-in, 16@10x10-in
-     << average_pooling_layer<tan_h>(10, 10, 16,
-                                     2)  // S4, 16@10x10-in, 16@5x5-out
-     << quantized_convolutional_layer<tan_h>(
-          5, 5, 5, 16, 120, padding::valid, true, 1, 1,
-          backend_type)  // C5, 16@5x5-in, 120@1x1-out
+     << average_pooling_layer(10, 10, 16,
+                              2)  // S4, 16@10x10-in, 16@5x5-out
+     << tanh_layer(5, 5, 16) << quantized_convolutional_layer<tan_h>(
+                                  5, 5, 5, 16, 120, padding::valid, true, 1, 1,
+                                  backend_type)  // C5, 16@5x5-in, 120@1x1-out
      << quantized_fully_connected_layer<tan_h>(
           120, 10, true, backend_type);  // F6, 120-in, 10-out
 }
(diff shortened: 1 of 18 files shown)


examples/mnist/quantized.cpp
```diff
@@ -33,17 +33,18 @@ void construct_net(network<sequential> &nn) {
...
           1, 1,
           backend_type)  // C3, 6@14x14-in, 16@10x10-in
-     << average_pooling_layer<tan_h>(10, 10, 16,
-                                     2)  // S4, 16@10x10-in, 16@5x5-out
-     << quantized_convolutional_layer<tan_h>(
-          5, 5, 5, 16, 120, padding::valid, true, 1, 1,
-          backend_type)  // C5, 16@5x5-in, 120@1x1-out
+     << average_pooling_layer(10, 10, 16,
+                              2)  // S4, 16@10x10-in, 16@5x5-out
+     << tanh_layer(5, 5, 16) << quantized_convolutional_layer<tan_h>(
```
beru: These 2 layers should be written on separate lines. However, you may need to put
```cpp
// clang-format off
...
// clang-format on
```
to do so.

test/test_serialization.h
```diff
@@ -529,7 +529,7 @@ TEST(serialization, sequential_model) {
   network<sequential> net1, net2;
 
   net1 << fully_connected_layer(10, 16) << tanh_layer(16)
-       << average_pooling_layer<relu>(4, 4, 1, 2)
+       << average_pooling_layer(4, 4, 1, 2) << relu_layer(2, 2, 1)
```
beru: The number of layers is increased to 4 from 3. You may update the below for loop condition as well.

beru: We can use `network::layer_size` method to retrieve the number of the layers.

beru: Please update constant loop conditions in other test cases.

karandesai-96:
> I just thought these tests in test/test_serialization.h are doing deserialization tests, so test case names might be a bit improper. But that's something irrelevant to this PR. :thought_balloon:

@beru Looks like I got the purpose of next PRs after decoupling is done :smile: i'll tackle this next.

test/test_serialization.h
karandesai-96: Oops, nasty little detail I missed. Thanks for pointing it out !

examples/mnist/quantized.cpp
karandesai-96: Agreed, that way it would be more readable. changing :white_check_mark:

karandesai-96:
Okay so again 3 commits for 3 layer class - one commit for each. Please make sure to rebase before merge. If there are any other changes required, I'm here :smile:

examples/mnist/quantized.cpp
edgarriba: Also you can get rid of the `_layer` prefix to improve readability. It's redundant information

test/test_serialization.h
edgarriba: Maybe it's out of the scope of this PR but as for `network::layer_size` I propose renaming the function name since at least to me it's ambiguous. You can define the size of a network from a very different perspectives, so I think that `network::num_layers` or similar would fit more with the functionality.

examples/mnist/quantized.cpp
karandesai-96: @edgarriba: I'll do a project wide refactoring / typedef of activation layer names once I get rid of old classes from **activation_function.h**. They are having conflicting names in separate namespaces, as I desicribed earlier. That day is not far now, as decoupling is almost over.

test/test_serialization.h
karandesai-96: Agreed with @edgarriba :+1:

karandesai-96:
@edgarriba Agreed and noted both of your points.

edgarriba:
@karandesai-96 regarding partial connected it should be deprecated, so average pooling must inherit from layer

karandesai-96:
@edgarriba Okay great, is there a related issue or shall I make one for tracking ?

edgarriba:
Open an issue

PR #608, opened 2017-03-06
Drop templating from max_unpooling_layer.

karandesai-96:
The `Activation` template parameter has been completely decoupled from `max_unpooling_layer`. Now only the decoupling of quantization layers remains.

Refactoring procedure is very much similar to #600 and #601.

### DIFF
--- a/tiny_dnn/layers/max_unpooling_layer.h
+++ b/tiny_dnn/layers/max_unpooling_layer.h
@@ -18,11 +18,9 @@ namespace tiny_dnn {
 /**
  * applies max-pooing operaton to the spatial data
  **/
-template <typename Activation = activation::identity>
-class max_unpooling_layer : public feedforward_layer<Activation> {
+class max_unpooling_layer : public layer {
  public:
-  CNN_USE_LAYER_MEMBERS;
-  typedef feedforward_layer<Activation> Base;
+  using layer::parallelize_;
 
   /**
    * @param in_width     [in] width of input image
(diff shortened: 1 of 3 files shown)


tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,74 +60,70 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
-  void forward_propagation(serial_size_t index,
-                           const std::vector<vec_t *> &in_data,
-                           std::vector<vec_t *> &out_data) override {
-    const vec_t &in = *in_data[0];
-    // vec_t&       out = *out_data[0];
-    vec_t &a = *out_data[1];
-    std::vector<serial_size_t> &max_idx =
-      max_unpooling_layer_worker_storage_[index].in2outmax_;
+  void forward_propagation(const std::vector<tensor_t *> &in_data,
+                           std::vector<tensor_t *> &out_data) override {
```
karandesai-96: Planning to make a kernel like other layers on a PR, probably based from master later.

edgarriba: yes, that's the correct way to handle this

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,74 +60,70 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
+                        std::vector<tensor_t *> &out_grad,
+                        std::vector<tensor_t *> &in_grad) override {
+    for (serial_size_t sample = 0; sample < in_data[0]->size(); sample++) {
+      vec_t &prev_delta = (*in_grad[0])[sample];
+      vec_t &curr_delta = (*out_grad[0])[sample];
+
+      std::vector<serial_size_t> &max_idx =
+        max_unpooling_layer_worker_storage_[sample].in2outmax_;
+
+      for_(parallelize_, 0, in2out_.size(), [&](const blocked_range &r) {
```
edgarriba: shouldn't be the outer loop in parallel?

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,74 +60,70 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
-                           std::vector<vec_t *> &out_data) override {
-    const vec_t &in = *in_data[0];
-    // vec_t&       out = *out_data[0];
-    vec_t &a = *out_data[1];
-    std::vector<serial_size_t> &max_idx =
-      max_unpooling_layer_worker_storage_[index].in2outmax_;
+  void forward_propagation(const std::vector<tensor_t *> &in_data,
+                           std::vector<tensor_t *> &out_data) override {
+    for (size_t sample = 0; sample < in_data[0]->size(); sample++) {
+      const vec_t &in = (*in_data[0])[sample];
```
beru: A tiny thing about placement of variables. I don't think writing `in_data[0]` twice is favorable in this case. In original implementation, I see the reference variable for `in_data[0]` outside of the loop. That's the way it should be. 

And since `in_data[0]->size()` does not change, it should be set to a local variable and the local variable should be used in loop condition. But some may disagree with my opinions since additional variable will be needed in this way.

tiny_dnn/layers/max_unpooling_layer.h
beru: Are you concerned about peformance? We can't be sure without testing both of the cases.
The way `for_` function is used is different in `conv2d_op_internal.h` and `fully_connected_op_internal.h`.

tiny_dnn/layers/max_unpooling_layer.h
karandesai-96: @beru I am not sure about the second case but I definitely agree with you on first case. I'm changing that.

tiny_dnn/layers/max_unpooling_layer.h
karandesai-96: @edgarriba The outer loop represents the samples (images) in a batch. We want pooling to be parallelized per sample. So I guess the inner loop should be parallel. @beru what do you think ?

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,74 +60,77 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
-    vec_t &curr_delta = *out_grad[1];
-    std::vector<serial_size_t> &max_idx =
-      max_unpooling_layer_worker_storage_[index].in2outmax_;
+    for (size_t sample = 0; sample < in_data[0]->size(); sample++) {
+      const vec_t &in_vec = in[sample];
+      vec_t &out_vec      = out[sample];
 
-    CNN_UNREFERENCED_PARAMETER(in_data);
+      std::vector<serial_size_t> &max_idx =
+        max_unpooling_layer_worker_storage_[sample].in2outmax_;
```
beru: `max_unpooling_layer_worker_storage_` is reallocated at `max_unpooling_layer::set_worker_count` and its storage size is the same as `worker_count`. So, accessing it with `sample` index is just not right.... right?

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,74 +60,77 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
-                           std::vector<vec_t *> &out_data) override {
-    const vec_t &in = *in_data[0];
-    // vec_t&       out = *out_data[0];
-    vec_t &a = *out_data[1];
-    std::vector<serial_size_t> &max_idx =
-      max_unpooling_layer_worker_storage_[index].in2outmax_;
+  void forward_propagation(const std::vector<tensor_t *> &in_data,
+                           std::vector<tensor_t *> &out_data) override {
+    const tensor_t &in = *in_grad[0];
+    tensor_t &out      = *out_grad[0];
```
beru: In original implementation `in` local variable references `in_data` and in this PR, it references `in_grad`.
How kind that is for reviewers!

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,74 +60,77 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
-                        const std::vector<vec_t *> &out_data,
-                        std::vector<vec_t *> &out_grad,
-                        std::vector<vec_t *> &in_grad) override {
-    vec_t &prev_delta = *in_grad[0];
-    vec_t &curr_delta = *out_grad[1];
-    std::vector<serial_size_t> &max_idx =
-      max_unpooling_layer_worker_storage_[index].in2outmax_;
+    for (size_t sample = 0; sample < in_data[0]->size(); sample++) {
+      const vec_t &in_vec = in[sample];
+      vec_t &out_vec      = out[sample];
```
beru: Is it OK to handle gradients in forward_propagation? In original implementation, `out_data` is handled but with this, `out_grad` is handled.

tiny_dnn/layers/max_unpooling_layer.h
karandesai-96: Silly typing error :sweat_smile: Sorry for the trouble.

tiny_dnn/layers/max_unpooling_layer.h
beru: @karandesai-96 I know outer loop represents the samples of a mini-batch. Good fix.

But it seems to me accessing `max_unpooling_layer_worker_storage_` member variable with `sample` index is wrong.

Talking about whether the outer loop or 2nd loop should be parallelized, I can't say much without actuall testing both of the cases.

karandesai-96:
@beru I have created #612 for further discussions about the matter. This is a separate problem in itself, a whole new one, which needs quite some discussion before implementation. In the mean time my suggestion is to temporarily exclude `max_unpooling_layer` from the layers namespace.

edgarriba:
@beru @Randl I'm about to merge this. Just give +1 sign

karandesai-96:
Gentle reminder, waiting for updates :rocket: :smile:

tiny_dnn/layers/max_unpooling_layer.h
beru: It looks like @karandesai-96 force-pushed a commit to update the code.

beru:
@edgarriba @karandesai-96 
> @beru @Randl I'm about to merge this. Just give +1 sign

According to @karandesai-96 , It looks like current `max_unpooling_layer` is broken.
And it seems to me accessing `max_unpooling_layer_worker_storage_` member variable with `sample` isn't right. I posted the comment but haven't received any response.

So, please clarify your stance about it before merging this PR. I suggest not to use `for_` function if you can't handle multi-threading.

beru:
Excluding `max_unpooling_layer` from the layers namespace is fine by me. But then, what's the point of updating `max_unpooling_layer` with this PR?

tiny_dnn/layers/max_unpooling_layer.h
karandesai-96: @beru Yes I did amend my recent commit.

karandesai-96:
@beru So I will have to make changes with `max_unpooling_layer_worker_stroge_` and remove `for_` as I'm unable to handle multi-threading. We discussed about this on Gitter but I forgot to update things here, sorry for the inconvenience.

karandesai-96:
@beru I just checked out worker storage of other layers (conv, deconv, max pool). They are not vectors, but single instances. Keeping worker storage as a vector and a class member looks like legacy code to me (please correct me if I am wrong). All layer worker storages have been shifted to backend classes, and so this should go there too. I made #612 for discussing this.

Also, I need not remove `for_` loop if I changed `worker_storage_` as the outer for loop is not multi threaded. I made the necessary changes in this commit.

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -62,78 +60,77 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
...
-  void set_worker_count(serial_size_t worker_count) override {
-    Base::set_worker_count(worker_count);
-    max_unpooling_layer_worker_storage_.resize(worker_count);
-    for (max_unpooling_layer_worker_specific_storage &mws :
-         max_unpooling_layer_worker_storage_) {
-      mws.in2outmax_.resize(out_.size());
-    }
+  void init_workers(serial_size_t sample_count) {
+    max_unpooling_layer_worker_storage_.in2outmax_.resize(out_.size());
   }
```
beru: The argument `sample_count` isn't used anymore. What's the point of making this method?
I guess the resizing operation can be simply done in the constructor method.

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -151,7 +148,7 @@ class max_unpooling_layer : public feedforward_layer<Activation> {
       in2outmax_;  // mapping max_index(out) => in (1:1)
   };
 
-  std::vector<max_unpooling_layer_worker_specific_storage>
+  max_unpooling_layer_worker_specific_storage
     max_unpooling_layer_worker_storage_;
```
beru: I think struct name `max_unpooling_layer_worker_specific_storage` is too long. Since it's an inner class, it can be renamed to `worker_specific_storage`. The member variable `max_unpooling_layer_worker_storage_` can also be renamed to just `worker_storage_`.

beru:
@karandesai-96 I checked the original `max_unpooling_layer.h` file but couldn't get how the inner values of `max_unpooling_layer_worker_storage_` are set though its container is resized.

I can't say much about this since I still don't know about `max_unpooling` operation.
Without working examples, it's hard for me to understand.

tiny_dnn/layers/max_unpooling_layer.h
karandesai-96: @beru By current design, it should be shifted to backend class. But that's a thing for the future, for now I'll make the name shorter.

tiny_dnn/layers/max_unpooling_layer.h
karandesai-96: @beru: I amended my recent commit. Worked on your review. :smile:

PR #618, opened 2017-03-13
Update readme.md

karandeepdps:
Updated trained model link(was proving .txt file) with a new model file(99.13% accuracy).

### DIFF
--- a/examples/mnist/readme.md
+++ b/examples/mnist/readme.md
@@ -222,7 +222,7 @@ int main(int argc, char **argv) {
 >Note:
 >Each image has 32x32 values, so dimension of first layer must be equal to 1024.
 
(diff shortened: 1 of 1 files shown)


edgarriba:
we should first discuss internally where do we host trained examples

beru:
`examples/cifar-weights` file was added in #590 
But I personally prefer not to include binary files into git repository.
How about attaching files to pull requests and mentioning its URLs in documents?

karandeepdps:
@beru and which on which site we should host files.

beru:
> and which on which site we should host files.
@karandeepdps We still haven't reached to a conclusion.
I think there are a few candidates.

1. cloud storages like GoogleDrive, Dropbox, 
2. adding the model file into repository just like #590 did
3. attaching the file to this PR comment and write its URL into `readme.md`

What do you think? @edgarriba

PR #613, opened 2017-03-09
Drop templating from quantization layers.

karandesai-96:
The `Activation` template parameter has been completely decoupled from the following layers:

1. `quantized_convolution_layer`
2. `quantized_deconvolution_layer`
3. `quantized_fully_connected_layer`

Changes have been made in layer implementation, tests, examples and serialization helpers.

### Future Roadmap:

This PR marks a checkpoint in decoupling activations job. All the layers act as standalone layers with no implicit application of activation function. Next, I will safely remove **activation_function.h** and put the newly defined _activation layers_ in `tiny_dnn::activation` namespace.

PR #619, opened 2017-03-14
Remove old activation functions and feedforward layer.

karandesai-96:
This PR concludes decoupling of activations procedure. Notable changes made here are:

- Removal of **activation_function.h**. Old namespace `tiny_dnn::activation` has been completely scraped off. New namespace `tiny_dn::activation has been created in **tiny_dnn/tiny_dnn.h** with short hand usages of all activation layers.

- Removal of redundant macros in serialization helpers. We don't need `REGISTER_CNN_LAYER_WITH_ACTIVATIONS` anymore.

- Removal of **feedforward_network.h** - it formed the basis of a teplated layer with activation. Now its existence is unworthy, deleted it off.

- Equivalent minor changes in serialization tests and a few examples here and there, to make builds pass.


karandesai-96:
Note: This PR must be merged only after #613 is merged.

edgarriba:
don't forget to update docs with new API

karandesai-96:
@edgarriba: I am currently writing "Adding your custom activation", and updating _README_ with new API. A Pull Request should be ready by tomorrow.

karandesai-96:
/ping: @beru @edgarriba I have two branches for upcoming PRs, which depend on this PR. Please review this and we can get it merged. :rocket:

karandesai-96:
The docs API has been modified in a separate branch. I have it locally, will push it immediately later and open a PR.

tiny_dnn/activations/activation_function.h
```diff
@@ -1,241 +0,0 @@
-/*
-    Copyright (c) 2013, Taiga Nomi
-    All rights reserved.
-
-    Use of this source code is governed by a BSD-style license that can be found
-    in the LICENSE file.
-*/
-#pragma once
```
edgarriba: @beru is there any too keep backward compatibility? with old api call new one?

beru: @edgarriba There can be a way to keep backward compatibility with old API through indirectly calling new one. One idea is using function templates that prototypes resemble old layer classes constructors. But this may cause overloading ambiguity and we don't have tests for old API anymore, so forfeiting old API looks reasonable decision to me.

tiny_dnn/tiny_dnn.h
```diff
@@ -116,6 +109,26 @@ using ave_unpool = tiny_dnn::average_unpooling_layer;
 
 }  // namespace layers
 
+namespace activation {
+
+using sigmoid = tiny_dnn::sigmoid_layer;
+
```
beru: Are these line spaces needed?

karandesai-96: Followed it like layers namespace above, just for the sake of uniformity. Shall I trim it off ?

tiny_dnn/activations/activation_function.h
edgarriba: Sure, just asking

tiny_dnn/tiny_dnn.h
beru: OIC. Then that should be fine!

examples/cifar10/test.cpp
```diff
@@ -8,7 +8,7 @@ using namespace std;
 // rescale output to 0-100
 template <typename Activation>
 double rescale(double x) {
-  Activation a;
+  Activation a(1);
```
beru: Making an instance just to call `scale` method doesn't look right but the method can't be `static` since it's `virtual`.

edgarriba:
ok, merging

PR #624, opened 2017-03-16
Xun 

XunJiao:


Randl:
Please explain what is the scope of PR.
Remove binary files, such as `*.o` and also logs.

0.00000001.log
```diff
@@ -0,0 +1,50 @@
+load models...
```
Randl: Please, remove logs

Makefile
```diff
@@ -0,0 +1,15 @@
+quantized:
```
Randl: You shouldn't change Makefile by yourself, just cmake files

cmake-uninstall.cmake
```diff
@@ -0,0 +1,21 @@
+if(NOT EXISTS "/home/rose/tiny-dnn/install_manifest.txt")
```
Randl: Why are all those files are in root folder?

Randl:
@XunJiao feel free to reopen when the problems are fixed

PR #531, opened 2017-01-22
Enable Coverity plug-in in Travis CI

kecsap:
The tiny-dnn project must be registered for Coverity and the API token changed in the commit before merge.

PR #626, opened 2017-03-20
Enable Coverity-Scan API token in Travis

nyanp:
This is continued from #531. This PR replaces dummy API token with encrypted token for coverity scan-build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,12 @@ language: cpp
 
 os:
   - linux
-  - osx
+  # use just one of the environments in coverity_scan branch in order not to consume too much API
+  # - osx
 
 compiler:
   - gcc
-  - clang
+  # - clang
 
 cache:
   pip: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,9 @@ env:
     - BUILD_TESTS=ON
     - BUILD_EXAMPLES=ON
     - COVERALLS=ON
-    - secure: "TOKEN"
+   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
+   #   via the "travis encrypt" command using the project repo's public key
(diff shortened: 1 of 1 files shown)


Randl:
I suppose to change notification email, and also branch. Master probably would mean checks are performed too often.

nyanp:
@Randl It's up to you to change your notification settings :) Any suggestions to avoid interrupting our mailbox?

Randl:
Well there is that site, https://www.kill-the-newsletter.com/ , which converts emails to RSS

edgarriba:
is there any way to show results in a badge? https://shields.io/

Randl:
I don't think so

edgarriba:
well I see this: https://img.shields.io/coverity/scan/3997.svg
however not sure which output gives

nyanp:
Coverity allows us to submit 4 builds/day.
https://scan.coverity.com/faq#frequency

We might need to separate the coverity build to ```coverity_scan``` branch and invoke build by hand (as they recommend at https://scan.coverity.com/faq#frequency) in order to avoid build breaking due to the API restriction.

I met an another problem around TBB with clang (clang 3.5.0 couldn't find ```std::is_tribially_copyable``` in the latest version of TBB https://travis-ci.org/tiny-dnn/tiny-dnn/jobs/214276470), but it should be treated as an another issue. I'll investigate the problem tomorrow.

bhack:
Keras removed coverity cause was not satisfied by the quality of the service.. probably clang static coverage is enough..

nyanp:
@bhack 
After struggling with that, I reached to the similar mind to Keras team... Anyway I think it's ok to keep ```coverity_scan``` branch as is and we should remove coverity API from master branch.

nyanp:
@edgarriba 
> I think it's ok to keep coverity_scan branch as is and we should remove coverity API from master branch.

What do you think about removing the coverity build from master branch?

edgarriba:
go ahead

nyanp:
ok👌

PR #629, opened 2017-03-25
use new FindTBB.cmake

beru:
Hi, this PR replaces `cmake/Modules/FindTBB.cmake` with new one obtained from https://github.com/justusc/FindTBB


beru:
> Have you tried?

@edgarriba Yes. I tested on Windows (Visual Studio 2015) and Linux.
New `FindTBB.cmake` supports recent versions of Visual Studio while the one that comes with `tiny-dnn` doesn't so that's why I created this PR.

PR #639, opened 2017-03-31
Provide automatic shape inference in activation layers.

karandesai-96:
This PR makes changes in the activation layers API, and allows the user to declare an activation layer without specifying its dimensions.

#### Prior to this PR:
```c++
nn << fully(784, 30) << relu(30) << fully(30, 10) << softmax(10);
```
The constructors required layer dimension to be specified.

#### Changes Introduced:
```c++
nn << fully(784, 30) << relu() << fully(30, 10) << softmax();
```
Note that the older way of declaration is still valid. For this to work properly, the `out_shape` of previous layer attached to any activation layer must be defined properly. (`fully` layers in our example here).

#### Implementation Details:

* A new virtual method `set_in_shape` is introduced in `layer` class, which currently throws an `nn-error` by default, as shape inference is currently not supported for any other layers except activation layers. Activation layers override this method.

* On declaring an activation layer without passing input shape, it sets the input shape to `shape3d(0, 0, 0)` hence its `size` is zero.

* In the `<<` operator ( of the format `layer1 << layer2`, where `layer2` is activation layer ), the size of `layer2` is checked, which if zero, then `in_shape` of `layer2` is set to be same as `out_shape` of `layer1`. This is how shape inference takes place in activation layers.

#### Future Plans:

This shape inference can be extended to all the layers gradually, and then finally the user will have to specify the input shape at only the first layer of the network, while all the intermediate shapes of tensors at different layers will be automatically deduced while stacking layers using `<<` operator.

### DIFF
--- a/test/test_average_pooling_layer.h
+++ b/test/test_average_pooling_layer.h
@@ -13,13 +13,13 @@
 namespace tiny_dnn {
 
 TEST(ave_pool, gradient_check) {  // sigmoid - cross-entropy
-  using loss_func        = cross_entropy;
-  using activation_layer = sigmoid_layer;
-  using network          = network<sequential>;
+  using loss_func  = cross_entropy;
+  using activation = sigmoid;
+  using network    = network<sequential>;
 
   network nn;
-  nn << fully_connected_layer(3, 8) << activation_layer(8)
-     << average_pooling_layer(4, 2, 1, 2) << activation_layer(2);  // 4x2 => 2x1
+  nn << fully_connected_layer(3, 8) << activation()
+     << average_pooling_layer(4, 2, 1, 2) << activation();  // 4x2 => 2x1
 
   const auto test_data = generate_gradient_check_data(nn.in_data_size());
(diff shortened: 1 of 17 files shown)


Randl:
Can we do the same for non-activation layers?    🙄

karandesai-96:
@Randl definitely, but I feel that `set_in_shape` will not be straightforward and common for all layers, since some layers do not store their dimensions as a shape3d variable ( specific layers have them in, say `fully_params` or `conv_params` ). So my plan is to combine all the changes in one PR and base it off master.

Later, we can convert `set_in_shape` of base layer class to be pure virtual, does it sound good ?

edgarriba:
LGTM

edgarriba:
@karandesai-96 what's he status of this branch? should we start merging to master?

karandesai-96:
@edgarriba Yes, one PR is remaining to be merged, that is #643. It just includes documentation modifications.

PR #644, opened 2017-04-05
Add several command line arguements to example mnist train program.

beru:
Hi, this PR adds several command line arguments to example mnist train program.
This follows the work done at #590.
I will update `readme.md` after reviewers checked the source code.

### DIFF
--- a/examples/mnist/train.cpp
+++ b/examples/mnist/train.cpp
@@ -11,7 +11,8 @@
 using namespace tiny_dnn;
 using namespace tiny_dnn::activation;
 
-static void construct_net(network<sequential>& nn, core::backend_t backend_type) {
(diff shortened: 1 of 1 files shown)


examples/mnist/train.cpp
```diff
@@ -55,12 +51,16 @@ static void construct_net(network<sequential>& nn) {
                                      true, backend_type);
 }
 
-static void train_lenet(const std::string& data_dir_path) {
+static void train_lenet(std::string data_dir_path,
```
edgarriba: Why non const and copy object?

beru: That's because I copied the code from cifar10 train.

PR #647, opened 2017-04-07
Update Dockerfile and fix build problem

vowstar:
The old Dockerfile have some problem while building NNPACK, such as missing confu, mismatched ninja version.

Before fix

```bash
...
Cloning into 'NNPACK'...
Traceback (most recent call last):
File "./configure.py", line 4, in <module>
import confu
ImportError: No module named confu
Removing intermediate container a7d5d50c6b88
The command '/bin/sh -c apt-get install ninja-build && pip install ninja-syntax && cd /software && git clone --recursive https://github.com/Maratyszcza/NNPACK.git && cd /software/NNPACK && python ./configure.py && ninja' returned a non-zero code: 1
```

After fix

```bash
...
[152/156] LINK bin/convolution-inference+relu-smoketest
[153/156] CXX test/softmax-output/smoke.cc
[154/156] LINK bin/softmax-output-smoketest
[155/156] CXX test/max-pooling-output/smoke.cc
[156/156] LINK bin/max-pooling-output-smoketest
---> 60abb17a95c2
Removing intermediate container 0b970dd3258e
# build success
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ RUN cd /software && \
 # Download and configure NNPACK
 RUN apt-get install ninja-build && \
     pip install ninja-syntax && \
+    pip install confu && \
     cd /software && \
     git clone --recursive https://github.com/Maratyszcza/NNPACK.git && \
     cd /software/NNPACK && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:17.04
 
 MAINTAINER Edgar Riba <edgar.riba@gmail.com>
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:17.04
```
Randl: Should we stay on LTS?

vowstar: The ninja version is too old in TLS, NNPACK can't compile. Docker CE stable version only support 1 year, so it's meaningless for using ubuntu LTS.

edgarriba: ping @bhack

bhack: Can we wait 13th April for merge this?

vowstar: ping @bhack

PR #651, opened 2017-04-09
Introduce the decoupled activations and layers API on master.

karandesai-96:
This PR finally brings the decoupled layers and activations API to master branch. Templating was removed to reduce build time and runtime memory usage. Profiling results are favourable. API is simpler than before.

#### Old API style:
```c++
net << conv<tan_h>(32, 32, 5, 1, 6)  // in:32x32x1, 5x5conv, 6fmaps
    << ave_pool<tan_h>(28, 28, 6, 2) // in:28x28x6, 2x2pooling
    << fc<tan_h>(14 * 14 * 6, 120)   // in:14x14x6, out:120
    << fc<identity>(120, 10);        // in:120,     out:10
```

#### New API style:
```c++
net << conv(32, 32, 5, 1, 6) << tanh()  // in:32x32x1, 5x5conv, 6fmaps
    << ave_pool(28, 28, 6, 2) << tanh() // in:28x28x6, 2x2pooling
    << fc(14 * 14 * 6, 120) << tanh()   // in:14x14x6, out:120
    << fc(120, 10);                     // in:120,     out:10
```

This PR breaks backward compatibility. Related PRs which were merged into `feat/decouple_activations` to incrementally carry out this work: 

* #529: Initial playground: activation_layer base class with relu_layer inheriting it.
* #554: Use pass-by-reference instead of pass-by-value in backward_activation.
* #564: Child classes of activation_layer for all currently supported activations.
* #573: Drop template completely from fully_connected_layer.
* #577: Serialization of activation layers.
* #600: Drop activation template from conv and deconv layers.
* #601: Drop activation template from max_pooling_layer.
* #604: Drop activation template from average pool, unpool layers.
* #605: Drop activation template from linear and lrn layers.
* #608: Drop templating from max_unpooling_layer.
* #613: Drop templating from quantization layers.
* #619: Remove old activation functions and feedforward layer.
* #631: Documentation for writing custom activations in tiny-dnn.
* #639: Provide automatic shape inference in activation layers.
* #643: Modify docs as per new activation API.

#### Caution: Commit hashes have changed due to rebase of branch.

Thanks @beru @bhack @edgarriba @nyanp @Randl for extensive reviewing cycles and design discussions.

PR #657, opened 2017-04-16
fix cifar10 train

beru:
Hi, this PR fixes `example_cifar_train` to run correctly.
I didn't test the program while reviewing #651, so couldn't find a bug.
https://github.com/tiny-dnn/tiny-dnn/pull/651/files#diff-b472b6857897337fec21ee643637c3f3R60


karandesai-96:
Silly typo while decoupling! :smile: Thanks for rectification @beru..

PR #658, opened 2017-04-16
Parallelize activation_layer forward_propagation and backward_propagation methods

beru:
Hi, this PR parallelizes `activation_layer`'s `forward_propagation` and `backward_propagation` methods.

After merging #651, `example_mnist_train` significantly slows down.

In old `feedforward_layer` class, `for_i` function is used inside `forward_activation` and `backward_activation` methods, so this PR recovers it.
https://github.com/tiny-dnn/tiny-dnn/pull/651/files#diff-b5252a229ea3c08ea12d86fe29fa5cd2L28

### DIFF
--- a/examples/mnist/train.cpp
+++ b/examples/mnist/train.cpp
@@ -37,25 +37,23 @@ static void construct_net(network<sequential> &nn,
   nn << convolutional_layer(32, 32, 5, 1,
                             6,  // C1, 1@32x32-in, 6@28x28-out
                             padding::valid, true, 1, 1, backend_type)
-     << tanh_layer(28, 28, 6)
-     << average_pooling_layer(28, 28, 6,
-                              2)  // S2, 6@28x28-in, 6@14x14-out
-     << tanh_layer(14, 14, 6)
+     << tanh_layer() << average_pooling_layer(28, 28, 6,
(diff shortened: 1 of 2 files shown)


examples/mnist/train.cpp
```diff
@@ -37,25 +37,23 @@ static void construct_net(network<sequential> &nn,
   nn << convolutional_layer(32, 32, 5, 1,
                             6,  // C1, 1@32x32-in, 6@28x28-out
                             padding::valid, true, 1, 1, backend_type)
-     << tanh_layer(28, 28, 6)
-     << average_pooling_layer(28, 28, 6,
-                              2)  // S2, 6@28x28-in, 6@14x14-out
-     << tanh_layer(14, 14, 6)
+     << tanh_layer() << average_pooling_layer(28, 28, 6,
+                                              2)  // S2, 6@28x28-in, 6@14x14-out
+     << tanh_layer()
```
karandesai-96: How about clang formatting guards around this declaration ? Indentation looks slightly distorted to me right now.

beru: I learned to stop worrying about indentation and love '.clang-format' uglify the code.

edgarriba: it can be solved by reducing the variables name size, but yeah that's another war

Issue #615, opened 2017-03-10
Question: how to use concat and slice


OleStauning:
Lets say I have a network defined by:
```
net << fc<relu>(32,32)
        << fc<relu>(32,10)
        << dropout(10,0.5)
        << fc<sigmoid>(10,3);
```
However, Instead if the first fully connected layer 32x32, I want to slice the network into 4 times 8 input nodes, each of the 4 slices should then be fed into a fully connected network 8x8 and thereby sharing the weights. After this I will concat the 4 times 8 nodes and continue with the 32x10 fully connected layer.

Can this be done? - with correct back-propagation of derivatives wrt. the shared weights?

Maybe it is easy, I just cannot find any code-examples where split and slice are used in networks...

Best regards

edgarriba:
I haven't played at all with graph networks but at a first glimpse I would say that `cc` should have a size of `cc(64, 64)`. I suggest you to build your model incrementally.

@wangyida last year you played around with this. Any idea?

OleStauning:
@edgarriba thanks for your answer.

I'm pretty sure the dimensions are correct in the network; otherwise I get an `layer dimension mismatch!` exception ( cc(64,64) gives this error ).

I worry that tiny-dnn does not support sharing of weights in the way I want. Maybe result of the forward- and backward passes are tightly coupled with the object representing the layer in the graph. This design would constrain the usage of layers so that they can only be used once in each evaluation. Maybe somebody can tell if this is the case?

Best regards

karandesai-96:
Putting a reminder for myself to explain this after this PR mentioned above gets merged and gives no bugs.

Issue #257, opened 2016-08-03
how to connect a concat layer


szad670401:
i wanna train a network that  last softmax output layer can be a concat layer which with multi-label output.

edgarriba:
@szad670401 
There are few examples of graph usage yet, feedback is always welcomed.

@Wangyida was experimenting with multi input/output graphs. Any update on that?

szad670401:
I follow the docs and designed graph model. but the concat layer seems can connected with multi-softmax-layer.
here's is my code.

> 
void gragh_model()
{
    network<graph>    net;
    shape3d x;

    layers::conv<relu>     in2(136,36,5,1,6);
    layers::ave_pool<relu> in3(132,32,6,2);
    layers::conv<relu>     in4(66,16,7,6,6);
    layers::ave_pool<relu> in5(60,10,6,2);
    layers::fc<relu> fcx(30*5*6,40);
    layers::concat  concat(4,10);
    layers::fc<softmax> out1(10,10);
    layers::fc<softmax> out2(10,10);
    layers::fc<softmax> out3(10,10);
    layers::fc<softmax> out4(10,10);
    adagrad opt;
    concat<<(out1,out2,out3,out4);
    in2<<in3<<in4<<in5<<fcx<<concat;
    
   // fcx << (out1,out2,out3,out4);
    construct_graph(net, { &in2 }, { &out1,&out2,&out3,&out4});

// 2training data, each data type is tensor_t and shape is [2x2]
//
//                        1st data for in1       2nd data for in1
//                              |                      |
//                              |   1st data for in2   |   2nd data for in2
//                              |         |            |         |
//    std::vector<tensor_t> data{ { { 1, 0 }, { 3, 2 } },{ { 0, 2 }, { 1, 1 } } };
//    std::vector<tensor_t> out { {           { 2, 5 } },{           { 3, 1 } } };
//
//    net.fit<mse>(opt, data, out, 1, 1);

}

Wangyida:
You can have a reference on this example for [ENet bottleneck module](https://github.com/Wangyida/tiny-cnn/blob/feat/generic-computational-graph/examples/deconv/train.cpp#L151-#L181). These lines constructed a graph based network in tiny-cnn.

szad670401:
the ENet just can output single label. I wanna output multi-label.
the concat layer seems can only input with multi-layers but can't output with multi-layers.

0x7ace80:
Hi,
I have the same issue.
[...]

PR #662, opened 2017-04-18
Replace node_tuple with layer_tuple to fix graph networks.

### LINKED ISSUE #146: Split and Concat layers
Hi @nyanp,

great work with tiny-cnn! It should be super helpful for CNN deployment!

I was just testing its abilities and tried a GoogLeNet architecture, when I noticed that it does not support concat/split layers (error: tiny-cnn does not support this layer type:Split). This should be quite easy to implement and would extend its 'range' quite a bit with networks requiring this capability.

### LINKED ISSUE #578: Misused comma operator in serialization test
In https://github.com/tiny-dnn/tiny-dnn/blob/master/test/test_serialization.h#L408
```cpp
s1 << (f2, f3) << c4;
```
intends to connect both f2 and f3 to s1 and c4, but it actually connects only f3 (this is how comma operator works).

Solution proposed by @edgarriba is to use a dynamic array (probably initialization list would work nice here: `s1 << {f2, f3} << c4;`) instead.

karandesai-96:
This PR replaces `node_tuple` with a `layer_tuple` and puts a constraint on its formation in such a way that the passed operands to `,` comma operator have layer class as its ancestor.

The comma operator for node_tuple didn't work if both operands were a different type. Hence one couldn't do `(l1, l2) << l3` and form a graph network if l1 and l2 were not the instances of same type of layer.

Fixes #578. 
Resolves #146, #257, #615.

### DIFF
--- a/test/test_node.h
+++ b/test/test_node.h
@@ -119,8 +119,8 @@ TEST(layer_tuple, chain_operator_tuple_layer1) {
   sigmoid_layer sgm(4);
 
   (fc1, fc2) << conc << sgm;
-  EXPECT_EQ(fc1.next_nodes()[0], &conc);
-  EXPECT_EQ(fc2.next_nodes()[0], &conc);
+  EXPECT_EQ(conc.prev_nodes()[0], &fc1);
+  EXPECT_EQ(conc.prev_nodes()[1], &fc2);
   EXPECT_EQ(conc.next_nodes()[0], &sgm);
 }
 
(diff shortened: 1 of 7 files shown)


karandesai-96:
Example formation of a graph network:
graph

As can be seen, a `fully_connected` layer and a `tanh` layer could be concatenated successfully now. This graph is a formation from code sample by https://github.com/tiny-dnn/tiny-dnn/issues/578#issuecomment-282402095

tiny_dnn/node.h
```diff
@@ -197,40 +209,37 @@ inline std::shared_ptr<U> &operator<<(std::shared_ptr<T> &lhs,
...
-  for (serial_size_t i = 0; i < static_cast<serial_size_t>(rhs.nodes_.size());
-       i++) {
-    connect(&lhs, rhs.nodes_[i], i, 0);
+template <typename T>
+inline layer_tuple<layerptr_t> &operator<<(T &lhs,
+                                           const layer_tuple<layerptr_t> &rhs) {
+  for (size_t i = 0; i < static_cast<size_t>(rhs.layers_.size()); i++) {
+    connect(&lhs, rhs.layers_[i], i, 0);
   }
   return rhs;
```
Randl: Seems like returning `rhs` as non-const reference is a problem since you pass it a const reference. Either add `const` qualifier to return type or drop it from parameter type (first is better, if possible)

Randl:
Can you add tests for the new functionality?

karandesai-96:
@Randl sure. On a separate note, the coverage drop is due to something else, I'm confused how.

karandesai-96:
@edgarriba Every time I make some changes, some of all jobs fail. The failure states that `layer.prev_nodes()` are ordered in a different way than specified in a `layer_tuple`.

`l1 << (l2, l3)` should have `l1.prev_nodes()` as `{&l2, &l3}` but on some jobs it is `{&l3, &l2}`, can't understand why.

tiny_dnn/node.h
karandesai-96: Right, that rectified the error.

test/test_node.h
```diff
@@ -92,4 +92,73 @@ TEST(layer_tuple, comma_operator_layer_tuple2) {
...
+  EXPECT_EQ(fc2->next_nodes()[0], sgm.get());
+}
+
+TEST(layer_tuple, chain_operator_tuple_layer1) {
+  fully_connected_layer fc1(2, 2);
+  fully_connected_layer fc2(2, 2);
+  concat_layer conc({shape3d(2, 1, 1), shape3d(2, 1, 1)});
+  sigmoid_layer sgm(4);
+
+  (fc1, fc2) << conc << sgm;
```
edgarriba: Just to see the logic working, we could also check previous nodes from conc

test/test_node.h
```diff
@@ -92,4 +92,73 @@ TEST(layer_tuple, comma_operator_layer_tuple2) {
...
+}
+
+TEST(layer_tuple, chain_operator_layer_tuple1) {
+  convolutional_layer conv(5, 5, 3, 1, 6);
+  slice_layer slice(shape3d(3, 3, 6), slice_type::slice_channels, 2);
+  tanh_layer tanh(3, 3, 3);
+  relu_layer relu(3, 3, 3);
+
+  conv << slice << (tanh, relu);
+  EXPECT_EQ(slice.prev_nodes()[0], &conv);
```
edgarriba: And here check next

test/test_node.h
```diff
@@ -92,4 +92,73 @@ TEST(layer_tuple, comma_operator_layer_tuple2) {
...
+}
+
+TEST(layer_tuple, chain_operator_layer_tuple2) {
+  auto conv  = std::make_shared<convolutional_layer>(5, 5, 3, 1, 6);
+  auto slice = std::make_shared<slice_layer>(shape3d(3, 3, 6),
+                                             slice_type::slice_channels, 2);
+  auto tanh = std::make_shared<tanh_layer>(3, 3, 3);
+  auto relu = std::make_shared<relu_layer>(3, 3, 3);
+
+  conv << slice << (tanh, relu);
```
edgarriba: And next

test/test_node.h
karandesai-96: There was a check, but the order in `prev_nodes ()` vector gets jumbled sometimes. I am unable to compare pointers directly based on index.

This problem happens only on clang.

edgarriba: Did you check locally with clang?

karandesai-96: Yes, it fails. Address of `fc1` is `prev_nodes()[1]` and address of `fc2` is `prev_nodes()[0]` when I use clang, hence the test fails.

karandesai-96: @edgarriba shall I get it done by #ifdef macros ?

edgarriba: No macros, we have to make sure that this behavior doesn't lead to future bugs

tiny_dnn/node.h
```diff
@@ -158,79 +158,105 @@ inline std::vector<node *> node::next_nodes() const {
...
 }
 
-template <typename T, typename U>
-inline U &operator<<(const node_tuple<T> &lhs, U &rhs) {
-  for (serial_size_t i = 0; i < static_cast<serial_size_t>(lhs.nodes_.size());
-       i++) {
-    connect(&*lhs.nodes_[i], &*rhs, 0, i);
+template <typename T>
+inline T &operator<<(const layer_tuple<std::shared_ptr<layer>> &lhs, T &rhs) {
+  for (size_t i = 0; i < static_cast<size_t>(lhs.layers_.size()); i++) {
```
beru: How about using Range-based for loop?

Randl: Anyway, why do we need cast here?

karandesai-96: @beru we need to provide index of node during connection to maintain order during operations such as channel concatenation during merging. So range-based for loops might not be a good choice.

beru: @karandesai-96 You're right. I didn't check the argument.

beru:
I wonder if
```cpp
typedef layer *layerptr_t;
```
is really required....

karandesai-96:
@beru @edgarriba how about now?

beru:
One thing I noticed is `const_cast` is used in several places.
```cpp
const_cast<layer *>(this)->ith_out_node
const_cast<layer *>(this)->ith_in_node
```
Maybe having `const` versions of `layer::ith_in_node` and `layer::ith_in_node` might be better?

beru:
@edgarriba Please check the change.

edgarriba:
LGTM

PR #664, opened 2017-04-20
fix missing 'a' -> 'out' renaming in commit b16eae6

mhfan:


karandesai-96:
Travis passed. Let's merge this :smile:

Issue #620, opened 2017-03-14
Cmake Log too noisy after layer-activation decoupling took place.


karandesai-96:
The build log gives out too many warnings after decoupling took place. As of now, when I made this issue, these are two links of latest Travis build logs from `master` and `feat/decouple_activations`:

1. `master`: https://travis-ci.org/tiny-dnn/tiny-dnn/builds/207023926
2. `feat/decouple_activations`: https://travis-ci.org/tiny-dnn/tiny-dnn/builds/210812939

Clearly can be seen that builds on master are quite silent. The warnings are trivial, but they have grown to be a huge number, so this issue is just to keep a track of them all.

karandesai-96:
Can't be totally fixed as of now, because a few indices are stored as `serial_size_t`. Well if we are referring to indices of an array, why is `serial_size_t` required ? An int should suffice..

So I guess relates to #553. `serial_size_t` to be restricted in serialization routines, `size_t` as place holders for values of activations / weights / biases etc. and plain `int` to refer to indices of an array / vector.

beru:
@karandesai-96 I think `size_t` is suitable for indexing arrays in most cases.
The type of the return value of `std::vector::size()` method is `size_t`.
http://www.cplusplus.com/reference/vector/vector/size/

Randl:
@karandesai-96 indices are actually unsigned. `size_t` is the variable to use for sizes, indices, etc.

PR #667, opened 2017-04-22
Reduce warnings

beru:
Hi, this PR reduces compiler warnings.


karandesai-96:
Partially fixes #620

tiny_dnn/layers/max_unpooling_layer.h
```diff
@@ -88,7 +88,9 @@ class max_unpooling_layer : public layer {
       for_(parallelize_, 0, in2out_.size(), [&](const blocked_range &r) {
         for (int i = r.begin(); i < r.end(); i++) {
           const auto &in_index = out2in_[i];
-          out_vec[i] = (max_idx[in_index] == i) ? in_vec[in_index] : float_t{0};
+          out_vec[i]           = (static_cast<int>(max_idx[in_index]) == i)
```
Randl: Why cast bool condition to int?

beru: Not casting bool condition to `int` here, casting `serial_size_t` to `int` to remove a warning.

Randl: Oh, right, I misread it.
Does it make sense to make `blocked_range` elements of type `serial_size_t` then (not really sure how this part works)?

beru: I wish it `size_t` was used from the beginning but since it's not, I need to be a machine to add type castings all over the place.

Randl: Simple replace in typedef won't do the thing? Except redundant/missing casts which might be added/removed slowly afterwards.

beru: Maybe so, C++ compiler can verify the code but I still haven't tried it.

edgarriba:
LGTM

edgarriba:
@beru you forgot this
https://github.com/tiny-dnn/tiny-dnn/blob/master/test/test_convolutional_layer.h#L21

could you recheck the other layers?

beru:
@edgarriba 
> @beru you forgot this
https://github.com/tiny-dnn/tiny-dnn/blob/master/test/test_convolutional_layer.h#L21
>
> could you recheck the other layers?

I couldn't get what you're trying to say. I forgot to do what?
Sth like this?
https://github.com/tiny-dnn/tiny-dnn/pull/667/files#diff-cc023dbe1a164eea581096cc7945bf58
You mean I forgot to add the unsigned-suffix `u` as well?
Does it give any specific warning? I couldn't find it on travis log.

By the way, I've just found other warnings on travis log so I will create another PR someday.

edgarriba:
not warning, just to normalize code

beru:
I see.... I guess I have to clarify why I have to put unsigned-suffix to remove the warning.
And if it's necessary to do so, maybe we should normalize the style with tests.
But don't you think it's a little bit tedious? I've been lazy over the course of my life.

PR #669, opened 2017-04-23
Some optimizations

beru:
Hi, this PR adds some optimizations.
About 5% faster than master branch with `example_cifar_train` program.

### DIFF
--- a/tiny_dnn/util/product.h
+++ b/tiny_dnn/util/product.h
@@ -497,30 +497,9 @@ inline void reduce(const typename T::value_type *src,
   }
 }
 
-template <typename T, typename dst_aligned>
-inline void fill_impl(typename T::value_type *dst,
-                      std::size_t size,
-                      typename T::value_type value) {
-  auto vals   = T::set1(value);
-  auto sz     = T::unroll_size;
-  auto sz4    = T::unroll_size * 4;
-  auto n4     = size / sz4;
-  auto n1     = (size % sz4) / sz;
-  auto remain = size % sz;
-  for (size_t i = 0; i < n4; ++i) {
-    T::template store<dst_aligned>(&dst[i * sz4 + sz * 0], vals);
-    T::template store<dst_aligned>(&dst[i * sz4 + sz * 1], vals);
-    T::template store<dst_aligned>(&dst[i * sz4 + sz * 2], vals);
-    T::template store<dst_aligned>(&dst[i * sz4 + sz * 3], vals);
-  }
-  size_t idx = n4 * sz4;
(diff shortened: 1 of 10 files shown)


tiny_dnn/util/product.h
```diff
@@ -217,218 +257,357 @@ struct double_avx {
...
-                           typename T::value_type *dst) {
-  for (std::size_t i = 0; i < size / T::unroll_size; i++) {
-    typename T::register_type d = T::load(&dst[i * T::unroll_size]);
-    typename T::register_type s = T::load(&src[i * T::unroll_size]);
-    T::store(&dst[i * T::unroll_size], T::add(d, s));
+void fill(T *dst, size_t size, T value) {
+// On recent x86/x64 processors, stos instructions are faster
+// than SSE/AVX store instructions and recent compilers generates
+// stos instructions with std::fill
+#if 1
```
Randl: What's the point of this?

Randl:
Please, rebase on master

tiny_dnn/util/product.h
beru: This comment explains why simple `std::fill` is used over inhouse `vectorize::detail::fill_impl`.
Simple unrolled SSE/AVX store instructions isn't the fastest with recent x86/x64 processors.

With Visual Studio, there're intrinsic functions for `stosd` instruction but I can't find equivalent once for gcc/clang. Fortunately, compilers follow `64-ia-32-architectures-optimization-manual.pdf` so they produce right instructions.

beru:
Rebased.

tiny_dnn/util/product.h
Randl: I meant, why do we need `#if 1`?

beru: A huge advantage of using `#if 1` is I can switch the code by change a letter.
But a lot of people blamed me for leaving dead code and some reasons.

Randl: But is it likely that you need to switch it on fly often? If not, isn't it better to just remove it and recover from history if needed?

beru: No, I don't switch it often. If we remove the code, I think we will be very likely to forget the reason why it is buried.

Randl: So maybe leave to comment but remove the code?

beru: I removed the code with recent commit. We can recover it later if we find it useful.

edgarriba:
LGTM

@beru hope you know what you do, I don't speak avx :smile:
Again, we should include some regression tests to see this kind of improvements

PR #673, opened 2017-04-28
Reduce use of preprocessor macro

beru:
Hi, this PR reduces use of preprocessor macros.

### DIFF
--- a/tiny_dnn/util/serialization_layer_list.h
+++ b/tiny_dnn/util/serialization_layer_list.h
@@ -1,25 +1,32 @@
 // Copyright (c) 2017, Taiga Nomi
-#ifndef CNN_NO_SERIALIZATION
+#pragma once
 
-CNN_REGISTER_LAYER(average_pooling_layer, avepool);
-CNN_REGISTER_LAYER(batch_normalization_layer, batchnorm);
-CNN_REGISTER_LAYER(concat_layer, concat);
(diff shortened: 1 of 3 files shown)


beru:
@karandesai-96 
Thank you for the comment.
I'm not really sure if this change will reduce compilation time since I haven't measured the difference.
I wish this won't cause huge problem!

tiny_dnn/util/serialization_layer_list.h
```diff
@@ -1,25 +1,32 @@
...
-CNN_REGISTER_LAYER(softmax_layer, softmax);
-CNN_REGISTER_LAYER(leaky_relu_layer, leaky_relu);
-CNN_REGISTER_LAYER(elu_layer, elu);
-CNN_REGISTER_LAYER(tanh_p1m2_layer, tanh_scaled);
+namespace tiny_dnn {
 
-#endif
+template <typename T>
+void register_layers(T* h) {
+  h->template register_layer<average_pooling_layer>("avepool");
```
Randl: Is `template` keyword necessary here?

beru: Yes, it is necessary with gcc. Microsoft's VC++ doesn't need it though.
I guess the relationship is like Mozilla Firefox and Internet Explorer.

PR #675, opened 2017-04-29
Reduce use of C preprocessor

beru:
Hi, this PR reduces use of C preprocessor.
Since this library doesn't support Visual Studio 2013 anymore, getting rid of some preprocessor macros shouldn't be a problem.


edgarriba:
LGTM

PR #679, opened 2017-04-30
Replace `int` with `size_t` for `parallel_for`

Randl:


Randl:
https://github.com/tiny-dnn/tiny-dnn/pull/667#discussion_r114071294 related

tiny_dnn/util/parallel_for.h
```diff
@@ -100,8 +100,8 @@ void parallel_for(int begin, int end, const Func &f, int grainsize) {
 
   dispatch_apply(blockCount, dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0),
                  ^(size_t block) {
-                   int blockStart = static_cast<int>(block * blockSize);
-                   int blockEnd   = blockStart + blockSize;
+                   size_t blockStart = static_cast<int>(block * blockSize);
```
beru: Is `static_cast<int>` needed?

edgarriba: don't think so, `block` and `blocksize` are both `size_t`

beru:
I could only find this one.
https://github.com/tiny-dnn/tiny-dnn/blob/1ac9b1c84aac10daefdae3c0e3268eba38cc97a1/tiny_dnn/core/kernels/tiny_quantized_fully_connected_kernel.h#L244

But I've never tried quantized one and I'm not sure if it's still maintained...

Aside from that, I didn't see any problems. so LGTM.

beru:
The last commit ede173946bd7cf2389f4204cf0e1781866f65a96 somehow failed with AppVeyor Visual Studio 14 Win64 build `network.gradient_check` test.

```
[ RUN      ] network.gradient_check
c:\projects\tiny-dnn\test\test_network.h(460): error: Value of: nn.gradient_check<loss_func>( test_data.first, test_data.second, epsilon<float_t>(), GRAD_CHECK_RANDOM)
  Actual: false
Expected: true
[  FAILED  ] network.gradient_check (3 ms)
```

I'm not sure of the cause though...

Randl:
There are those strange random fails on AppVeyor, not sure if it's our or their bug.

Issue #641, opened 2017-04-02
Provide automatic shape inference for intermediate layers during network declaration.


karandesai-96:
As per the current API of tiny-dnn, every layer of network has to specify input shape of the tensor which shall come from its previous layer stacked in the network.

For example (current API at feat/decouple_activations branch):
```c++
nn << fully(784, 30) << relu(30)
   << fully (30, 10) << softmax (10);
```

We can provide a way to do automatic shape inference at intermediate layers, such that it is only mandatory to specify input shape at first layer of network.

The API shall become:
```c++
nn << fully(784, 30) << relu()
   << fully(10) << softmax();
```
First layer must specify input dim, but subsequent layers have to specify only the output dim.

This must be done in such a way that existing constructors work properly, both for backward compatibility as well as keeping in mind that any layer can be the first layer of a particular network.

Refer PR #639 for more information on shape inference. (Enabled properly for activation layers.)

beru:
#508 is related

Issue #553, opened 2017-02-09
Isolate serial_size_t inside serialization routines


beru:
Use of `serial_size_t` should be limited inside serialization routines.
Right now `serial_size_t` type is used in many files but it should be simply replaced by plain `size_t`.

edgarriba:
:+1: to include it into serialization routines

beru:
#484 is related.

beru:
http://uscilab.github.io/cereal/serialization_functions.html

Currently, external (some are still internal) serialize functions such as 
```cpp
  template <class Archive, typename Activation>
  static inline void serialize(
    Archive& ar, tiny_dnn::average_pooling_layer<Activation>& layer) {
    layer.serialize_prolog(ar);
    ar(cereal::make_nvp("in_size", layer.in_),
       cereal::make_nvp("pool_size_x", layer.pool_size_x_),
       cereal::make_nvp("pool_size_y", layer.pool_size_y_),
       cereal::make_nvp("stride_x", layer.stride_x_),
       cereal::make_nvp("stride_y", layer.stride_y_),
       cereal::make_nvp("pad_type", layer.pad_type_));
  }
```

is used and `average_pooling_layer` class's `stride_x_`, `stride_y_`, `pool_size_x_`, `pool_size_y_` members use `serial_size_t` type. 

External **split** load/save functions can be used to address the matter.
```cpp
  template <class Archive, typename Activation>
  static inline void save(
    Archive& ar,
    const tiny_dnn::average_pooling_layer<Activation> &layer) {
    layer.serialize_prolog(ar);
    // omit casts
    uint8_t pool_size_x = layer.pool_size_x_;
    uint8_t pool_size_y = layer.pool_size_y_;
    uint8_t stride_x = layer.stride_x_;
    uint8_t stride_y = layer.stride_y_;
    ar(cereal::make_nvp("in_size", layer.in_),
       cereal::make_nvp("pool_size_x", pool_size_x),
       cereal::make_nvp("pool_size_y", pool_size_y),
       cereal::make_nvp("stride_x", stride_x),
       cereal::make_nvp("stride_y", stride_y),
       cereal::make_nvp("pad_type", layer.pad_type_));
  }

  template <class Archive, typename Activation>
  static inline void load(
    Archive& ar,
    const tiny_dnn::average_pooling_layer<Activation> &layer) {
    layer.serialize_prolog(ar);
    // omit casts
    uint8_t pool_size_x;
[...]

PR #681, opened 2017-05-01
Add a new layer: Global Average Pooling

karandesai-96:
Global Average Pooling layer was introduced in the paper "Network in Network" (2013). It is an alternative to keeping fully connected layers at the end of a deep convolutional neural network.

Global Average Pooling layer has been used in recent architectures such as Inception Network, ResNet and Xception Network. It performs channel wise pooling and hence a layer with `c` channels will produce an output of `c` units, which is usually followed by softmax, giving classification scores.

#### Figure:
image

#### Content in this PR:

- [x] Add forward and backward op kernels for internal backend.
- [x] Add a class for global average pooling params.
- [x] Layer class implementation ( `class global_average_pooling_layer : public layer` ).
- [x] Relevant tests.
- [x] Serialization helpers and their tests.

#### References:

1. **"Network In Network"** by Min Lin, Qiang Chen, Shuicheng Yan, 2013. ( [arXiv:1312.4400v3 cs.NE](https://arxiv.org/abs/1312.4400v3) ) 
2. **"Going Deeper with Convolutions"** by Szegedy et. al., 2014. ( [arXiv:1409.4842 cs.CV](https://arxiv.org/abs/1409.4842) ).

### DIFF
--- a/test/test_global_average_pooling_layer.h
+++ b/test/test_global_average_pooling_layer.h
@@ -15,6 +15,16 @@
 
 namespace tiny_dnn {
 
+TEST(global_ave_pool, read_write) {
+  global_average_pooling_layer l1(100, 100, 5);
+  global_average_pooling_layer l2(100, 100, 5);
+
+  l1.init_weight();
+  l2.init_weight();
+
+  serialization_test(l1, l2);
+}
+
 TEST(global_ave_pool, forward) {
   global_average_pooling_layer l(4, 4, 1);
 
--- a/test/test_global_average_pooling_layer.h
+++ b/test/test_global_average_pooling_layer.h
@@ -1,0 +1,119 @@
+/*
+    Copyright (c) 2017, Taiga Nomi
+    All rights reserved.
+
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
+#pragma once
+
+#include <string>
+
+#include "gtest/gtest.h"
+#include "testhelper.h"
+#include "tiny_dnn/tiny_dnn.h"
+
+namespace tiny_dnn {
+
+TEST(global_ave_pool, forward) {
+  global_average_pooling_layer l(4, 4, 1);
+
+  // clang-format off
+  vec_t in = {0, 1, 2, 3,
+              8, 7, 5, 6,
+              4, 3, 1, 2,
+              0,-1,-2,-3};
+  // clang-format on
+
+  vec_t expected = {2.25};
+  vec_t res      = l.forward({{in}})[0][0];
+
+  EXPECT_FLOAT_EQ(expected[0], res[0]);
+}
+
+TEST(global_ave_pool, setup_internal) {
+  global_average_pooling_layer l(4, 4, 1, core::backend_t::internal);
+
+  EXPECT_EQ(l.parallelize(), true);          // if layer can be parallelized
+  EXPECT_EQ(l.in_channels(), 1u);            // num of input tensors
+  EXPECT_EQ(l.out_channels(), 1u);           // num of output tensors
+  EXPECT_EQ(l.in_data_size(), 16u);          // size of input tensors
+  EXPECT_EQ(l.out_data_size(), 1u);          // size of output tensors
+  EXPECT_EQ(l.in_data_shape().size(), 1u);   // num of inputs shapes
+  EXPECT_EQ(l.out_data_shape().size(), 1u);  // num of output shapes
+  EXPECT_EQ(l.inputs().size(), 1u);          // num of input edges
+  EXPECT_EQ(l.outputs().size(), 1u);         // num of output edges
+  EXPECT_EQ(l.in_types().size(), 1u);        // num of input data types
+  EXPECT_EQ(l.out_types().size(), 1u);       // num of output data types
+  EXPECT_EQ(l.fan_in_size(), 16u);           // num of incoming connections
+  EXPECT_EQ(l.fan_out_size(), 1u);           // num of outgoing connections
+  EXPECT_STREQ(l.layer_type().c_str(),
+               "global-ave-pool");  // string with layer type
+}
+
+TEST(global_ave_pool, forward_multichannel) {
+  global_average_pooling_layer l(2, 2, 3);
+  // clang-format off
+  vec_t in = {0, 1,
+              2, 3, 8, 7,
+                    5, 6, 4, 3,
+                          1, 2};
+  // clang-format on
+
+  vec_t expected = {1.5, 6.5, 2.5};
+
+  vec_t res = l.forward({{in}})[0][0];
+
+  for (size_t i = 0; i < expected.size(); i++) {
+    EXPECT_FLOAT_EQ(expected[i], res[i]);
+  }
+}
+
+TEST(global_ave_pool, backward) {
+  global_average_pooling_layer l(4, 4, 1);
+  // clang-format off
+  vec_t in = {0, 1, 2, 3,
+              8, 7, 5, 6,
+              4, 3, 1, 2,
+              0,-1,-2,-3};
+
+  vec_t out_grad = {16};
+
+  vec_t in_grad_expected = {1, 1, 1, 1,
+                            1, 1, 1, 1,
+                            1, 1, 1, 1,
+                            1, 1, 1, 1};
+  // clang-format on
+
+  l.forward({{in}})[0];
+  vec_t in_grad = l.backward(std::vector<tensor_t>{{out_grad}})[0][0];
+
+  for (size_t i = 0; i < in_grad.size(); i++) {
+    EXPECT_FLOAT_EQ(in_grad_expected[i], in_grad[i]);
+  }
+}
+
+TEST(global_ave_pool, backward_multichannel) {
+  global_average_pooling_layer l(2, 2, 3);
+  // clang-format off
+  vec_t in = {0, 1,
+              2, 3, 8, 7,
+                    5, 6, 4, 3,
+                          1, 2};
+
+  vec_t out_grad = {4, 8, 12};
+
+  vec_t in_grad_expected = {1, 1,
+                            1, 1, 2, 2,
+                                  2, 2, 3, 3,
+                                        3, 3};
+  // clang-format on
+
+  l.forward({{in}})[0];
+  vec_t in_grad = l.backward(std::vector<tensor_t>{{out_grad}})[0][0];
+
+  for (size_t i = 0; i < in_grad.size(); i++) {
+    EXPECT_FLOAT_EQ(in_grad_expected[i], in_grad[i]);
+  }
+}
+}  // namespace tiny_dnn
--- a/tiny_dnn/core/kernels/global_avepool_op.h
+++ b/tiny_dnn/core/kernels/global_avepool_op.h
@@ -1,0 +1,37 @@
+/*
+    Copyright (c) 2017, Taiga Nomi
+    All rights reserved.
+
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
+#pragma once
+
+#include "tiny_dnn/core/framework/op_kernel.h"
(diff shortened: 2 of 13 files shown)


test/test_global_average_pooling_layer.h
```diff
@@ -0,0 +1,122 @@
...
+  vec_t res      = l.forward({{in}})[0][0];
+
+  EXPECT_FLOAT_EQ(expected[0], res[0]);
+}
+
+TEST(global_ave_pool, setup_internal) {
+  global_average_pooling_layer l(4, 4, 1, core::backend_t::internal);
+
+  EXPECT_EQ(l.parallelize(), true);              // if layer can be parallelized
+  EXPECT_EQ(l.in_channels(), serial_size_t(1));  // num of input tensors
```
edgarriba: check https://github.com/tiny-dnn/tiny-dnn/pull/667/files#diff-cc023dbe1a164eea581096cc7945bf58R22

tiny_dnn/core/params/global_avepool_params.h
```diff
@@ -0,0 +1,25 @@
...
+*/
+#pragma once
+#include "tiny_dnn/core/params/params.h"
+
+namespace tiny_dnn {
+namespace core {
+
+class global_avepool_params : public Params {
+ public:
+  index3d<serial_size_t> in;
```
edgarriba: `size_t`

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -0,0 +1,130 @@
...
+  /* Forward and backward ops */
+  std::shared_ptr<core::OpKernel> kernel_fwd_;
+  std::shared_ptr<core::OpKernel> kernel_back_;
+
+  void init_backend(backend_t backend_type) {
+    core::OpKernelConstruction ctx =
+      core::OpKernelConstruction(layer::device(), &params_);
+
+    if (backend_type == backend_t::internal ||
+        backend_type == backend_t::nnpack || backend_type == backend_t::avx) {
```
edgarriba: just `backend_t::internal`

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -0,0 +1,130 @@
...
+                               backend_t backend_type = core::default_engine())
+    : global_average_pooling_layer(
+        in_shape.width_, in_shape.height_, in_shape.depth_, backend_type) {}
+
+  /**
+   * @param in_width     [in] width of input image
+   * @param in_height    [in] height of input image
+   * @param in_channels  [in] the number of input image channels (depth)
+  **/
+  global_average_pooling_layer(serial_size_t in_width,
```
edgarriba: I guess that `serial_size_t` is because of layer members, but we should start to think in fixing this

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -0,0 +1,130 @@
...
+
+  std::string layer_type() const override {
+    return std::string("global-ave-pool");
+  }
+
+  std::pair<serial_size_t, serial_size_t> pool_size() const {
+    return std::make_pair(params_.in.width_, params_.in.height_);
+  }
+
+#ifndef CNN_NO_SERIALIZATION
```
edgarriba: not needed anymore https://github.com/tiny-dnn/tiny-dnn/pull/667/files#diff-123f62ed8f57f99ec37c4e4dffe64146L40

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -0,0 +1,130 @@
...
+  global_average_pooling_layer(serial_size_t in_width,
+                               serial_size_t in_height,
+                               serial_size_t in_channels,
+                               backend_t backend_type = core::default_engine())
+    : layer({vector_type::data}, {vector_type::data}) {
+    set_global_avepool_params(shape3d(in_width, in_height, in_channels),
+                              shape3d(in_channels, 1, 1));
+
+    init_backend(backend_type);
+    layer::set_backend_type(backend_type);
```
edgarriba: it makes more sense to set the backend type inside the `init_backend`routine. Others layers will also need this -> next PR

edgarriba:
@karandesai-96 last comment, it's fine for now to implement the layer with Ops and backends strategy, but for future works, just implement basic forward and backward routines. Refactoring will come later. Just make sure before adding too many lines of code that we can have different backends.
@beru AVX for this?

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -0,0 +1,130 @@
...
+ * applies channel-wise global average pooling to spatial data.
+ **/
+class global_average_pooling_layer : public layer {
+ public:
+  using layer::parallelize_;
+
+  global_average_pooling_layer(const shape3d &in_shape,
+                               backend_t backend_type = core::default_engine())
+    : global_average_pooling_layer(
+        in_shape.width_, in_shape.height_, in_shape.depth_, backend_type) {}
```
Randl: Do we need separate `width` and `height`? We can generalize to any number of dimensions by passing just area and depth, without caring on number of dimensions. Or passing just depth and deducing area. Or even deducing all the parameters similarly to activations.

beru:
@edgarriba 
> @beru AVX for this?

Writing `vectorize::accumulate_2d` might be good.
It can also be used in internal backend.

tiny_dnn/layers/global_average_pooling_layer.h
karandesai-96: we need to fall back to internal backend else it will fail a few jobs on travis. similar thing is done for certain other layers.

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -109,7 +109,8 @@ class global_average_pooling_layer : public layer {
       core::OpKernelConstruction(layer::device(), &params_);
 
     layer::set_backend_type(backend_type);
-    if (backend_type == backend_t::internal || backend_type == backend_t::avx || backend_type == backend_t::nnpack) {
+    if (backend_type == backend_t::internal || backend_type == backend_t::avx ||
+        backend_type == backend_t::nnpack) {
```
edgarriba: really, I think it can lead to future bugs. To prevent this you can use macros at testing time
https://github.com/tiny-dnn/tiny-dnn/blob/master/test/test_convolutional_layer.h#L212

tiny_dnn/layers/global_average_pooling_layer.h
beru: #553 is related.

tiny_dnn/core/kernels/global_avepool_op.h
```diff
@@ -0,0 +1,37 @@
...
+    in the LICENSE file.
+*/
+#pragma once
+
+#include "tiny_dnn/core/framework/op_kernel.h"
+#include "tiny_dnn/core/kernels/global_avepool_op_internal.h"
+
+namespace tiny_dnn {
+
+class GlobalAvePoolOp : public core::OpKernel {
```
beru: This isn't particularly about Global Averaging Pool but,
@edgarriba Why `core::OpKernel` and its descendants are `CamelCase`?
Is it a bad thing or too late to change their names to `snake_case`?
I'm not really a fellow member of coding conventions police though.

tiny_dnn/layers/global_average_pooling_layer.h
beru: It makes me wonder why we have to always supply dimentional parameters when we construct layers... (I don't take internal implemetation into consideration right now)
#641 is related

The way I see it, this layer doesn't learn weights or sth during back propagation and works as a filter.

tiny_dnn/core/kernels/global_avepool_op_internal.h
```diff
@@ -0,0 +1,52 @@
...
+  const core::global_avepool_params &params,
+  const bool layer_parallelize) {
+  for_i(layer_parallelize, in_data.size(), [&](int sample) {
+    const vec_t &in = in_data[sample];
+    vec_t &out      = out_data[sample];
+
+    const size_t pool_area = params.in.width_ * params.in.height_;
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      for (size_t j = 0; j < pool_area; j++) {
+        out[i] += in[i * pool_area + j];
```
beru: You should use a local variable when accumulating values.

tiny_dnn/core/kernels/global_avepool_op_internal.h
```diff
@@ -0,0 +1,52 @@
...
+      }
+      out[i] /= pool_area;
```
beru: Division is very costly so doing that inside a loop wastes a fractional amount of precious electricity .
In many cases, you can replace divisions with reciprocal multiplications by sacrificing code readability.

karandesai-96: We had this readability vs micro optimization discussion before, but yes - this is a tight for loop. I can sacrifice readability.

bhack:
https://github.com/ridiculousfish/libdivide/blob/master/README.md

beru:
@bhack `livdivide` is for integer division. So I think it can't be used for floating-point values in place.
And `float / int` operation, divisor `int` type value is automatically converted to `float` type by C++ compiler (but if divisor is literal and it is equal to 2 to the power n, maybe optimized but that's not the case here).

I can be wrong though.

Aside from that, I've noticed that you tend to write only URL so I have to make a wild guess.
The size of my brain is like :chicken: so it'd be great if you could attach your comment when introducing URL.

tiny_dnn/core/kernels/global_avepool_op_internal.h
beru: There's such a dilemma. And probablly Global Averaging Pooling layer operation may not be the hot spot of NN computation. So it's possible that the performance gain from optimizing this layer isn't very rewardful.

So if we need to organize optimization work, writing profiling feature and benchmark tests (layerwise? opwise?) will probablly aid the work.

bhack:
Yes sorry I've pasted the wrong url from the bookmark. The correct one is this and doesn't require a comment. http://stackoverflow.com/questions/35506226/will-the-compiler-optimize-division-into-multiplication

beru:
@bhack Thank you for the correction.

I guess doing micro-optimization for CPU is like training how to walk faster. 
So in practical, we should rely on hardware accelerators.

tiny_dnn/core/kernels/global_avepool_op_internal.h
Randl: Have you checked generated assembly? I m 99% sure generated code has this optimisation

karandesai-96: Also, we need to check whether these things we care about so much, are already handled by our `-O31 optimization. I am not sure about specific details. http://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

beru: @Randl I've just confirmed `vdivss` instruction is used with `-mavx -O3` compile options.

Please read the page.
http://stackoverflow.com/a/35506382
https://github.com/tiny-dnn/tiny-dnn/pull/564#discussion_r100667964

tiny_dnn/core/kernels/global_avepool_op.h
edgarriba: at some point I had some class naming conflicts that I could solve by using `CamelCase` then we came with this convention stuff

tiny_dnn/core/kernels/global_avepool_op_internal.h
Randl: I see. Then it might make sense to optimize it into single division. @karandesai-96 Do you care to run a simple benchmark?

karandesai-96:
Right so we're relying on hardware accelerators alone. And about @edgarriba's comment - I am adding AVX implementation in next PR.

PR #687, opened 2017-05-06
Add AVX implementation for Global Average Pooling layer.

karandesai-96:
Since global average pooling layer calculates average of all activations per channel, we pick up contigious 8 floats and keep on performing vertical sum channelwise. At the end a net sum is accumulated by horizontal sum. This is repeated for all channels of a layer.

Current code falls back to internal backend if nnpack or other unsupported backend is chosen.


karandesai-96:
@beru I think we should decide specific naming conventions for these vector blocks. Just for the sake of consistency. Right now to avoid confusion, I preferred suffixing such variables with a `_m`.

beru:
@karandesai-96 I say NAY to Systems Hungarian notation.

tiny_dnn/core/kernels/avx_kernel_common.h
```diff
@@ -56,6 +56,17 @@ inline __m128d madd128_sd(__m128d a, __m128d b, __m128d c) {
...
+// x = ( x3, x2, x1, x0 )
+// use _mm_cvtss_f32 if you need a float result instead of a __m128
+inline __m128 hsum128_ps(__m128 x) {
+  // twoSum = ( x3+x2, x1+x0, x3+x2, x1+x0 )
+  const __m128 twoSum = _mm_hadd_ps(x, x);
+  // fourSum = ( x3+x2+x1+x0, x3+x2+x1+x0, x3+x2+x1+x0, x3+x2+x1+x0 )
+  const __m128 fourSum = _mm_hadd_ps(twoSum, twoSum);
+  return fourSum;
+}
+
```
beru: Please check this page. http://stackoverflow.com/questions/6996764/fastest-way-to-do-horizontal-float-vector-sum-on-x86

It looks like this is faster.
```cpp
// in  : ( x3, x2, x1, x0 )
// out : (  -,  -,  -, x3+x2+x1+x0 )
inline __m128 hsum128_ps(__m128 x)
{
  // loDual = ( -, -, x1, x0 )
  const __m128 loDual = x;
  // hiDual = ( -, -, x3, x2 )
  const __m128 hiDual = _mm_movehl_ps(x, x);
  // sumDual = ( -, -, x1+x3, x0+x2 )
  const __m128 sumDual = _mm_add_ps(loDual, hiDual);
  // lo = ( -, -, -, x0+x2 )
  const __m128 lo = sumDual;
  // hi = ( -, -, -, x1+x3 )
  const __m128 hi = _mm_shuffle_ps(sumDual, sumDual, 0x1);
  // sum = ( -, -, -, x0+x1+x2+x3 )
  const __m128 sum = _mm_add_ss(lo, hi);
  return sum;
}
```

tiny_dnn/core/kernels/global_avepool_grad_op.h
```diff
@@ -27,10 +28,28 @@ class GlobalAvePoolGradOp : public core::OpKernel {
...
+                                          context.parallelize());
+#endif
+    } else if (engine == core::backend_t::nnpack) {
+      // fallback to internal implementation as nnpack implementation
+      // is not available
+      kernels::global_avepool_grad_op_internal(prev_delta, curr_delta, params,
+                                               context.parallelize());
+    } else {
+      throw nn_error("Not supported engine: " + to_string(engine));
+    }
```
beru: I don't think all backends should ready its own versions of every kernels as it's highly likely impossible.
so I would simply write fallback routine in this way.
```cpp
    if (engine == core::backend_t::avx) {
#ifdef CNN_USE_DOUBLE
      // todo (kd): add avx implementation for CNN_USE_DOUBLE
      kernels::global_avepool_grad_op_internal(prev_delta, curr_delta, params,
                                               context.parallelize());
#else
      kernels::global_avepool_grad_op_avx(prev_delta, curr_delta, params,
                                          context.parallelize());
#endif
    } else {
      kernels::global_avepool_grad_op_internal(prev_delta, curr_delta, params,
                                               context.parallelize());
    }
```

tiny_dnn/core/kernels/global_avepool_op.h
```diff
@@ -27,10 +28,28 @@ class GlobalAvePoolOp : public core::OpKernel {
...
+                                     context.parallelize());
+#endif
+    } else if (engine == core::backend_t::nnpack) {
+      // graceful fallback to internal implementation as nnpack
+      // implementation is not available
+      kernels::global_avepool_op_internal(in_data, out_data, params,
+                                          context.parallelize());
+    } else {
+      throw nn_error("Not supported engine: " + to_string(engine));
+    }
```
beru: I would write fallback routine in this way.
```cpp
    if (engine == core::backend_t::avx) {
#ifdef CNN_USE_DOUBLE
      // todo (kd): add avx implementation for CNN_USE_DOUBLE
      kernels::global_avepool_op_internal(in_data, out_data, params,
                                          context.parallelize());
#else
      kernels::global_avepool_op_avx(in_data, out_data, params,
                                     context.parallelize());
#endif
    } else {
      kernels::global_avepool_op_internal(in_data, out_data, params,
                                          context.parallelize());
    }
```

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -109,8 +109,9 @@ class global_average_pooling_layer : public layer {
       core::OpKernelConstruction(layer::device(), &params_);
 
     layer::set_backend_type(backend_type);
-    if (backend_type == backend_t::internal || backend_type == backend_t::avx ||
-        backend_type == backend_t::nnpack) {
+    if (backend_type == core::backend_t::internal ||
+        backend_type == core::backend_t::avx ||
+        backend_type == core::backend_t::nnpack) {
```
beru: What if other backends will appear in future?
Isn't it inconvenient that implementors have to update this code?

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,83 @@
...
+        };
+        __m256i imask = _mm256_loadu_si256(
+          (__m256i const *)(mask_src + 8 - nremains_per_channel));
+        __m256 in_m = _mm256_setzero_ps();
+        in_m =
+          _mm256_maskload_ps(&in[depth_index + 8 * nblocks_per_channel], imask);
+        sum_m = _mm256_add_ps(sum_m, in_m);
+      }
+      out[i] = _mm_cvtss_f32(hsum256_ps(sum_m));
+      out[i] /= pool_area;
```
beru: It is advised to use reciprocal multiplication instead of division.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,83 @@
...
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256 in_m = _mm256_load_ps(&in[depth_index + 8 * j]);
+        sum_m       = _mm256_add_ps(sum_m, in_m);
+      }
+      if (nremains_per_channel) {
+        int32_t mask_src[] = {
+          -1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0, 0,
+        };
+        __m256i imask = _mm256_loadu_si256(
+          (__m256i const *)(mask_src + 8 - nremains_per_channel));
```
beru: You should make this mask variable outside of the loop.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,83 @@
...
+  for_i(layer_parallelize, prev_delta.size(), [&](int sample) {
+    vec_t &prev       = prev_delta[sample];
+    const vec_t &curr = curr_delta[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      const float_t pi = curr[i] / pool_area;
+      size_t j         = 0;
+
+      while (j < nblocks_per_channel) {
+        __m256 prev0 = _mm256_set1_ps(pi);
```
beru: What's the point of making this variable inside the inner loop even though the value is always the same.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,83 @@
...
+  const bool layer_parallelize) {
+  const size_t pool_area           = params.in.width_ * params.in.height_;
+  const size_t nblocks_per_channel = pool_area / 8;
+
+  for_i(layer_parallelize, prev_delta.size(), [&](int sample) {
+    vec_t &prev       = prev_delta[sample];
+    const vec_t &curr = curr_delta[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      const float_t pi = curr[i] / pool_area;
```
beru: It is advised to replace division with reciprocal multiplication.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,83 @@
...
+
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
+#pragma once
+
+namespace tiny_dnn {
+namespace kernels {
+
+#ifdef CNN_USE_AVX
```
beru: `global_avepool_op_avx` and `global_avepool_grad_op_avx` functions are referenced even when `CNN_USE_AVX` isn't defined. So if you reference them, you should keep them.

In `tiny_dnn/core/kernels/conv2d_op_avx.h` file, I divided interface function `conv2d_op_avx` and implementation function `avx_conv2d_5x5_kernel`.

tiny_dnn/layers/global_average_pooling_layer.h
karandesai-96: This is done in all the layers right now. I couldn't find a better alternative hence went for consistency.

tiny_dnn/core/kernels/avx_kernel_common.h
karandesai-96: It was used earlier but I don't need it anymore.

tiny_dnn/layers/global_average_pooling_layer.h
beru: My understanding is that we can always fallback to internal backend when selected backend doesn't support layer's operation.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,120 @@
...
+
+inline void avx_global_avepool_forward_kernel(
+  const tensor_t &in_data,
+  tensor_t &out_data,
+  const core::global_avepool_params &params,
+  const bool layer_parallelize) {
+  const size_t pool_area            = params.in.width_ * params.in.height_;
+  const size_t pool_area_inv        = 1.0f / pool_area;
```
beru: It's impossible for `size_t` type variable to retain floating point value.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,120 @@
...
+                                  const bool layer_parallelize) {
+#ifdef CNN_USE_AVX
+  avx_global_avepool_forward_kernel(in_data, out_data, params,
+                                    layer_parallelize);
+#else
+  CNN_UNREFERENCED_PARAMETER(in_data);
+  CNN_UNREFERENCED_PARAMETER(out_data);
+  CNN_UNREFERENCED_PARAMETER(params);
+  CNN_UNREFERENCED_PARAMETER(layer_parallelize);
+  throw nn_error("TinyDNN has not been compiled with AVX support.");
```
beru: I'd write fallback call to internal backend routine `global_avepool_op_internal` instead of throwing an exceotion.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,120 @@
...
+  const bool layer_parallelize) {
+#ifdef CNN_USE_AVX
+  avx_global_avepool_backward_kernel(prev_delta, curr_delta, params,
+                                     layer_parallelize);
+#else
+  CNN_UNREFERENCED_PARAMETER(prev_delta);
+  CNN_UNREFERENCED_PARAMETER(curr_delta);
+  CNN_UNREFERENCED_PARAMETER(params);
+  CNN_UNREFERENCED_PARAMETER(layer_parallelize);
+  throw nn_error("TinyDNN has not been compiled with AVX support.");
```
beru: I'd write fallback call to internal backend routine `global_avepool_grad_op_internal` instead of throwing an exceotion.

In that way, you can remove backend type check in `global_average_pooling_layer::init_backend` method in `tiny_dnn/layers/global_average_pooling_layer.h` file.

tiny_dnn/core/kernels/global_avepool_op_avx.h
Randl: IIRC that's what's done now in all layers (exception). Probably fallback to `internal` backend with warning is better solution

beru: Hmmm @Randl and @karandesai-96  are right, I didn't check other layers implementation.
Giving warning at compile time with `#pragma message`?

tiny_dnn/core/kernels/global_avepool_op_avx.h
karandesai-96: I rectified this one - similar to what is done in conv2d and fully connected kernels.

tiny_dnn/core/kernels/global_avepool_op_avx.h
Randl: Yeah, that's what I thought too

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,120 @@
...
+
+#ifdef CNN_USE_AVX
+
+inline void avx_global_avepool_forward_kernel(
+  const tensor_t &in_data,
+  tensor_t &out_data,
+  const core::global_avepool_params &params,
+  const bool layer_parallelize) {
+  const size_t pool_area            = params.in.width_ * params.in.height_;
+  const float_t pool_area_inv       = 1.0f / pool_area;
```
beru: You don't need to use `float_t` in here. Just `float` is OK. And don't forget to cast `pool_area` to `float` with `static_cast`, otherwise I think compiler gives a warning.

I'd use `_mm_set_ss` to hold `pool_area` value in `__m128` typed variable and use `_mm_rcp_ss` to compute a reciprocal. In that way, you can multiply it with the result of `hsum256_ps` with `_mm_mul_ss`.

beru: You can use `_mm_cvtsi32_ss` intrinsic function to set single dword(32bit) variable to first slot in `__m128`.

tiny_dnn/core/kernels/global_avepool_grad_op.h
```diff
@@ -27,10 +28,17 @@ class GlobalAvePoolGradOp : public core::OpKernel {
...
-    kernels::global_avepool_grad_op_internal(prev_delta, curr_delta, params,
-                                             context.parallelize());
+    const core::backend_t engine = context.engine();
+
+    if (engine == core::backend_t::avx) {
+      kernels::global_avepool_grad_op_avx(prev_delta, curr_delta, params,
+                                          context.parallelize());
+    } else {
+      // fallback to internal implementation as nnpack implementation
+      // is not available
```
beru: The above comment is outdated.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,196 @@
...
+  const size_t nblocks_per_channel  = pool_area / 8;
+  const size_t nremains_per_channel = pool_area & 7;
+
+  int32_t mask_src[] = {
+    -1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0, 0,
+  };
+  __m256i imask =
+    _mm256_loadu_si256((__m256i const *)(mask_src + 8 - nremains_per_channel));
+
+  for_i(layer_parallelize, in_data.size(), [&](int sample) {
```
beru: Please use `size_t sample` instead of `int sample` for_i functor lambda's argument.
#679 is related.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,196 @@
...
+    const vec_t &in = in_data[sample];
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256 sum_m             = _mm256_setzero_ps();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256 in_m = _mm256_load_ps(&in[depth_index + 8 * j]);
+        sum_m       = _mm256_add_ps(sum_m, in_m);
```
beru: You should unroll this loop 8x times.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,196 @@
...
+  for_i(layer_parallelize, in_data.size(), [&](int sample) {
+    const vec_t &in = in_data[sample];
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256d in_m = _mm256_load_pd(&in[depth_index + 4 * j]);
+        sum_m        = _mm256_add_pd(sum_m, in_m);
```
beru: You should unroll this loop 8x times.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,196 @@
...
+  const size_t pool_area           = params.in.width_ * params.in.height_;
+  const double pool_area_inv       = 1.0f / static_cast<double>(pool_area);
+  const size_t nblocks_per_channel = pool_area / 4;
+
+  for_i(layer_parallelize, prev_delta.size(), [&](int sample) {
+    vec_t &prev       = prev_delta[sample];
+    const vec_t &curr = curr_delta[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      const double pi = curr[i] * pool_area_inv;
```
beru: You should rewrite this with AVX intrinsics functions.

tiny_dnn/core/kernels/global_avepool_op_avx.h
karandesai-96: I didn't understand this, what do you mean by unrolling 8x times? Aren't we doing vertical additions till the end then followed by horizontal addition?

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,196 @@
...
+
+  for_i(layer_parallelize, in_data.size(), [&](int sample) {
+    const vec_t &in = in_data[sample];
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256d in_m = _mm256_load_pd(&in[depth_index + 4 * j]);
```
beru: Are you sure the address `&in[depth_index + 4 * j]` is always aligned to 32 bytes?
If not, you should use `_mm256_loadu_pd`.

tiny_dnn/core/kernels/global_avepool_op_avx.h
beru: It's related with instruction latencies and throughputs, Instruction-level parallelism, 16 YMM registers etc...
The Haswell microarchitecture, The Skylake microarchitecture, they have 8 execution ports. etc, etc..

karandesai-96:
@beru I have worked with the review comments. Please have a look once more.
@edgarriba some patterns are different than other layers - such as fallback, macro protection etc. if this works well, we can make it uniform throughout.

edgarriba:
@karandesai-96 sounds good

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256d in_m = _mm256_loadu_pd(&in[depth_index + 4 * j]);
+        sum_m        = _mm256_add_pd(sum_m, in_m);
+      }
+      if (nremains_per_channel) {
+        __m256d in_m = _mm256_setzero_pd();
```
beru: You don't need to initialize `in_m` with `_mm256_setzero_pd` intrinsinc function.
It seems like `vmaskmovpd` instruction does the job.
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm256_maskload_pd&expand=3279

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256d in_m = _mm256_loadu_pd(&in[depth_index + 4 * j]);
+        sum_m        = _mm256_add_pd(sum_m, in_m);
+      }
+      if (nremains_per_channel) {
```
beru: I suppose you don't need to use `if` statement.
If `nremains_per_channel` is `0`, `imask` will be set to all `0`, so actual load won't occur with `_mm256_maskload_pd`.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+  for_i(layer_parallelize, in_data.size(), [&](size_t sample) {
+    const vec_t &in = in_data[sample];
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256d in_m = _mm256_loadu_pd(&in[depth_index + 4 * j]);
+        sum_m        = _mm256_add_pd(sum_m, in_m);
```
beru: Loop unrolling is one of the key techniques for SIMD optimization.
http://stackoverflow.com/questions/21090873/loop-unrolling-to-achieve-maximum-throughput-with-ivy-bridge-and-haswell
If you think it's tedious to do so, I'll do that for you.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      size_t j       = 0;
+      __m256d prev_m = _mm256_set1_pd(curr[i] * pool_area_inv);
+
+      while (j < nblocks_per_channel) {
+        _mm256_storeu_pd(&prev[i * pool_area + 4 * j], prev_m);
+        j++;
+      }
+
+      if (nremains_per_channel) {
```
beru: You don't need to use this `if` statement.
If `imask` is set correctly, `_mm256_maskstore_pd` will do the job in all 4 cases.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// double version
+inline void global_avepool_grad_op_avx(
+  tensor_t &prev_delta,
+  const tensor_t &curr_delta,
+  const core::global_avepool_params &params,
+  const bool layer_parallelize) {
+  const size_t pool_area            = params.in.width_ * params.in.height_;
+  const double pool_area_inv        = 1.0f / static_cast<double>(pool_area);
```
beru: This is double path, you should write in this way.
`const double pool_area_inv        = 1.0 / pool_area;`

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+  __m256i imask =
+    _mm256_loadu_si256((__m256i const *)(mask_src + 4 - nremains_per_channel));
+
+  for_i(layer_parallelize, prev_delta.size(), [&](size_t sample) {
+    vec_t &prev       = prev_delta[sample];
+    const vec_t &curr = curr_delta[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      size_t j       = 0;
+      __m256d prev_m = _mm256_set1_pd(curr[i] * pool_area_inv);
```
beru: You can write in this way.
```cpp
__m256d prev_m = _mm256_broadcast_sd(&curr[i]);
prev_m = _mm256_mul_pd(prev_m, pool_area_inv_m);
```

And you can ready `pool_area_inv_m` outside of the loop with sth like
```cpp
__m256d pool_area_inv_m = _mm256_set1_pd(pool_area_inv);
```

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+
+      while (j < nblocks_per_channel) {
+        _mm256_storeu_pd(&prev[i * pool_area + 4 * j], prev_m);
+        j++;
```
beru: You should unroll this loop to achieve further optimization.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256 sum_m             = _mm256_setzero_ps();
+      const size_t depth_index = i * pool_area;
+      for (size_t j = 0; j < nblocks_per_channel; j++) {
+        __m256 in_m = _mm256_load_ps(&in[depth_index + 8 * j]);
+        sum_m       = _mm256_add_ps(sum_m, in_m);
+      }
+      if (nremains_per_channel) {
```
beru: You don't need to use this `if` statement since you use `_mm256_maskload_ps` function later.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+        __m256 in_m = _mm256_setzero_ps();
```
beru: You don't need to initialize `in_m` with `_mm256_setzero_ps` intrinsic function.
`_mm256_maskload_ps` should do the job properly.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+        in_m =
+          _mm256_maskload_ps(&in[depth_index + 8 * nblocks_per_channel], imask);
+        sum_m = _mm256_add_ps(sum_m, in_m);
+      }
+      out[i] = _mm_cvtss_f32(hsum256_ps(sum_m));
+      out[i] *= pool_area_inv;
```
beru: Above 2 lines are not efficient. If you write in this way, redundant `store` and `load` operations will happen.
Can't you write in this way?
```cpp
_mm_store_ss(&out[i], _mm_mul_ss(hsum256_ps(sum_m), pool_area_inv_m));
```

And you need to ready `pool_area_inv_m` variable with
```cpp
__m128 pool_area_inv_m = _mm_set1_ps(1.0f / static_cast<float>(pool_area));
```
outside of the loop.

beru:
@karandesai-96 
> @beru I have worked with the review comments. Please have a look once more.

I checked the code but you still haven't written it in proper ways.
Though I'm unsure how further we should optimize this implementation.

Did you write sample code that uses  global average pooling layer?
How do you measure the performance of this AVX implementation for Global Average Pooling layer?

tiny_dnn/layers/global_average_pooling_layer.h
```diff
@@ -109,14 +109,17 @@ class global_average_pooling_layer : public layer {
...
-      return;
-    } else {
-      throw nn_error("Not supported engine: " + to_string(backend_type));
+    if (backend_type == core::backend_t::avx) {
+#ifndef CNN_USE_AVX
+      nn_warn(
+        "tiny-dnn has not been compiled with AVX support, "
+        "fallback to internal backend for global avepool layer.\n");
+      layer::set_backend_type(core::backend_t::internal);
+#endif
```
beru: Is `nn_warn` really necessary? Is silently redirecting to internal backend implemenation is a bad thing?

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,186 @@
...
+      }
+    }
+  });
+}
+
+#endif  // CNN_USE_DOUBLE
+#else   // CNN_USE_AVX
+
+// due to fallback to internal backend, these functions will never be called.
+// empty declarations are just to avoid compilation errors while USE_AVX=OFF
```
beru: Instead of adding empty declarations, you can use preprocessor macros to disable the calls to `global_avepool_op_avx` and `global_avepool_grad_op_avx` when `CNN_USE_AVX` isn't defined.

Calling internal backend functions should be just fine.

tiny_dnn/layers/global_average_pooling_layer.h
karandesai-96: @beru: Yes. This warning will be raised only when the user specified `backend_t::avx` in layer constructor but didn't built the library with AVX support. It seems proper to warn the user that his/her specified argument did not work as it should have.

karandesai-96:
@beru: Thanks for the extensive reviews. It helped me thinking about these kinds of optimizations beyond the horizon. It will certainly help me for next turns. I will proceed with loop unrolling now. 

> Did you write sample code that uses global average pooling layer?
> How do you measure the performance of this AVX implementation for Global Average Pooling layer?

I wrote a small benchmark myself, based on the code snippet I wrote in my [GSoC Proposal (Page 6)](https://docs.google.com/document/d/18RtcDatYzZ1gpztFUM8K2MUaYfIAwvVFIJKCR-xf6BE/edit).

I will share the execution timings for internal backend, AVX without unrolling and AVX with unrolling soon.

tiny_dnn/core/kernels/global_avepool_op_avx.h
karandesai-96: @beru: What if `nblocks_per_channel` is less than 8 ? 

> If you think it's tedious to do so, I'll do that for you.
I'm fine with that as well, because this Ivy Bridge, Haswell etc. is pretty much going over my head :sweat_smile:

tiny_dnn/layers/global_average_pooling_layer.h
beru: Can't we kill `backend_t::avx` in `tiny_dnn/core/backend.h`?
What's the point of increasing erroneous paths?

tiny_dnn/core/kernels/global_avepool_op_avx.h
beru: @karandesai-96 
> What if nblocks_per_channel is less than 8 ?

You can process remaining data with not unrolled code.
Having 4x and 2x loop unrolling is also fine.

I recommend you to add `vectorize::accumulate` in `tiny_dnn/util/product.h` file.
In that way, reusing optimized code would be easy.

beru:
@karandesai-96 
> I wrote a small benchmark myself, based on the code snippet I wrote in my GSoC Proposal (Page 6). I will share the execution timings for internal backend, AVX without unrolling and AVX with unrolling soon.

If you include the benchmark program to this PR, other people can also run it and check the performance.

karandesai-96:
@beru Performed loop unrolling and added simple benchmarking code.
Here are the results:
image

Also I cleaned up and rearranged my commit history to hope that this PR won't get squashed. xD

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -35,8 +35,31 @@ inline void global_avepool_op_avx(const tensor_t &in_data,
       __m256d sum_m            = _mm256_setzero_pd();
       const size_t depth_index = i * pool_area;
       for (size_t j = 0; j < nblocks_per_channel; j++) {
-        __m256d in_m = _mm256_loadu_pd(&in[depth_index + 4 * j]);
-        sum_m        = _mm256_add_pd(sum_m, in_m);
+        if (j + 8 < nblocks_per_channel) {
```
beru: You shouldn't use `if` inside loop statements.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -35,8 +35,31 @@ inline void global_avepool_op_avx(const tensor_t &in_data,
...
+          __m256d in_m0 = _mm256_loadu_pd(&in[depth_index + 4 * j]);
+          __m256d in_m1 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 1)]);
+          __m256d in_m2 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 2)]);
+          __m256d in_m3 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 3)]);
+          __m256d in_m4 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 4)]);
+          __m256d in_m5 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 5)]);
+          __m256d in_m6 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 6)]);
+          __m256d in_m7 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 7)]);
+          sum_m         = _mm256_add_pd(
+            sum_m, vectorize::accumulate(in_m0, in_m1, in_m2, in_m3, in_m4,
```
beru: You should perform accumulation outside of the loop.

tiny_dnn/util/product.h
```diff
@@ -516,6 +516,68 @@ void fill(T *dst, size_t size, T value) {
 
 }  // namespace detail
 
+#ifdef CNN_USE_AVX
+#ifdef CNN_USE_DOUBLE
+
+// vertically accumulate eight AVX registers into single register.
+// registers hold double precision values.
+inline __m256d accumulate(__m256d a,
```
beru: Unfortunately, we have to use `const __m256d &` for VC++ 32bit build.

tiny_dnn/util/product.h
```diff
@@ -516,6 +516,68 @@ void fill(T *dst, size_t size, T value) {
...
+  const __m256d ab = _mm256_add_pd(a, b);
+  const __m256d cd = _mm256_add_pd(c, d);
+  return _mm256_add_pd(ab, cd);
+}
+
+#else  // CNN_USE_DOUBLE
+
+// vertically accumulate eight AVX registers into single register.
+// registers hold single precision values.
+inline __m256 accumulate(__m256 a,
```
beru: We have to use `const __m256 &` argument type for VC++ 32bit build.

tiny_dnn/util/product.h
```diff
@@ -516,6 +516,68 @@ void fill(T *dst, size_t size, T value) {
 
 }  // namespace detail
 
+#ifdef CNN_USE_AVX
+#ifdef CNN_USE_DOUBLE
```
beru: Please check other implementation in `tiny_dnn/util/product.h` file.
There are several classes.

```cpp
detail::double_avx
detail::float_avx
detail::double_sse
detail::float_sse
detail::scalar_generic<double>
detail::scalar_generic<float>
```

With ugly template metaprogramming, we can reduce code duplications.

karandesai-96: There are no usages of these classes project wide according to IntelliJ CLion so I assumed that it was some legacy code. I'll modify accordingly.

beru: > There are no usages of these classes project wide according to IntelliJ CLion so I assumed that it was some legacy code. I'll modify accordingly.

It's my fault. There's a lot of room for improvement with `tiny_dnn/util/product.h`.

tiny_dnn/core/kernels/global_avepool_op_avx.h
karandesai-96: How would that exactly add up ?

karandesai-96:
@beru working on your suggestions to use `VECTORIZE_TYPE::add` slows down a bit due to calling overhead. (accumulate -> add -> _mm256_add)
```plaintext
// before:
------------------------------------------------------------------------------------------
Benchmark                                                   Time           CPU Iterations
------------------------------------------------------------------------------------------
bm_global_avepool_forward_internal/iterations:1000    5712979 ns      41563 ns       1000
bm_global_avepool_forward_avx/iterations:1000          716491 ns      44371 ns       1000

// after:
------------------------------------------------------------------------------------------
Benchmark                                                   Time           CPU Iterations
------------------------------------------------------------------------------------------
bm_global_avepool_forward_internal/iterations:1000    5190217 ns      52399 ns       1000
bm_global_avepool_forward_avx/iterations:1000         1056757 ns      45231 ns       1000

```

karandesai-96:
On the other hand, removing excess if conditions from for loops and breaking the for loop into three sequential for loops speeds up a bit:
```plaintext
------------------------------------------------------------------------------------------
Benchmark                                                   Time           CPU Iterations
------------------------------------------------------------------------------------------
bm_global_avepool_forward_internal/iterations:1000    5071432 ns      46034 ns       1000
bm_global_avepool_forward_avx/iterations:1000          646088 ns      46132 ns       1000
```

karandesai-96:
Instead of template metaprogramming, I suggest using macros. I tried this out:

```c++
#if defined(CNN_USE_AVX)
#ifdef CNN_USE_DOUBLE
#define __M256 __m256d
#define _MM256_ADD _mm256_add_pd

#else

#define __M256 __m256
#define _MM256_ADD _mm256_add_ps
#endif
```

This won't create an extra function call overhead like that suggested through template metaprogramming. And macros help to reduce the code as well as we achieve a speedup as compared to earlier case of 780k - 800k nanoseconds timings.

Results:
```plaintext
------------------------------------------------------------------------------------------
Benchmark                                                   Time           CPU Iterations
------------------------------------------------------------------------------------------
bm_global_avepool_forward_internal/iterations:1000    5136436 ns      48891 ns       1000
bm_global_avepool_forward_avx/iterations:1000          644082 ns      46916 ns       1000
```

Randl:
I'm not sure that would work project-wide, but it might be convenient, unless we are forced to define some 1000 LoC of macros :)

karandesai-96:
@Randl I think we shall require only certain commonly used ones, besides only one set of them would exist at a time due to `CNN_USE_AVX` / `CNN_USE_DOUBLE` etc. protections.

beru:
> Instead of template metaprogramming, I suggest using macros.

I don't think preprocessor macros should be used often (though I use it when I have to)
There is `__forceinline` keyword for VC++, and `__attribute__((always_inline))` for GCC.
So if using these will remove extra function calls, we should stick with TMP.

```cpp
#ifdef _MSC_VER
#define MUSTINLINE __forceinline
#else
#define MUSTINLINE __attribute__((always_inline)) inline
#endif
```

I used `VECTORIZE_TYPE` macro in `tiny_dnn/util/product.h` since my C++ TMP skills are not good.

> I think we shall require only certain commonly used ones, besides only one set of them would exist at a time due to CNN_USE_AVX / CNN_USE_DOUBLE etc. protections.

I'd like this library to support multiple numeric data types in future. But using macros in that way limits the possibilties.

beru:
@karandesai-96 

For `tiny_dnn_benchmarks` VC++ project, we need this include directory setting. `..\benchmark-src\include\`

By adding it, I could successfully build the project.

beru:
Here's my attempt.

```cpp
// float version
inline void global_avepool_op_avx(const tensor_t &in_data,
                                  tensor_t &out_data,
                                  const core::global_avepool_params &params,
  const size_t pool_area            = params.in.width_ * params.in.height_;
  const size_t nblocks_per_channel  = pool_area / 8;
  const size_t nremains_per_channel = pool_area & 7;
  const size_t n4 = nblocks_per_channel / 4;
  const size_t n2 = (nblocks_per_channel & 3) / 2;
  const size_t n1 = nblocks_per_channel & 1;

  static const int32_t mask_src[] = {
    -1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0, 0,
  };
  __m256i imask =
    _mm256_loadu_si256((__m256i const *)(mask_src + 8 - nremains_per_channel));
  __m128 one_m = _mm_cvtsi32_ss(_mm_setzero_ps(), 1);
  __m128 pool_area_m = _mm_cvtsi32_ss(_mm_setzero_ps(), pool_area);
  __m128 pool_area_inv_m = _mm_div_ss(one_m, pool_area_m);

  for_(layer_parallelize, 0, in_data.size(), [&](const blocked_range &r) {
    for (size_t sample = r.begin(); sample < r.end(); ++sample) {
      const vec_t &in = in_data[sample];
      vec_t &out      = out_data[sample];
      const float *pin0 = &in[0];
      for (size_t i = 0; i < params.in.depth_; i++) {
        const float *pin = pin0;
        __m256 v0 = _mm256_loadu_ps(pin + 000);
        __m256 v1 = _mm256_loadu_ps(pin + 010);
        __m256 v2 = _mm256_loadu_ps(pin + 020);
        __m256 v3 = _mm256_loadu_ps(pin + 030);
        __m256 sum0 = _mm256_setzero_ps();
        __m256 sum1 = _mm256_setzero_ps();
        __m256 sum2 = _mm256_setzero_ps();
        __m256 sum3 = _mm256_setzero_ps();
        for (size_t j = 0; j < n4; ++j) {
          __m256 n0 = _mm256_loadu_ps(pin + 040);
          __m256 n1 = _mm256_loadu_ps(pin + 050);
          __m256 n2 = _mm256_loadu_ps(pin + 060);
          __m256 n3 = _mm256_loadu_ps(pin + 070);
          sum0 = _mm256_add_ps(sum0, v0);
          sum1 = _mm256_add_ps(sum1, v1);
          sum2 = _mm256_add_ps(sum2, v2);
          sum3 = _mm256_add_ps(sum3, v3);
          v0 = n0;
          v1 = n1;
          v2 = n2;
          v3 = n3;
          pin += 040;
        }
        if (n2) {
          sum0 = _mm256_add_ps(sum0, v0);
          sum1 = _mm256_add_ps(sum1, v1);
          pin += 020;
        }
        sum0 = _mm256_add_ps(sum0, sum1);
        if (n1) {
          sum2 = _mm256_add_ps(sum2, _mm256_loadu_ps(pin + 000));
          pin += 010;
        }
        sum3 = _mm256_add_ps(sum3, _mm256_maskload_ps(pin, imask));
        sum2 = _mm256_add_ps(sum2, sum3);
        sum0 = _mm256_add_ps(sum0, sum2);
        _mm_store_ss(&out[i], _mm_mul_ss(hsum256_ps(sum0), pool_area_inv_m));
        pin0 += pool_area;
      }
    }
  });
}
```

```
Run on (4 X 3400 MHz CPU s)
05/20/17 22:40:52
-------------------------------------------------------------------------------------------
Benchmark                                                    Time           CPU Iterations
-------------------------------------------------------------------------------------------
bm_global_avepool_forward_internal/iterations:10000     812486 ns       9375 ns      10000
bm_global_avepool_forward_avx/iterations:10000           91806 ns       3125 ns      10000
```

I thought I could make it faster.... :confused:

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,211 @@
...
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+
+      // unrolling the loop 8x and 4x times for maximum throughput
+      const size_t loop_8x = nblocks_per_channel / 8;
+      const size_t loop_4x = (nblocks_per_channel - loop_8x * 8) / 4;
+      for (size_t j = 0; j < 8 * loop_8x; j += 8) {
+        __m256 in_m0 = _mm256_loadu_pd(&in[depth_index + 4 * j]);
```
beru: For double version of `global_avepool_op_avx`, you need to change type of these local variables from `__m256` to `__m256d`.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,211 @@
...
+        __m256 in_m1 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 1)]);
+        __m256 in_m2 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 2)]);
+        __m256 in_m3 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 3)]);
+        __m256 in_m4 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 4)]);
+        __m256 in_m5 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 5)]);
+        __m256 in_m6 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 6)]);
+        __m256 in_m7 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 7)]);
+        sum_m        = _mm256_add_pd(
+          sum_m, vectorize::accumulate(in_m0, in_m1, in_m2, in_m3, in_m4, in_m5,
+                                       in_m6, in_m7));
```
beru: You don't need to use vectorize::accumulate inside inner loops.
Please check this example code.
https://github.com/tiny-dnn/tiny-dnn/pull/687/#issuecomment-302877433

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,211 @@
...
+    const vec_t &in = in_data[sample];
+    vec_t &out      = out_data[sample];
+
+    for (size_t i = 0; i < params.in.depth_; i++) {
+      __m256d sum_m            = _mm256_setzero_pd();
+      const size_t depth_index = i * pool_area;
+
+      // unrolling the loop 8x and 4x times for maximum throughput
+      const size_t loop_8x = nblocks_per_channel / 8;
+      const size_t loop_4x = (nblocks_per_channel - loop_8x * 8) / 4;
```
beru: You can move `loop_8x` and `loop_4x` to outside of `params.in.depth_` loop.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,211 @@
...
+        __m256 in_m3 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 3)]);
+        __m256 in_m4 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 4)]);
+        __m256 in_m5 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 5)]);
+        __m256 in_m6 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 6)]);
+        __m256 in_m7 = _mm256_loadu_pd(&in[depth_index + 4 * (j + 7)]);
+        sum_m        = _mm256_add_pd(
+          sum_m, vectorize::accumulate(in_m0, in_m1, in_m2, in_m3, in_m4, in_m5,
+                                       in_m6, in_m7));
+      }
+      for (size_t j = 8 * loop_8x; j < 8 * loop_8x + 4 * loop_4x; j += 4) {
```
beru: `loop_4x` is `0` or `1` in this case. So you don't need to use `for` statement.

tiny_dnn/core/kernels/global_avepool_op_avx.h
```diff
@@ -0,0 +1,211 @@
...
+      for (size_t j = 8 * loop_8x + 4 * loop_4x; j < nblocks_per_channel; j++) {
+        sum_m = _mm256_add_pd(sum_m, _mm256_loadu_pd(&in[depth_index + 4 * j]));
+      }
+
+      __m256d in_m =
+        _mm256_maskload_pd(&in[depth_index + 4 * nblocks_per_channel], imask);
+      sum_m = _mm256_add_pd(sum_m, in_m);
+
+      out[i] = _mm_cvtsd_f64(hsum256_pd(sum_m));
+      out[i] *= pool_area_inv;
```
beru: How about rewrite these 2 lines as
```cpp
_mm_store_sd(&out[i], _mm_mul_sd(hsum256_pd(sum_m), pool_area_inv_m));
```
?

tiny_dnn/layers/global_average_pooling_layer.h
beru: @edgarriba How about disabling `backend_t::avx` entry when `CNN_USE_AVX` isn't defined?

tiny_dnn/util/product.h
```diff
@@ -516,6 +540,36 @@ void fill(T *dst, size_t size, T value) {
...
+}
+
+// vertically accumulate four AVX registers into single register.
+template <typename T>
+MUSTINLINE T accumulate(const T &a, const T &b, const T &c, const T &d) {
+  const T ab = VECTORIZE_TYPE::add(a, b);
+  const T cd = VECTORIZE_TYPE::add(c, d);
+  return VECTORIZE_TYPE::add(ab, cd);
+}
+
```
beru: These `accumulate` functions are too primitive for the purposes.
I don't think it is very meaningful to make them.
What I proposed was to make a function that sums a range of values specified by a pointer and a size arguments.

karandesai-96: This sounds better. I will refactor the methods.

tiny_dnn/core/kernels/global_avepool_op_avx.h
karandesai-96: Oops. Silly error.

tiny_dnn/util/product.h
beru: Thank you for understanding this. After you'll have finished writing new `vectorize::accumulate` function, you can also call it from `global_avepool_op_internal` function which might make the purpose of `global_avepool_op_avx.h` a bit ambiguous...

beru: @edgarriba I've just noticed that we still haven't had mask operation in `tiny_dnn/util/product.h` file. So implementing `vectorize::accumulate` function with TMP may be not ready yet ....
I've realized this since I just started refactoring `tiny_dnn/util/product.h` file.

karandesai-96: @beru for a temporary solution I have kept masking outside of accumulate. As in I am accumulating only full blocks of 8 floats / 4 doubles. That can be refactored once product.h is revamped.

karandesai-96:
@beru I cannot reproduce this bug locally. All tests pass, how shall I proceed ?

tiny_dnn/util/product.h
```diff
@@ -516,6 +551,61 @@ void fill(T *dst, size_t size, T value) {
...
+CNN_MUSTINLINE VECTORIZE_TYPE::register_type accumulate(const float *start,
+                                                        const size_t &nblocks) {
+#endif
+  const size_t n4                  = nblocks / 4;
+  const size_t n2                  = (nblocks % 4) / 2;
+  const size_t n1                  = nblocks % 2;
+  VECTORIZE_TYPE::register_type v0 = VECTORIZE_TYPE::load<aligned>(start + 000);
+  VECTORIZE_TYPE::register_type v1 = VECTORIZE_TYPE::load<aligned>(start + 010);
+  VECTORIZE_TYPE::register_type v2 = VECTORIZE_TYPE::load<aligned>(start + 020);
+  VECTORIZE_TYPE::register_type v3 = VECTORIZE_TYPE::load<aligned>(start + 030);
```
beru: These 8x displacement indexes are for float path. So you need to use 4x ones for double path.

beru: Maybe you can use VECTORIZE_TYPE::unroll_size.
I thought I could apply octal notation for float path 8x displacements but it can't be used for double path.
So you can use C++ User-defined literals to build quaternary notation just for the sake of exploiting language features.

tiny_dnn/util/product.h
```diff
@@ -516,6 +551,66 @@ void fill(T *dst, size_t size, T value) {
...
+// vertically accumulate 'n' AVX registers into single register.
+template <typename aligned>
+#ifdef CNN_USE_DOUBLE
+CNN_MUSTINLINE VECTORIZE_TYPE::register_type accumulate(const double *start,
+                                                        const size_t &nblocks) {
+#else
+CNN_MUSTINLINE VECTORIZE_TYPE::register_type accumulate(const float *start,
+                                                        const size_t &nblocks) {
+#endif
+  const size_t sz = VECTORIZE_TYPE::unroll_size;
```
beru: I'd rename this variable to `vsz` or I'd rather use `VECTORIZE_TYPE::unroll_size` directly in `VECTORIZE_TYPE::load` statements.

beru:
@karandesai-96 Please rebase.

beru:
I'm not sure why `target_cost.train_unbalanced_data_1dim` test on https://travis-ci.org/tiny-dnn/tiny-dnn/jobs/236604806 failed.... Maybe it's worth investigating...

This isn't related with the changes with this PR, but how about renaming
`VECTORIZE_TYPE` to `CNN_VECTORIZE_TYPE`?
Also renaming `CNN_MUSTINLINE` to `CNN_BE_INLINED_ALREADY` sounds perfectly fine if no one opposes.

beru:
And `CNN_MUSTINLINE` macro can be used in other source files. So moving it to `tiny_dnn/util/macro.h` file would be possible so we can spread the macro like a wildfire.

tiny_dnn/util/macro.h
```diff
@@ -12,3 +12,9 @@
 #if defined _WIN32 && !defined(__MINGW32__)
 #define CNN_WINDOWS
 #endif
+
+#ifdef _MSC_VER
+#define CNN_BE_INLINE_ALREADY __forceinline
+#else
+#define CNN_BE_INLINE_ALREADY __attribute__((always_inline)) inline
+#endif
```
beru: nyanp suggested to add a fallback to inline so I think it'd be good to follow it.
https://github.com/tiny-dnn/tiny-dnn/pull/733#discussion_r118842307

And I recommended to use the macro name `CNN_BE_INLINE_ALREADY` as a joke. :joy:

karandesai-96: @beru `CNN_BE_INLINE_ALREADY` sounds great anyways :joy: can we keep it ?!

karandesai-96: Oka let me revert it back. 
image

tiny_dnn/layers/global_average_pooling_layer.h
edgarriba: Ups didn't saw this one. That's another solution. But then what about adding a static assert with a proper warning message?

edgarriba:
@karandesai-96 is a lot of work to organize this PR? I see three different tasks here:
1. Implement GAP in AVX
2. Include benchmarks framework support (you should add some documentation about how to run the benchmarks)
3. Add benchmarks​ for GAP

benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+include(../cmake/DownloadProject/DownloadProject.cmake)
+download_project(
+        PROJ benchmark
+        GIT_REPOSITORY https://github.com/google/benchmark.git
```
edgarriba: We might want to specify a version here

karandesai-96:
@edgarriba I couldn't pay much attention to this as I was entangled in debugging the parameters API. A new PR for that soon, but keeping that aside I will wrap it up here now...

benchmarks/CMakeLists.txt
karandesai-96: Done. v1.1.0

edgarriba:
@karandesai-96 could you rename 494e1bba1ad2918c536ecb5db30ffca900959be9 saying something like "add google bancharmk framework"?

beru:
@karandesai-96 You need to replace `CNN_MUSTINLINE` with `CNN_MUST_INLINE` in file `tiny_dnn/util/product.h`. How couldn't you notice the little underbar trap I added...

karandesai-96:
@beru Be right back, make a coffee with stronger caffeine content xD

CMakeLists.txt
```diff
@@ -58,11 +58,12 @@ option(USE_DOUBLE     "Build tiny-dnn with double precision computations"  OFF)
...
 
-option(BUILD_TESTS    "Set to ON to build tests"              OFF)
-option(BUILD_EXAMPLES "Set to ON to build examples"           OFF)
-option(BUILD_DOCS     "Set to ON to build documentation"      OFF)
-option(COVERALLS      "Set to ON to build with code coverage" OFF)
-option(PROFILE        "Set to ON to build with profiling"     OFF)
+option(BUILD_TESTS      "Set to ON to build tests"              OFF)
+option(BUILD_EXAMPLES   "Set to ON to build examples"           OFF)
+option(BUILD_DOCS       "Set to ON to build documentation"      OFF)
+option(BUILD_BENCHMARKS "Set to ON to build documentation"      OFF)
```
Randl: `s/documentation/benchmark/`.

benchmarks/bm_global_avepool.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
```
Randl: Licence comment is missing

cmake/clang-cxx-dev-tools.cmake
```diff
@@ -3,6 +3,7 @@
 
 file(GLOB_RECURSE ALL_CXX_SOURCE_FILES
   ${CMAKE_SOURCE_DIR}/tiny_dnn/*.h
+  ${CMAKE_SOURCE_DIR}/benchmarks/*.h
```
Randl: `*.cpp` too

tiny_dnn/core/kernels/global_avepool_grad_op.h
```diff
@@ -27,10 +28,17 @@ class GlobalAvePoolGradOp : public core::OpKernel {
...
     fill_tensor(prev_delta, float_t{0});
 
-    // only internal kernel op implemented yet, so use it regardless
-    // of the specified backend engine
-    kernels::global_avepool_grad_op_internal(prev_delta, curr_delta, params,
-                                             context.parallelize());
+    const core::backend_t engine = context.engine();
+
+    if (engine == core::backend_t::avx) {
+#ifdef CNN_USE_AVX
```
Randl: This would mean that if `engine` is set to `core::backend_t::avx` and `CNN_USE_AVX` is undefined, nothing happens? This is kinda not intuitive and also inconsistent with other layers where we fall back to internal in this case.

tiny_dnn/layers/global_average_pooling_layer.h
beru: I think removing `backend_t::avx` entry if `CNN_USE_AVX` isn't defined is a better solution.
We don't need to rely static_assert or nn_warn exception.
This is C++ header only library, we can keep it simple.

tiny_dnn/util/macro.h
```diff
@@ -12,3 +12,11 @@
...
 #define CNN_WINDOWS
 #endif
+
+#if defined(_MSC_VER)
+#define CNN_MUSTINLINE __forceinline
+#elif defined(__GNUC__) || defined(__clang__) || defined(__ICC)
+#define CNN_MUSTINLINE __attribute__((always_inline)) inline
+#else
+#define CNN_MUSTINLINE inline
+#endif
```
beru: @karandesai-96 You_should_start_liking_to_insert_an_underbar_between_words.

tiny_dnn/core/kernels/global_avepool_grad_op.h
beru: @Randl I think killing `core::backend_t::avx` entry and all the AVX related codes when `CNN_USE_AVX` isn't defined would be a simpler solution.

beru: I mean disabling the code with preprocessor directives.

edgarriba: Vote for it

bhack: :+1:

beru: @edgarriba You need to remember that voting system is powerless before you.

edgarriba:
Good to merge?

karandesai-96:
@edgarriba will fix the typo and license comment once I finish travelling.

tiny_dnn/util/macro.h
karandesai-96: @Beru `CNN_SCREAMING_SNAKE_CASE_FOR_THE_WIN`

tiny_dnn/util/macro.h
```diff
@@ -12,3 +12,11 @@
 #if defined _WIN32 && !defined(__MINGW32__)
 #define CNN_WINDOWS
 #endif
+
+#if defined(_MSC_VER)
+#define CNN_MUSTINLINE __forceinline
```
beru: There's sth missing between `T` and `I`. (Hint : 0x5F)

tiny_dnn/util/macro.h
```diff
@@ -12,3 +12,11 @@
...
+#elif defined(__GNUC__) || defined(__clang__) || defined(__ICC)
+#define CNN_MUST_INLINE __attribute__((always_inline)) inline
+#else
+#define CNN_MUSTINLINE inline
```
beru: Feel free to perform statement coverage test for this path.

edgarriba: is that what makes appveyor fail?

beru: This one https://github.com/tiny-dnn/tiny-dnn/pull/687#discussion_r121675457
is the cause.

tiny_dnn/util/macro.h
karandesai-96: How did I miss this :confused: thanks for pointing out

edgarriba:
cool! @beru @karandesai-96 see https://github.com/tiny-dnn/tiny-dnn/issues/766

PR #691, opened 2017-05-07
Enable ASan in gcc/clang for tests

Randl:
Add AddressSanitizer to tests in debug mode.


edgarriba:
Are you assuming that `enable_testing` switch to Debug mode?

Randl:
I assume there is no point to use ASan in Release mode, due to meaningless error messages.

edgarriba:
if you run cmake in release mode and build tests, you'll add sanitizer flags

Randl:
so you suppose to enable it in all 4 modes?

edgarriba:
Nope just saying that we should include and if statement to make sure that we include this flags in debug mode. Check
https://github.com/tiny-dnn/tiny-dnn/blob/master/CMakeLists.txt#L38

Randl:
Should we? We are adding flags to `CMAKE_CXX_FLAGS_DEBUG`, that should guarantee that ASan works only in debug mode, shouldn't it?

Randl:
ping @edgarriba

CMakeLists.txt
```diff
@@ -272,6 +273,24 @@ if(PROFILE AND (CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
...
+        set(ASAN_FLAGS "-fsanitize=address -fno-omit-frame-pointer")
+        if (HasUseAfterScope)
+            set(ASAN_FLAGS "${ASAN_FLAGS} -fsanitize-address-use-after-scope")
+        endif()
+    endif()
+    set(CMAKE_CXX_FLAGS         "${CMAKE_CXX_FLAGS} ${ASAN_FLAGS}")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${ASAN_FLAGS}")
+    set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} ${ASAN_FLAGS}")
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} ${ASAN_FLAGS}")
+    set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} ${ASAN_FLAGS}")
```
beru: I'm not so sure but
```
set(CMAKE_CXX_FLAGS         "${CMAKE_CXX_FLAGS} ${ASAN_FLAGS}")
```
isn't enough?

Are these also necessary?
```
    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${ASAN_FLAGS}")
    set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} ${ASAN_FLAGS}")
    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} ${ASAN_FLAGS}")
    set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} ${ASAN_FLAGS}")
```

Randl: AFAIK `CMAKE_CXX_FLAGS` is used if none of 4 modes is chosen. It's not common flags as might be speculated

beru: Hmmmm then how about this?
```
add_definitions(${ASAN_FLAGS})
```

Randl: OK, I was actually wrong, and `CMAKE_CXX_FLAGS`are mutual flags. I'll fix it later today

Randl:
ping @edgarriba @nyanp

Randl:
@beru we should merge https://github.com/tiny-dnn/tiny-dnn/pull/536 ,  it completely went out of my mind. That might resolve maxthread problem

PR #693, opened 2017-05-07
Fully connected xtensor integration

Randl:
I've started integration of xtensor, but its very raw yet.


bhack:
xtensor 0.10.1 is out.

tiny_dnn/core/kernels/fully_connected_grad_op.h
```diff
@@ -65,32 +67,36 @@ class FullyConnectedGradOp : public core::OpKernel {
...
+    xt::xarray<float_t> dW             = to_xtensor(context.input_grad(1));
+    xt::xarray<float_t> dB             = to_xtensor(context.input_grad(2));
+    xt::xarray<float_t> *db            = params.has_bias_ ? &dB : nullptr;
+    xt::xarray<float_t> prev_delta     = to_xtensor(context.input_grad(0));
+    xt::xarray<float_t> curr_delta     = to_xtensor(context.output_grad(0));
+    xt::xarray<float_t> dummy;  // need lvalue for non-const reference
 
     // initialize outputs
-    fill_tensor(prev_delta, float_t{0});
+    prev_delta = xt::zeros<double>(prev_delta.shape());
```
beru: Should `double` be used for the template parameter? Not `float_t`?

beru:
I see a lot of `xt::xarray<float_t>` syntaxes but they seem alienating.
Can't you wrap them with tensor class?

beru:
I bet it's super-duper hard but these in `tiny_dnn/util/util.h` file
```cpp
 typedef std::vector<float_t, aligned_allocator<float_t, 64>> vec_t;
 typedef std::vector<vec_t> tensor_t;		 typedef std::vector<vec_t> tensor_t;
```
should be replaced with new class tailored with `xtensor` library.

I'd say directly using `xtensor` library would be a bit problematic.
By doing so, this library will be hardly dependent on xtensor and decoupling it will be hard.

edgarriba:
I agree with @beru, we should wrap xtensor functionalities to our `Tensor` class or just create a new `XTensor`. Besides, is xtensor in correct place or should be inside core?

SylvainCorlay:
At a very high level, instead of taking const references to `xtensor` or `xarray` as argument, I would recommend making the function template and take `xexpression`s

```cpp
template <class E1, class E2>
void foo(const xexpression<E1>&, const xexpression<E2>&);
```

this would make the functions work with both container types of xtensor (`xarray` and `xtensor`) but also any unevaluated tensor expressions.

Our bindings to numpy and julia also relies on that. `pyarray` and `pytensor` are valid xexpressions and wrap numpy arrays inplace.

edgarriba:
@Randl so then what about encapsulate `xtensor` to `Tensor` ?

Randl:
@edgarriba that would kinda limit us in term of standard tensor functions, but I'll do my best to make integration less invasive

Randl:
@edgarriba that would kinda limit us in term of standard tensor functions, but I'll do my best to make integration less invasive

SylvainCorlay:
The main benefit of using expressions instead of a `Tensor` container is that you can abstract out the underlying storage. In the long run, this could enable the use of tiny-dnn with tensors backed by database calls, or bindings to Julia and Python whose arrays can be wrapped in place as tensor expressions.

beru:
Abstracting storage type will be needed if this library dynamically supports several numerical types.
Integer types are used to hold `half float` or `short float` and for quantization kernels.

`xtensor` library should be good for the implementation, but another class will be needed for flexible data pipeline.

SylvainCorlay:
@beru I don't understand what you mean. xtensor's main feature is that it completely abstract storage.

If you do for example `stack(a, b, c)`, you get an un-evaluated tensor *expression* where `a` may be backed by a database, `b` an in-memory container and `c` another composed expression. Also, you can create xtensor expression from any existing containers.

SylvainCorlay:
@beru edited the above I misread your comment.

beru:
@SylvainCorlay It seems to me that Compile time Polymorhism is pretty much OK with the implementation. But I fear that doesn't give us runtime type flexibity automatically due to its nature.

Though I haven't used xtensor library yet, so your comments are pretty helpful.

SylvainCorlay:
Cool. I think that for sort of thing, one should favor compile-time polymorphism (CRTP) over runtime polymporphism.

Indeed, adding the overhead of virtual function table lookups to math operations is a performance killer.

The right place for dynamic polymorphism is certainly at a higher level, for e.g. dataframes with columns of different types.

beru:
@SylvainCorlay 
> Cool. I think that for sort of thing, one should favor compile-time polymorphism (CRTP) over runtime polymorphism. Adding the overhead of virtual function table lookups to math operations is a performance killer.

If this library will continue to use statically defined numerical computational type (which is `float` or `double`) throughout the neural network. That should be fine.

@wangyida implemented quantized layers in 2016 gsoc.
https://summerofcode.withgoogle.com/archive/2016/projects/4623962327744512/

If you looked at the implementation, you can see type conversions take place very often which spoil the speed benefit of quantized computation.
https://github.com/tiny-dnn/tiny-dnn/blob/0e372ddb1cf483e5d694a62c91868b6241c887ca/tiny_dnn/core/kernels/tiny_quantized_conv2d_kernel.h#L35

beru:
@SylvainCorlay Do you use CRTP (The curiously recurring template pattern) often?
It reminds me of the good old days with WTL...

I'm not a virtuoso-like savvy C++ programmer, so reading tons of error msgs is sometimes hard for me.
For example, deceased Maxim Shemanarev's Anti Grain Geometry library is a masterpiece but so didn't use it effortlessly.

SylvainCorlay:
Yes, CRTP is the way to go!

Randl:
@beru You're using VS, right? Can you check what's the issue with it?

CMakeLists.txt
```diff
@@ -264,9 +264,9 @@ endif()
 
 ####
 # Set compiler options
-set(CMAKE_CXX_FLAGS         "${CMAKE_CXX_FLAGS} ${EXTRA_C_FLAGS}")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${EXTRA_C_FLAGS_RELEASE}")
-set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} ${EXTRA_C_FLAGS_DEBUG}")
+set(CMAKE_CXX_FLAGS         "${CMAKE_CXX_FLAGS} ${EXTRA_C_FLAGS} -fsanitize=address")
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${EXTRA_C_FLAGS_RELEASE} -fsanitize=address")
+set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} ${EXTRA_C_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address -O0")
```
beru: Microsoft Visual C++ doesn't support these options.

beru:
@Randl I don't have Windows in my laptop. I can use it tomorrow though.

beru:
@Randl
I'd reluctantly admit using `xexpression` template parameters.
But unlike `xtensor` type, it can't let compiler check fixed dimensionality so you may put assert statements.
And you should probablly rename `to_xtensor` and `from_xtensor` functions to `to_xarray` and `from_xarray`.

tiny_dnn/core/kernels/fully_connected_grad_op.h
```diff
@@ -65,32 +67,38 @@ class FullyConnectedGradOp : public core::OpKernel {
...
+    xt::xarray<float_t> curr_delta = to_xtensor(context.output_grad(0));
+    xt::xarray<float_t> dummy;  // need lvalue for non-const reference
 
     // initialize outputs
-    fill_tensor(prev_delta, float_t{0});
+    prev_delta = xt::zeros<float_t>(prev_delta.shape());
 
     // call the algorithm depending on the selected engine type
 
     const core::backend_t engine = context.engine();
```
beru: If you don't use `xtensor` type, dimensionality can't be checked at compile time.
Current fully_connected kernels only support certain types of data, so if you wildly expand genericness of the parameters for some reasons, I advise you to add some defensive checking code as well.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,14 +14,21 @@ namespace kernels {
...
+          class E3,
+          class E4,
+          value_is_float<E1> * = nullptr,
+          are_all_xexpr<E1, E2, E3, E4> * = nullptr>
+inline void avx_fully_connected_forward_kernel(const E1 &in_data,
+                                               const E2 &W,
+                                               const E3 &bias,
+                                               E4 &out_data,
+                                               const fully_params &params,
+                                               const bool layer_parallelize) {
```
beru: I'd say this generic function interface is overkill.
Because current `avx_fully_connected_forward_kernel` implementation only supports specific types of data.

Randl: yeah but that allows you to pass `xtensor`, `xarray`, `xview`, etc...

tiny_dnn/core/kernels/fully_connected_grad_op.h
Randl: Sure, I just wanted to get a working version first, and then improve it to a standard version which would be expanded to the other layers.

beru:
I'm not so sure about the scope of xtensor integration.

If getting rid of fixed dimensionality of current implementation is the purpose, using `xarray` for tensor storage seems kinda okay. But to apply such change throughout this library, you may simplify the code by removing other backends so that you can modify internal backend routines without worrying about other things in the process.

tiny_dnn/core/kernels/fully_connected_op_avx.h
beru: Making interface generic for specific implementation is a wrong move.

tiny_dnn/core/kernels/fully_connected_grad_op.h
beru: Hoo boy.... applying this method to all other layers? That must be an endeavour.

tiny_dnn/core/kernels/fully_connected_op_avx.h
Randl: I was going to use it for perfect forwarding as advised by @SylvainCorlay . Another option is to use the patter proposed upper in the thread. I chose this one since it's less invasive and shorter (doesn't involve multiple `e1.derived_cast();`)

beru: Please do not put new wine into old bottles.

beru: I guess I shouldn't use STL vector type for this very concrete implementation...
If I had used pointers and length parameters, changes of this PR would have been much simpler..

Randl: > Please do not put new wine into old bottles.

What do you mean?

Randl:
ping @edgarriba @bhack

tiny_dnn/core/framework/tensor.h
```diff
@@ -351,16 +277,20 @@ class Tensor {
    * matrix view from
    *                                                     // offset 4.
    */
-  Tensor subView(std::initializer_list<size_t> const &start,
-                 std::initializer_list<size_t> const &new_shape) {
-    return subview_impl(start, new_shape);
+  template <class... Ranges>
+  Tensor subView(Ranges... ranges) {
+    return Tensor(xt::view(storage_, from_init_lists(ranges)...));
```
edgarriba: @Randl so is this wroking?

Randl: no it doesn't since init list can't be forwarded (at least I think that's the reason)

Randl:
@SylvainCorlay Could you please take a look at https://github.com/tiny-dnn/tiny-dnn/pull/693/files#diff-54658142cc96675fb83229795a17df30R267 ? I'm trying to forward parameters as `xt::range` but something definitely goes wrong

tiny_dnn/core/framework/tensor.h
```diff
@@ -268,9 +281,10 @@ class Tensor {
    * with offset zero
    *
    */
-  // TODO
   Tensor subView(std::initializer_list<size_t> const &new_shape) {
-    return Tensor(xt::broadcast(storage_, new_shape));
+    Tensor res = Tensor(xt::broadcast(storage_, storage_.shape()));
```
edgarriba: does then the new tensor share storage with the previous or it's just a copy of the view?

Randl: I'll write additional tests to make sure, but it should share the storage

edgarriba: nice, yeah we have to make sure that it happens

edgarriba:
@Randl why are network test failing, do didn't touch that codes right? Besides, could you split this PR in two? xtensor interface and xtensor fully-connected integration. Other project are somehow pending of the first

Randl:
@edgarriba I think of reverting FC code, since it need to use Tensor and not xtensor now. So that'll be general xtensor integration ("big" PR)  and then I'll go with small layer integration (like in your PR). Also this PR got messed due to rebases, I'll reopen it when more or less ready

Randl:
Reopened as https://github.com/tiny-dnn/tiny-dnn/pull/739

PR #701, opened 2017-05-09
One assert statement fixed

vekkuli:


Randl:
Thank you! Good catch

PR #706, opened 2017-05-11
Removed redundant clang-check and coveralls from travis

nishnik:
Calculating coveralls was taking around 10s-20s for each run.
So doing it only once is a better option.

Same goes for clang-check.

Now these both would be done only once.

Results:
#699 (the latest commit) took 29 minutes to run
this takes less than 20 minutes (I ran travis on my repo) ([here](https://travis-ci.org/nishnik/tiny-dnn/builds/231088931) and [here](https://travis-ci.org/nishnik/tiny-dnn/builds/231094418))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,12 +97,12 @@ before_script:
 script:
   - make -j2
   - test/tiny_dnn_test
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" == "g++-4.8" ] && [ "$USE_SSE" == "ON" ] && [ "$USE_AVX" == "ON" ] && [ "$USE_DOUBLE" == "ON" ]; then
       make clang-format-check;
     fi
 
 after_success:
-  - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" == "g++-4.8" ]; then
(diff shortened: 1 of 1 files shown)


Randl:
Well, probably you should leave two checks for coveralls, if I got it right

.travis.yml
```diff
@@ -97,12 +97,12 @@ before_script:
...
   - make -j2
   - test/tiny_dnn_test
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" == "g++-4.8" ] && [ "$USE_SSE" == "ON" ] && [ "$USE_AVX" == "ON" ] && [ "$USE_DOUBLE" == "ON" ]; then
       make clang-format-check;
     fi
 
 after_success:
-  - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" == "g++-4.8" ]; then
+  - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" == "g++-4.8" ] && [ "$USE_SSE" == "ON" ] && [ "$USE_AVX" == "ON" ] && [ "$USE_DOUBLE" == "ON" ]; then
```
edgarriba: with this we are no covering float based codes

nishnik: What can be done for that?

Randl: I believe we need all three runs -_-

nishnik: Done!

nishnik:
ping @Randl

Randl:
ping @edgarriba

PR #713, opened 2017-05-14
Added ImageNet inference example using pretrained AlexNet

shoaibahmed:
Added ImageNet inference example using pretrained AlexNet converted from
caffe using example_caffe_convert_model by feeding the deploy prototxt
and caffemodel.


beru:
Please apply `clang-format` to updated files otherwise travis-ci fails.

shoaibahmed:
Should I add a new pull request for the updated version?

edgarriba:
@shoaibahmed thanks for your contribution, but do you think that you can extend this PR and use this as interface?
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/models/alexnet.h

/cc @karandesai-96 will be working on a model-zoo during the summer, so maybe you guys can define the proper way to handle this

shoaibahmed:
Thanks for the guideline. I will extend the previously defined interface. Will close the current PR then.

bhack:
You can do here. You need to just commit to your branch

shoaibahmed:
@edgarriba I have looked at the defined interface. The defined interface does not perfectly conform to the model defined in caffe.  I am currently loading the complete model (including both weights and graph) from the converted caffe prototxt and caffemodel file. Here is the visualization of the loaded model which clearly indicates the differences with the defined interface.
graph

Do let me know what you have in mind. Thanks!

edgarriba:
@shoaibahmed where did you get the model from? I  mean since it works, feel free to modify it. /cc @karandesai-96

Issue #714, opened 2017-05-17
Fully connected layer runs (much) slower if parallelization is on


marty1885:
Hello. I found that turning off parallelization in fully_connected_layer by setting `ctx.setParallelize` to false in `fully_connected_layer.h` makes it run faster.
Epoching a MLP of size 1024, 784, 120, 10 against MNIST(by modifying the example code) takes 67s with parallelization on and 63.17 when parallelization is off.
Further more, training a MLP of size 60, 60, 60, 60, 15, 1 against 14000 input data with parallelization on is 11x slower than parallelization off. (Sorry, I don't have permission to share this piece of code.)
This seems to happen due to tiny-dnn is doing matrix multiplication in parallel for FC layers even if the layer is too small to benefit from parallelization.
Modified version of MNIST example: https://gist.github.com/marty1885/0f419893c7cc814a34773ef55bd0e671

Tested on: 
OS: Manjaro Linux (Linux 4.11)
Compiler: GCC 6.2.0/6.3.1
Compile flags: -O3 / -O3 -march=native
CPU: Intel 17 6700
RAM: DDR4 2133 8G*2 (Dual Channel)

Randl:
```cpp
nn  << fc(60, 60, true, backend_type)  << relu()
<< fc(60, 60, true, backend_type)  << relu()       
 << fc(60, 60, true, backend_type)  << relu()
<< fc(60, 15, true, backend_type)  << relu() 
<< fc(15, 1, true, backend_type)  << softmax(1); 
```

This can be considered as an example for second network you mentioned? Which activation and backend were used?

Polyacene:
@marty1885 @Randl  I just checked that I have a similar  issue on a 71 x 251 x 64 network.
The code runs faster when compiled with
`g++ -O3 -march=native -pthread -std=c++11`
compared to
`g++ -O3 -march=native -fopenmp -std=c++11`
Note that in the first version I only get a 100% load, while the second goes up to a 800% load.

In config.h I enabled
```#define CNN_USE_AVX
#define CNN_USE_OMP
#define CNN_USE_EXCEPTIONS
#define CNN_USE_STDOUT
```

Polyacene:
@marty1885 @Randl I checked my simple interpolation example [interpolate.C.txt](https://github.com/tiny-dnn/tiny-dnn/files/1007426/interpolate.C.txt), and there you observe a similar behavior.
[...]

PR #716, opened 2017-05-19
Add protected layer::for_i method that uses layer_::parallelize_ member

beru:
Hi, this PR adds protected `layer::for_i` method that uses `layer_::parallelize_` member.
The method calls `for_i` function defined in `tiny_dnn/util/parallel_for.h` file.
The benefit of the method is it automatically uses `layer_::parallelize_` member as a first parameter of `for_i` function.

This PR also adds other changes such as
- Add a missing new line character in the error message of `example_mnist_test` program.
- Remove using-declarations that introduce `layer::parallelize_` member into derived class public scope. (I concluded they aren't necessary but please correct me if I'm wrong about this)

I briefly checked if there's any performance improvement with the changes, and it seems very little.
Though, I didn't check the effect of all the changes. So there can be negative effects in some cases.

#714 is somehow related.

### DIFF
--- a/examples/mnist/test.cpp
+++ b/examples/mnist/test.cpp
@@ -72,7 +72,7 @@ void recognize(const std::string &dictionary, const std::string &src_filename) {
 
 int main(int argc, char **argv) {
   if (argc != 2) {
-    cout << "please specify image file";
+    cout << "please specify image file" << std::endl;
     return 0;
   }
   recognize("LeNet-model", argv[1]);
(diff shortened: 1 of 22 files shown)


beru:
Maybe, moving layer's operation implementations into kernels is expected so adding `layer::for_i` method is somewhat wrong. But we still have a lot of layers that have its operation impl in their methods. So in that regard, I guess adding the method can be justifiable from implementors's point of view.

PR #717, opened 2017-05-20
Reduce frequent memory allocation

beru:
Hi, this PR reduces an amount of memory allocation and copy.

Data layout of mini-batch is reordered after this commit.
https://github.com/tiny-dnn/tiny-dnn/commit/588fc2256116ebce84c9d93ce02a3ca4d24bcca2#diff-0a7e84c2581225319e60be95f342f76bR186

Let me address this comment. @reunanen
https://github.com/tiny-dnn/tiny-dnn/commit/588fc2256116ebce84c9d93ce02a3ca4d24bcca2#diff-0a7e84c2581225319e60be95f342f76bR193 

from : [sample][channel][feature]
to   : [channel][sample][feature]
data layout reoredering takes place in `nodes::reorder_for_layerwise_processing`

and 

from : [channel][sample][feature]
to   : [sample][channel][feature]

backward data layout reoredering takes place in `sequential::normalize_out` and `graph::merge_outs`

The downside of this change is worsened code readability.

### DIFF
--- a/tiny_dnn/layers/fully_connected_layer.h
+++ b/tiny_dnn/layers/fully_connected_layer.h
@@ -66,12 +66,12 @@ class fully_connected_layer : public layer {
   void forward_propagation(const std::vector<tensor_t *> &in_data,
                            std::vector<tensor_t *> &out_data) override {
     // forward fully connected op context
-    auto ctx = OpKernelContext(in_data, out_data);
-    ctx.setParallelize(layer::parallelize());
-    ctx.setEngine(layer::engine());
+    fwd_ctx_.set_in_out(in_data, out_data);
+    fwd_ctx_.setParallelize(layer::parallelize());
+    fwd_ctx_.setEngine(layer::engine());
 
     // launch fully connected kernel
-    kernel_fwd_->compute(ctx);
+    kernel_fwd_->compute(fwd_ctx_);
   }
 
   void back_propagation(const std::vector<tensor_t *> &in_data,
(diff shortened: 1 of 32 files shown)


Randl:
Should we use smart pointers? Have you noticed significant speedup on you computer?

beru:
> Should we use smart pointers?

Maybe not. (Though I'm not sure what kind of smart ptrs you have in mind.)
If new Tensor class would provide reordered interface without changing inner data, that can be helpful.

> Have you noticed significant speedup on you computer?

The performance gain is around 1.5% with `example_mnist_train` program.
So, it's not significant at all.

Here are log files. 
[master_72fcbfc312c35b96d651c89e170f2493a6deb395.txt](https://github.com/tiny-dnn/tiny-dnn/files/1016005/master_72fcbfc312c35b96d651c89e170f2493a6deb395.txt)
[reduce_memory_allocation_177cc3b58c95d072003ebdb2ebf96eae99093b29.txt](https://github.com/tiny-dnn/tiny-dnn/files/1016006/reduce_memory_allocation_177cc3b58c95d072003ebdb2ebf96eae99093b29.txt)

I tested on Windows10 and this is the CMake command I used.
`cmake -G "Visual Studio 14 2015 Win64" -DUSE_AVX2=ON -DBUILD_EXAMPLES=ON ..\`

Randl:
> Though I'm not sure what kind of smart ptrs you have in mind.

http://en.cppreference.com/w/cpp/memory/shared_ptr and http://en.cppreference.com/w/cpp/memory/unique_ptr

> If new Tensor class would provide reordered interface without changing inner data, that can be helpful.

Not sure that it can be easily introduce, but it's worth checking

tiny_dnn/layers/fully_connected_layer.h
```diff
@@ -117,6 +117,12 @@ class fully_connected_layer : public layer {
   /* The layer parameters */
   fully_params params_;
 
+  /* forward op context */
+  OpKernelContext fwd_ctx_;
```
edgarriba: OK nice move, but maybe context in a future should also contain shared attributes among all instances of the layer e.g. cuda/ocl kernels, etc

beru:
> http://en.cppreference.com/w/cpp/memory/shared_ptr and http://en.cppreference.com/w/cpp/memory/unique_ptr

@Randl 

For the first one `shared_ptr`, I think it's better to manage ownerships of tensor data explicitly. But well... if we could have used garbage collector, we would't have this memory copy issue? I guess the problem partlly comes from lack of `span` (former `array_view`) class.

For the second one `unique_ptr`, I regard it as a replacement for `std::auto_ptr` and it's not meant for copying instance so maybe it's hard to find a use case in here? It's good to prevent memory leaks but the obstacle I have here is to reduce unnecessary memory copies.

And I wrote some comments about inefficient part of the code with this PR.
I thought I could reduce the problem but it still remains all over the places.

[exp.hpp.txt](https://github.com/tiny-dnn/tiny-dnn/files/1017094/exp.hpp.txt)

PR #722, opened 2017-05-21
use SPMD instead of parallel for for OpenMP in for_i

marty1885:
This PR uses SPMD pattern instead of parallel for for OpenMP in `for_i`. Which reduces function call overhead, thus making `for_i` with OpenMP run faster.
This PR makes a training a MLP of size 60,60,60,60,15,1 with 14000 input vectors 2x faster. And traning a MLP of size 1024, 784, 120, 10 for MNIST 3% faster. Not sure if there are any negative performance effects elsewhere.

#714 is related to this.

### DIFF
--- a/tiny_dnn/util/parallel_for.h
+++ b/tiny_dnn/util/parallel_for.h
@@ -26,6 +26,10 @@
 #include <tbb/tbb.h>
 #endif
 
+#ifdef CNN_USE_OMP
+#include <omp.h>
+#endif
+
 #if !defined(CNN_USE_OMP) && !defined(CNN_SINGLE_THREAD)
 #include <future>
 #include <thread>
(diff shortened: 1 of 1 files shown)


Randl:
Please apply clang-format

beru:
Please refer below document.

Personally, I change this line
https://github.com/tiny-dnn/tiny-dnn/blob/72fcbfc312c35b96d651c89e170f2493a6deb395/.travis/clang-format.sh#L9
from
`STYLE=$(which clang-format-4.0)`
to
`STYLE=$(which clang-format)`
before using
```sh
$ .travis/clang-format.sh
```

edgarriba:
@marty1885 could you apply style format so that we can proceed to merging?

PR #677, opened 2017-04-30
Update file header comments and AUTHORS file

beru:
Hi, this PR updates file header comments and AUTHORS file.
#228 is related.

PR #729, opened 2017-05-24
Recurrent neural networks

prlz77:
This PR extends tiny-dnn to allow recurrent models. List of extensions:
   
**Recurrent Cell**: most basic rnn module implementing equations for a simple rnn cell.
- [x] Inference.
- [x] Training.
- [x] Unit tests.
- [x] Serialization.

Next PR will add a wrapper on cells that will manage state transitions.


tiny_dnn/layers/recurrent_cell_layer.h
```diff
@@ -0,0 +1,174 @@
...
+    params_.has_bias_   = has_bias;
+    params_.activation_ = new activation_type;
+  }
+
+  void init_backend(backend_t backend_type) {
+    core::OpKernelConstruction ctx =
+      core::OpKernelConstruction(layer::device(), &params_);
+
+    if (backend_type == backend_t::internal) {
+      kernel_fwd_.reset(new RecurrentCellOp(ctx));
```
edgarriba: @prlz77 you need to add this workaround solution to pass the tests
`if (backend_type == backend_t::internal | backend_type == backend_t::avx ) {`

besides, take a look at https://github.com/tiny-dnn/tiny-dnn/blob/94419fe11e78f3338ea01bae874c30b58aa1d097/tiny_dnn/core/backend.h#L42
probably we should rethink this small routine and its impact

prlz77: Hi thanks! I have fixed it.

Indeed, there is a lot of repetition with the backends. I wanted to simplify it globally with this PR but I think it might be better to do a new PR removing all the backend checks until the ops.

Having a default() somewhere seems correct to me, but it feels forced right now.

edgarriba:
@prlz77 need help with this?

prlz77:
@edgarriba Well, it compiles and passes tests on my computer but tests here still complain about AVX, do you have any clue why?

prlz77:
I just realized it fails with:
``/home/travis/.travis/job_stages: line 54: test/tiny_dnn_test: No such file or directory``

beru:
You should make a topic branch with a meaningful name.

prlz77:
@beru ok, I had it into ``rnn`` in my computer but merged in master because of the file you reference:
```
Create a topic branch where you want to base your work.
* This is usually the **master** branch.
```
Should I close this PR and open another new PR then?

beru:
> Should I close this PR and open another new PR then?

@prlz77 Not really. Just a recommendation.
I'm trying to reproduce the problem locally...

tiny_dnn/core/kernels/recurrent_cell_op.h
```diff
@@ -0,0 +1,97 @@
...
+    DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+    FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+    DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+    SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+   LIABILITY,
+    OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+   USE
+    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
beru: Please replace long file headers into short version.
You may refer #677

tiny_dnn/core/kernels/recurrent_cell_op.h
```diff
@@ -0,0 +1,97 @@
...
+
+    // initialize outputs
+    fill_tensor(out_data, float_t{0});
+    fill_tensor(next_h, float_t{0});
+
+    // call the algorithm depending  on the selected engine type
+
+    const core::backend_t engine = context.engine();
+
+    if (engine == core::backend_t::internal || engine == core::backend_t::avx) {
```
beru: I don't think this `if` statement is needed since what we have now is `recurrent_cell_op_internal` function. We can't choose other options.

tiny_dnn/layers/recurrent_cell_layer.h
```diff
@@ -0,0 +1,169 @@
...
+    ctx.setParallelize(layer::parallelize());
+    ctx.setEngine(layer::engine());
+
+    // launch recurrent kernel
+    kernel_back_->compute(ctx);
+  }
+
+  std::string layer_type() const override { return "recurrent"; }
+
+#ifndef CNN_NO_SERIALIZATION
```
beru: You don't need this preprocessor macro.
Please refer #667 
An empty `struct serialization_buddy` will be there if `CNN_NO_SERIALIZATION` isn't defined so there's no need for `#ifdef CNN_USE_GEMMLOWP` preprocessor directives in layer classes.

tiny_dnn/layers/recurrent_layer.h
```diff
@@ -0,0 +1,128 @@
...
+    ctx.setParallelize(layer::parallelize());
+    ctx.setEngine(layer::engine());
+
+    // launch recurrent connected kernel
+    kernel_back_->compute(ctx);
+  }
+
+  std::string layer_type() const override { return "recurrent"; }
+
+#ifndef CNN_NO_SERIALIZATION
```
beru: Same here. You can remove this preprocessor directive and simply put `friend struct serialization_buddy;` line even if `CNN_NO_SERIALIZATION` isn't defined.

I initially put these preprocessor directives to many layer classes but later I realized that they're pretty messy.

tiny_dnn/layers/recurrent_layer.h
```diff
@@ -0,0 +1,128 @@
...
+    params_.out_size_ = out_size;
+    params_.has_bias_ = has_bias;
+  }
+
+  void init_backend(backend_t backend_type) {
+    core::OpKernelConstruction ctx =
+      core::OpKernelConstruction(layer::device(), &params_);
+
+    if (backend_type == backend_t::internal || backend_type == backend_t::avx ||
+        backend_type == backend_t::nnpack) {
```
beru: Why do we need this `if` statement? Are there any restrictions?

tiny_dnn/tiny_dnn.h
```diff
@@ -84,6 +86,13 @@ using ave_pool = tiny_dnn::average_pooling_layer;
 
 using fc = tiny_dnn::fully_connected_layer;
 
+// TODO(prlz77) RNN Wrapper
+// template<class T>
+// using recurrent = tiny_dnn::recurrent_layer<T>;
+
+template <class activation_type>
+using recurrent_cell = tiny_dnn::recurrent_cell_layer<activation_type>;
```
beru: Activation functions are decoupled from layer class template parameter.
Please check the work done by karandesai-96 at #651

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
```diff
@@ -0,0 +1,104 @@
...
+    DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+    FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+    DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+    SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+   LIABILITY,
+    OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+   USE
+    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
beru: Please use short file header. This long one isn't used anymore.

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
```diff
@@ -0,0 +1,104 @@
...
+
+namespace tiny_dnn {
+
+class RecurrentCellGradOp : public core::OpKernel {
+ public:
+  explicit RecurrentCellGradOp(const core::OpKernelConstruction &context)
+    : core::OpKernel(context) {}
+
+  void compute(const core::OpKernelContext &context) override {
+    auto params = OpKernel::params_->recurrent_cell();
```
beru: You should use a reference here.

tiny_dnn/core/params/params.h
```diff
@@ -24,6 +25,7 @@ class Params {
   fully_params fully() const;
   maxpool_params &maxpool();
   global_avepool_params &global_avepool();
+  recurrent_cell_params recurrent_cell() const;
```
beru: The return value type should be referenced type.
Please check current HEAD files such as this.
https://github.com/tiny-dnn/tiny-dnn/blob/7d5ce1307a38da18641ee51dba72ecdaba0da4cf/tiny_dnn/core/params/params.h#L19

Maybe you should use `git rebase master` on your PR branch.

tiny_dnn/layers/recurrent_layer.h
prlz77: This is a dummy file, recurrent layer (the wrapper) is not done here yet! It is in a branch in my fork, it shall be removed from this PR. Thanks for noticing :+1:

tiny_dnn/core/kernels/recurrent_cell_op.h
prlz77: Ok, just followed what I saw at fully connected, indeed I was talking with @edgarriba about removing redundancy here.

tiny_dnn/layers/recurrent_cell_layer.h
prlz77: Ok, thx, again just copied from fully ^^!

tiny_dnn/tiny_dnn.h
prlz77: I know, but these activations are internal in the cell, so it is not cell + activation but cell part + activation + cell part. So I am not sure if it could be done with the new API.

prlz77: I mean, the activation is just a small part of the rnn formula, it is not applied afterwards, such as after a convolution or a fully connected.

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
prlz77: Thx!

tiny_dnn/tiny_dnn.h
beru: Oh I didn't know about that. Thank you for telling me that.

But still I think you should probablly pass an activation function layer instance via variable argument.
If you pass activation function type as a template parameter, serialization of recurrent_cell_layer will not be simple nor easy.

prlz77: :+1: I will put it as a function.

beru: I just applied `Needs discussion` label to this PR. I hope other reviewers will share their insights about designing interface.

tiny_dnn/layers/recurrent_cell_layer.h
beru: A lot of tiny things have changed since you initially created the branch, so you should really rebase to the latest origin master branch.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -23,7 +23,7 @@ inline void recurrent_cell_op_internal(const tensor_t &in_data,
                                        tensor_t &out_h,
                                        const recurrent_cell_params &params,
                                        const bool layer_parallelize) {
-  for_i(layer_parallelize, in_data.size(), [&](int sample) {
+  for_i(layer_parallelize, in_data.size(), [&](serial_size_t sample) {
```
beru: This should be `size_t`, not `serial_size_t`.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -120,7 +120,7 @@ inline void recurrent_cell_op_internal(const tensor_t &prev_out,
 
            if (params.has_bias_) {
              // vec_t& dc;
-             for (int o = r.begin(); o < r.end(); o++) {
+             for (serial_size_t o = r.begin(); o < r.end(); o++) {
```
beru: Replace `serial_size_t` to `size_t`.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -139,7 +139,7 @@ inline void recurrent_cell_op_internal(const tensor_t &prev_out,
 
            if (params.has_bias_) {
              // vec_t& db;
-             for (int o = r.begin(); o < r.end(); o++) {
+             for (serial_size_t o = r.begin(); o < r.end(); o++) {
```
beru: Replace `serial_size_t` to `size_t`.

tiny_dnn/core/params/recurrent_cell_params.h
```diff
@@ -17,13 +17,13 @@ class recurrent_cell_params : public Params {
  public:
   serial_size_t in_size_;
   serial_size_t out_size_;
-  activation_layer *activation_;
+  std::shared_ptr<activation_layer> activation_{};
   bool has_bias_;
 };
 
 // TODO(nyanp): can we do better here?
```
beru: You should not copy the comment here.

tiny_dnn/layers/recurrent_cell_layer.h
```diff
@@ -28,6 +25,7 @@ namespace tiny_dnn {
  * x(t-1)    x(t)
  *
  **/
+// TODO (@prlz77) template might cause serialization errors. To be discussed.
```
beru: This comment isn't accurate.

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
```diff
@@ -0,0 +1,61 @@
...
+
+    // initialize outputs
+    fill_tensor(prev_output_delta, float_t{0});
+    fill_tensor(prev_state_delta, float_t{0});
+
+    // call the algorithm depending on the selected engine type
+
+    const core::backend_t engine = context.engine();
+
+    if (engine == core::backend_t::internal || engine == core::backend_t::avx) {
```
beru: I wonder if this `if` statement is needed.

tiny_dnn/layers/layer.h
```diff
@@ -310,6 +313,9 @@ class layer : public node {
    *important
    **/
   virtual serial_size_t fan_out_size() const { return out_shape()[0].width_; }
+  virtual serial_size_t fan_out_size(serial_size_t) const {
+    return fan_out_size();
+  }
```
beru: This method is strange, an argument isn't used.

tiny_dnn/layers/layer.h
```diff
@@ -299,6 +299,9 @@ class layer : public node {
    *important
    **/
   virtual serial_size_t fan_in_size() const { return in_shape()[0].width_; }
+  virtual serial_size_t fan_in_size(serial_size_t) const {
+    return fan_in_size();
+  }
```
beru: This method is strange. An argument isn't used.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -0,0 +1,152 @@
...
+                               prev_h[sample][o], r.end() - r.begin(),
+                               &dW[sample][o * params.out_size_ + r.begin()]);
+           }
+
+           for (serial_size_t i = 0; i < params.in_size_; i++) {
+             vectorize::muladd(&curr_state_delta[sample][r.begin()],
+                               prev_out[sample][i], r.end() - r.begin(),
+                               &dU[sample][i * params.out_size_ + r.begin()]);
+           }
+
```
beru: You shouldn't repeatedly use `r.end() - r.begin()` inside loops. Use a local variable to hold the value.

beru: This also applies to `r.begin()`.

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
prlz77: Well we check it twice everywhere: in the layer and in the op. I was talking with @edgarriba because this is redundant. However, the if should be present at least once to throw an error if an unsupported backend is required. I removed it from the layer and kept it in the backend. I can move it to the layer again if prefered.

tiny_dnn/layers/layer.h
prlz77: To initialize reurrent weights, I had to modify the fan in and fan out, I also modified it in the layer.h.

tiny_dnn/layers/layer.h
prlz77: This is to allow initializing weights of different sizes inside a layer, needed for recurrents.

tiny_dnn/layers/recurrent_cell_layer.h
prlz77: Hmm, could you help me to write something better? I don't really know how exactly is serialization working in tiny-dnn.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -112,34 +112,37 @@ inline void recurrent_cell_op_internal(const tensor_t &prev_out,
          [&](const blocked_range &r) {
            // accumulate weight-step using delta
            // dW[c * out_size + i] += current_delta[i] * prev_out[o]
-           for (serial_size_t o = 0; o < params.out_size_; o++) {
-             vectorize::muladd(&curr_output_delta[sample][r.begin()],
-                               out_h[sample][o], r.end() - r.begin(),
-                               &dV[sample][o * params.out_size_ + r.begin()]);
+           size_t begin = r.begin();
```
edgarriba: `const` if it's not mutable

test/test_recurrent_cell_layer.h
```diff
@@ -0,0 +1,211 @@
...
+  vec_t out          = (*o[0])[0];
+  vec_t out_expected = {2.5, 2.5};  // 0+1+2+3+0.5
+
+  for (size_t i = 0; i < out_expected.size(); i++) {
+    EXPECT_NEAR(out_expected[i], out[i], 1E-4);
+  }
+}
+
+#ifdef CNN_USE_NNPACK
+TEST(recurrent_cell, forward_nnp) {
```
edgarriba: have you tested this?

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
beru: @prlz77 @edgarriba 
Why can't we use `kernels::recurrent_cell_op_internal` function in other backends?
Is there any reasons?

tiny_dnn/core/kernels/recurrent_cell_op.h
beru: @prlz77 @edgarriba 
I wonder why some people don't simply choose to fallback to internal backend implementation...

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
prlz77: You mean as a fallback? Off course we can: the if must be extended with all the possible backends if we want to throw an error for unsupported backends as it is done with other layers. Which option is the most correct? We could also drop the if but it would be incoherent with the rest of the layers.

tiny_dnn/layers/recurrent_cell_layer.h
beru: You shouldn't really use template parameter for layer classes.
If you do that, writing serialization routines will be tedious.

tiny_dnn/layers/layer.h
beru: If you could write sigle line comments above new layer methods, that should be helpful.

test/test_recurrent_cell_layer.h
prlz77: no, should it throw an error if the backend is not implemented? or it should fallback to internal?

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
beru: Welcome to this thread. https://github.com/tiny-dnn/tiny-dnn/pull/687#discussion_r121146468

prlz77: I see, so:
```c++
#ifdef CNN_USE_AVX
 if (engine == core::backend_t::avx) {
do_avx_stuff
} else
#endif
{
do_internal_stuff
}
```
Would be the recommended procedure?

tiny_dnn/layers/recurrent_cell_layer.h
```diff
@@ -53,15 +52,17 @@ class recurrent_cell_layer : public layer {
...
    * @param out_dim [in] number of elements of the output
    * @param has_bias [in] whether to include additional bias to the layer
+   * @param activation [in] activation function to be used internally
    **/
   recurrent_cell_layer(serial_size_t in_dim,
                        serial_size_t out_dim,
-                       bool has_bias          = true,
-                       backend_t backend_type = core::default_engine())
+                       bool has_bias                = true,
+                       activation_layer *activation = new tanh_layer,
```
edgarriba: is there any reason why you just the raw pointer? I see that later you convert it to a `shared_ptr`

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
beru: @prlz77 
I would like to write in that way. We can only redirect to certain layer implementation if that exists in `tiny_dnn/core/kernels` folder for a selected backend if it's enabled by compile time configuration setting (such as CNN_USE_AVX).

It's simply impossible to implement all the layer operations for every backends. So, I think it is very natural to fallback to standard internal backend routines when other options aren't available. Why do we have to add error path for no reasons?

test/test_recurrent_cell_layer.h
beru: @prlz77 
What made you write the forward_nnp test even though you don't write recurrent_cell_op kernel that uses NNPACK?

tiny_dnn/core/kernels/recurrent_cell_grad_op.h
prlz77: :+1: I agree with you

tiny_dnn/layers/recurrent_cell_layer.h
prlz77: I thought it would be easier at the user level to write rnn_cell(in, out, bias, new tanh) than rnn_cell(in, out, bias, std::shared_ptr<tanh>()) but I don't really mind chaging it for a shared pointer.

test/test_recurrent_cell_layer.h
prlz77: I was just thinking "leave it for the future", but thinking it better you are right, it makes no sense

docs/Doxyfile
```diff
@@ -1,2303 +1,18 @@
...
-IGNORE_PREFIX          =
-
-#---------------------------------------------------------------------------
-# Configuration options related to the HTML output
-#---------------------------------------------------------------------------
-
-# If the GENERATE_HTML tag is set to YES doxygen will generate HTML output
-# The default value is: YES.
-
+INPUT                  = /home/prlz77/Code/tiny-dnn/tiny_dnn \
```
Randl: Remove those

edgarriba:
@beru @Randl @bhack @karandesai-96 please review this again, want to merge this by today so that Pau can proceed with xtensor integration and next tasks

docs/Doxyfile
prlz77: I checked out the upstream file, so now it is like in upstream/master

CMakeLists.txt
```diff
@@ -58,12 +58,13 @@ option(USE_DOUBLE     "Build tiny-dnn with double precision computations"  OFF)
...
-option(BUILD_TESTS    "Set to ON to build tests"              OFF)
-option(BUILD_EXAMPLES "Set to ON to build examples"           OFF)
-option(BUILD_DOCS     "Set to ON to build documentation"      OFF)
-option(COVERALLS      "Set to ON to build with code coverage" OFF)
-option(PROFILE        "Set to ON to build with profiling"     OFF)
-option(USE_ASAN       "Set to ON to build with ASan"          OFF)
+option(BUILD_TESTS      "Set to ON to build tests"              OFF)
+option(BUILD_EXAMPLES   "Set to ON to build examples"           OFF)
+option(BUILD_DOCS       "Set to ON to build documentation"      OFF)
+option(BUILD_BENCHMARKS "Set to ON to build benchmarks"         OFF)
```
edgarriba: I'm not sure why are you including this since this was introduced by @karandesai-96 
just make sure that you didn't mess up anything :smile:

beru:
@prlz77 Commit logs are messy. Maybe you should squash your commits and rebase.

CMakeLists.txt
prlz77: The git pull --rebase added the upstream differences as if they where commits... However I did a git diff with upstream and the only differences are the recurrent code, no messing.

dirkgr:
What happened to this? The history just stops?

beru:
@dirkgr Please check #806.

PR #730, opened 2017-05-25
Add test for convolution without bias (fails)

Randl:


beru:
In the file `tiny_dnn/core/kernels/conv2d_op.h`, `Conv2dOp::compute` method, `vector::operator[]` method is used with the local variable `bias` but size checking is absent.

I would change 3rd parameter type of these functions
- `kernels::conv2d_op_internal`
- `kernels::conv2d_op_nnpack`
- `kernels::conv2d_op_avx`
from `const vec_t &` to `const vec_t *` so that we can pass `nullptr`.

`std::optional` is from C++17 and tiny-dnn uses C++11, so we can't use it.

Randl:
While using a pointer is an obvious solution, that would hurt the quality
of code IMHO. Having a pointer parameter in the middle of reference
parameters is very inviting for mistakes. However, I have no better
solution as for now.

edgarriba:
we can pass a boolean flag saying whether there is bias or just pass the whole tensor and check the size

edgarriba:
well, wait `conv_params` already has a this flag
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/params/conv_params.h#L68

Randl:
that's what is done in FC layer IIRC

edgarriba:
@Randl take into account that once we refactor with Tensor, we can get rid of most of `*_params` since e.g. all size stuff will be encoded to the Tensor members

PR #733, opened 2017-05-28
Rewrite product.h

beru:
Hi, this PR rewrites `product.h`. In this way, I think it is somehow simplified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,10 @@ if(CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
         set(EXTRA_C_FLAGS "${EXTRA_C_FLAGS} -mavx2 -mfma -march=core-avx2")
     endif(USE_AVX2 AND COMPILER_HAS_AVX2_FLAG)
 
+    if(GCC_VERSION_MAJIOR LESS 5)
+        list(APPEND CMAKE_CXX_FLAGS "-fabi-version=6")
+    endif()
+
     # include extra flags to the compiler
     # TODO: add info about those flags.
     set(EXTRA_C_FLAGS "${EXTRA_C_FLAGS} -Wall -Wpedantic -Wno-narrowing")
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,7 +1,6 @@
 include(CTest)
 enable_testing()
 
-
 include(../cmake/DownloadProject/DownloadProject.cmake)
 #set(gtest_disable_pthreads on) #TODO(randl): Windows?
 download_project(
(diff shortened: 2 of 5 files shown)


beru:
I need to update the code to support old gcc's default ABI ...
https://stackoverflow.com/a/27297025

edgarriba:
What's the status of this?

CMakeLists.txt
```diff
@@ -234,6 +234,10 @@ if(CMAKE_COMPILER_IS_GNUCXX OR MINGW OR
         set(EXTRA_C_FLAGS "${EXTRA_C_FLAGS} -mavx2 -mfma -march=core-avx2")
     endif(USE_AVX2 AND COMPILER_HAS_AVX2_FLAG)
 
+    if(GCC_VERSION_MAJIOR LESS 5)
```
kecsap: I think this is wrong: GCC_VERSION_MAJIOR -> GCC_VERSION_MAJOR

test/CMakeLists.txt
```diff
@@ -15,6 +14,10 @@ add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR})
 
 enable_testing()
 
+if(GCC_VERSION_MAJIOR LESS 5)
```
kecsap: Again: GCC_VERSION_MAJIOR -> GCC_VERSION_MAJOR

CMakeLists.txt
beru: Thank you for finding the mistake.

test/CMakeLists.txt
beru: Thank you for finding the mistake.

tiny_dnn/util/macro.h
```diff
@@ -12,3 +12,10 @@
 #if defined _WIN32 && !defined(__MINGW32__)
 #define CNN_WINDOWS
 #endif
+
+#ifdef _MSC_VER
+#define CNN_MUSTINLINE __forceinline
+#else
+#define CNN_MUSTINLINE __attribute__((always_inline)) inline
```
beru: I asked @karandesai-96 to add updated version of macros at #687.
I will rebase this branch once it'll be merged to master.

beru:
> What's the status of this?

@edgarriba It's in abandoned state.

GCC4 default ABI doesn't allow function overloading with `__m128` and `__m256` parameter types. So I guess I need to wrap the types with structs or classes just like original code does...

bhack:
Can we use -fabi-version=4 compiler option ?

bhack:
Or the is the second solution here adopted by Eigen https://stackoverflow.com/questions/5926236/overloading-conflict-with-vector-types-m128-m256-in-gcc

bhack:
Line 31 of https://eigen.tuxfamily.org/dox/SSE_2PacketMath_8h_source.html

beru:
@bhack I think changing abi version is like giving users (who use old gcc) an additional requirement.
So I need to choose the workaround which is not very pleasant from developer's point of view.

edgarriba:
@beru should we close this or are you still working on that?

beru:
@edgarriba I shall close this.

PR #739, opened 2017-05-31
Xtensor integration

Randl:
This PR integrates xtensor as Tensor storage.


beru:
Please consider splitting this pull request into multiple pull requests
- Adding recent version of xtensor (this alone makes a lot of diffs so...)
- Code fix (Maybe it's good to divide this as many as you like)
- Documentation fix (Postponing this wouldn't hurt... so that we can focus on code fix)

Randl:
@beru  I know that things got messy for a couple of reasons (mainly me fixing/commenting random things while working on code), but I don't see easy way to divide the PR (maybe I'm not enough proficient with git though).

However, basically only [three files](https://github.com/tiny-dnn/tiny-dnn/pull/739/files/616d399349e012ab5ce39dd4e4b62d1c11b297f6..82f05ff814d498df207597e8edfcbf7ab7e2fb45) have major changes: `test_tensor.h`, `tensor.h`, and `tensor_utils.h`.

edgarriba:
maybe you might want to squash your commits to have a clear view of what's happens. If it's not the case we'll merge your PR as a single commit. I'll check later today in detail everything

CMakeLists.txt
```diff
@@ -199,13 +199,13 @@ endif()
...
+# set c++ standard to c++14.
 # Note: not working on CMake 2.8. We assume that user has
-#       a compiler with C++11 support.
+#       a compiler with C++14 support.
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-message(STATUS "C++11 support has been enabled by default.")
+message(STATUS "C++14 support has been enabled by default.")
```
edgarriba: @nyanp @beru @bhack everybody okay to require c++14?

tiny_dnn/core/framework/tensor_storage.h
```diff
@@ -1,152 +0,0 @@
...
-  void resize(const std::vector<size_t> &sz) { host_data_.resize(product(sz)); }
-
-  void resize(std::initializer_list<size_t> const &sz) {
-    host_data_.resize(product(sz), U(0));
-  }
-
-  size_t size() const { return host_data_.size(); }
-
- private:
-  void toDevice() const {
```
edgarriba: we might want to reuse this codes, I don't recommend to get rid of this yet

test/test_tensor.h
```diff
@@ -25,36 +25,22 @@ TEST(tensor, constructors) {
 
   // check that t2 values has been copyied to t1
   for (size_t i = 0; i < t1.size(); ++i) {
-    EXPECT_EQ(t1.host_data()[i], float_t(2.0));
+    EXPECT_EQ(*std::next(t1.host_begin(), i), float_t(2.0));
```
edgarriba: is there any way to do this with a range based loop?

test/test_tensor.h
```diff
@@ -63,441 +49,54 @@ TEST(tensor, shape) {
...
-
-  for (size_t i = 0; i < 4; ++i) {
-    EXPECT_EQ(ptr11[i], float_t(1.0));
-  }
-
-  for (size_t i = 0; i < 4; ++i) {
-    EXPECT_EQ(ptr22[i], float_t(2.0));
-  }
+  Tensor<float_t> tensor({3, 3, 3, 3}, 2);
+  auto t_view = tensor.subView({0, 2}, {0, 1}, {0, 3}, {0, 3});
```
edgarriba: should we add the interface `channels()`?

edgarriba:
@SylvainCorlay @wolfv you might want to take a look at this PR. Feedback will be appreciated :smile:

tiny_dnn/core/framework/tensor_utils.h
```diff
@@ -11,103 +11,38 @@
...
  * @param os
  * @param tensor
  * @return
  */
-template <typename T, size_t kDim>
-inline std::ostream &operator<<(std::ostream &os,
-                                const Tensor<T, kDim> &tensor) {
-  const std::array<size_t, kDim> &shape = tensor.shape();
-  for (size_t i = 0; i < shape[0]; ++i) print_last_n_dimesions(os, tensor, i);
+template <typename T, typename S>
```
Randl: @edgarriba You mean parameters one under another?

tiny_dnn/core/framework/tensor.h
```diff
@@ -26,83 +25,67 @@ namespace tiny_dnn {
...
+   * will be of size 2x3x4x5x6. Note: tensor isn't initialized by default
    * @param shape array containing N integers, sizes of dimensions
    * @return
    */
-  explicit Tensor(const std::vector<size_t> &shape) {
-    offset_ = size_t(0);
-    size_   = product(shape);
-    std::copy(shape.begin(), shape.end(), shape_.begin());
-    storage_ptr_ = std::make_shared<TensorStorageType>(shape);
+  explicit Tensor(std::initializer_list<size_t> const &shape) {
```
Randl: When I removed vector constructor, it stopped building, so I assume we do.

tiny_dnn/core/framework/tensor.h
```diff
@@ -231,150 +188,90 @@ class Tensor {
...
-  explicit Tensor(const TensorStoragePointer storage,
-                  const size_t offset,
-                  std::initializer_list<size_t> const &shape) {
-    offset_      = offset;
-    size_        = product(shape);
-    storage_ptr_ = storage;
-    std::copy(shape.begin(), shape.end(), shape_.begin());
+  template <typename... Values>
+  Tensor<U, xt::xview<Storage &, xt::xrange<Values>...>> subView(
+    std::initializer_list<Values>... lists) {
```
Randl: Basically it can be any container. Will extend if needed.

tiny_dnn/core/framework/tensor_storage.h
Randl: If it will be reused, than probably inside `Tensor`, not different class. Luckily, git stores the history, so there is no point to keep dead (as for now) code

test/test_tensor.h
Randl: Actually this is a dirty fix that somehow made it to final version. Will think about something more elegant

test/test_tensor.h
Randl: Which is?

CMakeLists.txt
bhack: This doesn't require full c++14 coverage but just enable c++14 flag right? Probably we need to require minimal clang/msvc/gcc version

beru: > everybody okay to require c++14?

That should be decided by nyanp.

If we'll use C++14, `/README.md` file should be updated. (But this can be done in future PR)

> This doesn't require full c++14 coverage but just enable c++14 flag right? Probably we need to require minimal clang/msvc/gcc version

That's a good idea. 

And checking compiler features with ...
http://en.cppreference.com/w/cpp/compiler_support
http://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros
might help us to be ready for perceived issues from the users that try to use the library on outdated platforms.

Randl: gcc version bumped by 1, 4.8->4.9, same for clang: 3.5->3.6, VS version untouched

bhack: Have you bumped for Travis only? Is https://cmake.org/cmake/help/v3.6/variable/CMAKE_LANG_COMPILER_VERSION.html available?

nyanp: So we might need to change the first line of README.

```
tiny-dnn is a C++11 implementation of deep learning.
```

Basically I'm not keen on omitting old compilers (sometimes I am surprised at how many people in enterprise/embedded world need to use old compilers), but it's ok for me if we don't use C++14 outside of xtensor, because xtensor supports at least msvc2015 update2.

@Randl 
Is there any C++14 feature outside of xtensor?

tiny_dnn/core/framework/tensor.h
wolfv: I think this is also something to be a bit cautious about, since people might assume that the initializer list is used as the initial values of the Tensor instead of the shape.

In xtensor we have a static named constructor `xarray<double>::from_shape(std::initializer_list...)` for this purpose now.

tiny_dnn/core/framework/tensor.h
```diff
@@ -26,83 +25,67 @@ namespace tiny_dnn {
...
+   * Tensor with that shape and filling it with value. For example,
+   * given shape = {2,3,4,5,6}, value = 0 tensor will be of size 2x3x4x5x6
+   * filled with zeros
+   * @param shape  shape array containing N integers, sizes of dimensions
+   * @param value value to fill
+   */
+  explicit Tensor(std::initializer_list<size_t> const &shape, U value) {
+    storage_ = xt::xarray<U>(shape);
+    fill(value);
+  }
```
wolfv: I think we have a constructor for this, too: `xt::xarray<U>(shape, value)`.

tiny_dnn/core/framework/tensor.h
```diff
@@ -126,88 +109,62 @@ class Tensor {
...
-   * Checked version of access to indexes in tensor (throw exceptions
-   * for out-of-range error)
+   *
+   * @return the total number of elements in Tensor
+   */
+  const size_t size() const {
+    auto shape = storage_.shape();
+    return std::accumulate(shape.begin(), shape.end(), 1,
+                           std::multiplies<size_t>());
+  }
```
wolfv: `storage_.size()` does the same :)

tiny_dnn/core/framework/tensor.h
wolfv: I don't think it can be any container, as the size has to be known at compile time.

If you want "dynamic views" you can have a look at the `dynamic_view` implementation in `xstridedview.hpp` (might be useful for the future). Usage is not really well documented right now, but there are some examples in `test_dynamicview.cpp`.

tiny_dnn/core/framework/tensor_utils.h
edgarriba: oh no, I mean the print output

tiny_dnn/core/framework/tensor_storage.h
edgarriba: okay, makes sense

test/test_tensor.h
edgarriba: something that returns the number of axis, for debug might help

CMakeLists.txt
Randl: @nyanp Meanwhile we don't have anything requiring C++14 AFAIK

tiny_dnn/core/framework/tensor_utils.h
Randl: I'm using xtensor printing, so it a bit hard to alternate

tiny_dnn/core/framework/tensor.h
```diff
@@ -231,150 +188,91 @@ class Tensor {
...
+                                         const Tensor<T, S> &tensor);
 
-  /* pointer to TensorStorage */
-  TensorStoragePointer storage_ptr_;
+ private:
+  Storage storage_;
 };
 
-}  // namespace tiny_dnn
+}  // namespace tiny_dnn
```
beru: Please add a new line feed here.

tiny_dnn/core/framework/tensor.h
Randl: I don't think we care about compile-time size, but I'll check it when(if) it'll be needed.

tiny_dnn/core/framework/tensor.h
```diff
@@ -26,83 +25,67 @@ namespace tiny_dnn {
...
  public:
   /**
    * Initializes an empty tensor.
    * @return
    */
-  Tensor() {
-    offset_ = size_ = size_t(0);
-    storage_ptr_    = std::make_shared<TensorStorageType>();
-  }
+  Tensor() { storage_ = xt::xarray<U>(); }
```
beru: The decltype of `Tensor::storage_` is template parameter type `typename Storage = xt::xarray<U>`.
So assigning a specific type `xt::xarray<U>()` instance gives me a question.

But... it seems like `Tensor` class implmentors somehow want this library to be hardly dependent on xtensor library, so in that regard it may not be a big issue anymore. But then what's the reason for having the template parameter?

And I think using member initialization list is preferable.

beru:
@Randl 
> I know that things got messy for a couple of reasons (mainly me fixing/commenting random things while working on code), but I don't see easy way to divide the PR (maybe I'm not enough proficient with git though).

I'm not proficient with git but here's my guess.

In another branch, you can push a commit of adding `QuantStack/xtensor` library files (probablly the most recent released version 0.10.2) to `tiny-dnn:feat/xtensor_integration` branch without making a PR and its review process.

After then I think you should squash the related commits of this PR branch (`Randl:xtensor_integration`) into a single commit. Then you can rebase the PR branch to `tiny-dnn:feat/xtensor_integration` branch.

I haven't verified the process locally so may the force be with you while you'd be solving conflicts.

CMakeLists.txt
beru: @Randl It'd be helpful if you mention the minimum versions of supported compilers in `/README.md`.

Talking about nyanp's comment
> but it's ok for me if we don't use C++14 outside of xtensor, because xtensor supports at least msvc2015 update2.

I wonder if people would follow the mentioned wish/rule? and it reminds me of my long-ago memory that my high school classmates smoke puff on a balcony.

tiny_dnn/core/framework/tensor.h
```diff
@@ -231,150 +189,91 @@ class Tensor {
...
-   * matrix view from
-   *                                                     // offset 4.
+   * Reshape tensor
+   * @param shape new shape
    */
-  Tensor subView(std::initializer_list<size_t> const &start,
-                 std::initializer_list<size_t> const &new_shape) {
-    return subview_impl(start, new_shape);
-  }
+  void resize(const std::vector<size_t> &shape) { storage_.reshape(shape); }
```
edgarriba: This is duplicated with resize, right?

CMakeLists.txt
Randl: I think just requiring to keep currently supported compilers supported is enough

tiny_dnn/core/framework/tensor.h
beru: Will this preserve the positions of elements?

Randl: @edgarriba yes
@beru if you mean absolute indexes than no

beru: > if you mean absolute indexes than no
@Randl I checked the code.
https://github.com/QuantStack/xtensor/blob/master/include/xtensor/xcontainer.hpp#L789
The method doesn't rearrange elements which is against the intuition that the users can access the same elements with same indexes after calling the `resize` method.

tiny_dnn/core/framework/tensor.h
```diff
@@ -38,7 +38,7 @@ class Tensor {
    * Initializes an empty tensor.
    * @return
    */
-  Tensor() { storage_ = xt::xarray<U>(); }
+  Tensor() { storage_ = Storage(); }
```
beru: Is this needed?
```cpp
storage_ = Storage();
```

tiny_dnn/core/framework/tensor_utils.h
```diff
@@ -120,48 +55,62 @@ void binary_tensor_tensor_elementwise_operation(Tensor<TD, kDim> &dst,
...
 }
 
-template <typename TD, typename TS, typename F, size_t kDim>
-void unary_tensor_elementwise_operation(Tensor<TD, kDim> &dst,
-                                        const Tensor<TS, kDim> &src,
+template <typename TD, typename TDS, typename TS, typename TSS, typename F>
+void unary_tensor_elementwise_operation(Tensor<TD, TDS> &dst,
+                                        const Tensor<TS, TSS> &src,
                                         F f) {
   dst.resize(src.shape());
```
beru: @Randl I guess you should change this line to
```cpp
dst.reshape(src.shape());
```

tiny_dnn/core/framework/tensor_utils.h
```diff
@@ -120,48 +55,62 @@ void binary_tensor_tensor_elementwise_operation(Tensor<TD, kDim> &dst,
...
+          typename TDS,
+          typename TS1,
+          typename TS1S,
+          typename TS2,
+          typename F>
+void binary_tensor_scalar_operation(Tensor<TD, TDS> &dst,
+                                    const Tensor<TS1, TS1S> &src1,
                                     TS2 src2,
                                     F f) {
   dst.resize(src1.shape());
```
beru: @Randl I checked locally that all `dst.resize` in `tiny_dnn/core/framework/tensor_utils.h` should be replaced by `dst.reshape` since you removed the method in the last commit.

Randl:
@edgarriba @nyanp @bhack @beru Can we merge?

bhack:
Are there still some @wolfv open comments?

tiny_dnn/core/framework/tensor.h
```diff
@@ -26,83 +25,66 @@ namespace tiny_dnn {
...
  public:
   /**
    * Initializes an empty tensor.
    * @return
    */
-  Tensor() {
-    offset_ = size_ = size_t(0);
-    storage_ptr_    = std::make_shared<TensorStorageType>();
-  }
+  Tensor() { storage_ = Storage(); }
```
beru:
```cpp
  Tensor() { }
```

I think this is better.

tiny_dnn/core/framework/tensor.h
```diff
@@ -26,83 +25,66 @@ namespace tiny_dnn {
...
    * @return
    */
-  explicit Tensor(const std::vector<size_t> &shape) {
-    offset_ = size_t(0);
-    size_   = product(shape);
-    std::copy(shape.begin(), shape.end(), shape_.begin());
-    storage_ptr_ = std::make_shared<TensorStorageType>(shape);
+  explicit Tensor(std::initializer_list<size_t> const &shape) {
+    storage_ = Storage(shape);
   }
```
beru:
```cpp
  explicit Tensor(std::initializer_list<size_t> const &shape)
    :
    storage_(shape)
  {
  }
```

I'd write in this way.

tiny_dnn/core/framework/tensor.h
```diff
@@ -26,83 +25,66 @@ namespace tiny_dnn {
...
+   * Constructor that accepts an initializer list of shape and create a
+   * Tensor with that shape and filling it with value. For example,
+   * given shape = {2,3,4,5,6}, value = 0 tensor will be of size 2x3x4x5x6
+   * filled with zeros
+   * @param shape  shape array containing N integers, sizes of dimensions
+   * @param value value to fill
+   */
+  explicit Tensor(std::initializer_list<size_t> const &shape, U value) {
+    storage_ = Storage(shape, value);
+  }
```
beru:
```cpp
  explicit Tensor(std::initializer_list<size_t> const &shape, U value)
    :
    storage_(shape, value)
  {
  }
```

I'd write in this way.

beru:
@Randl
> Can we merge?

Please check my latest comments about using member initialization list in constructors.
I checked locally and confirmed that doing in that way passes all tests with `tiny_dnn_test` program.

I see a lot of commented out dead codes in `tiny_dnn/core/framework/tensor.h`.
Please don't forget to take care of them in future or organize remaining work to handle.

CMakeLists.txt
nyanp: @Randl Thanks, it's ok for me :)
@beru IMO it's always a trade-off and it's tough to apply strict rule. It's a good (and enough) idea to clarify supported compilers to readme.

beru: I guess we can let a C++ compiler check if non-C++11 features are used with some stub code and modified compiler options. But it seems like no one is strongly motivated to be a cpp cop here. Let it be... let it be...

SylvainCorlay:
:tada: woot :tada: !

PR #743, opened 2017-06-01
add necessary cmake options for building

zchrissirhcz:


edgarriba:
@zchrissirhcz nice work and thanks! I'd add more info: the point of having `BUILD_EXAMPLES=OFF` by default is because during the cmake configuration everything is setup to install tiny-dnn to a given directory so that you can do the following:
```
cd /path/to/tiny_dnn
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=/path/to/install ..

# it will copy all the tiny_dnn headers to the destionation
make install
```
It is also recommended when you use cmake to have a build directory. Could you update your PR explaining a briefly my comments?

Issue #740, opened 2017-05-31
code repetition


nishnik:
I was thinking that why not we create a text file with layer names and then use #define's to create functions.
Ex:
Suppose the file 'layers.inc` has:
```
TINYDNN_LAYER(elementwise_add_layer)
TINYDNN_LAYER(average_pooling_layer)
TINYDNN_LAYER(average_unpooling_layer)
..
..
.
```
Now in `serialization_functions.h`, we can just write:
```
#define CREATE_SPECIALIZE(CLASS) \
template <class Archive> \
struct specialize<Archive, \
                  tiny_dnn::CLASS, \
                  cereal::specialization::non_member_serialize> {}; 

#define TINYDNN_LAYER(Class) CREATE_SPECIALIZE(Class)
#include "layers.inc"
#undef TINYDNN_LAYER
```
And also:
```
#define IMPLEMENT_SERIALIZE(CLASS) \
template <class Archive> \
void serialize(Archive &ar, tiny_dnn::CLASS &layer) { \
  serialization_buddy::serialize(ar, layer); \
}

#define TINYDNN_LAYER(Class) IMPLEMENT_SERIALIZE(Class)
#include "layers.inc"
#undef TINYDNN_LAYER
```

And similarly in `serialization_layer_list.h`

edgarriba:
@nishnik please step into our developers room so that we an discuss it 
https://gitter.im/tiny-dnn/developers

nyanp:
@nishnik 
Thanks! I think it's a great idea :) Cc: @beru

beru:
Sounds good. However, I reduced preprocessor macros in the past #673 so I'd like to avoid adding similar mechanism.

If we remove `layer::serialize` method, we don't need to put `specialize` structs in `serialization_functions.h` file.

And adding this method in `serialization_buddy` struct
```cpp
  template <class Archive>
  static inline void serialize(Archive &ar,
                               tiny_dnn::layer &layer) {
    // TODO: implement
  }
```

and adding this template function

```cpp
template <class Archive, typename T>
typename std::enable_if< std::is_base_of<tiny_dnn::layer, T>::value >::type
serialize(Archive &ar, T &layer) {
  serialization_buddy::serialize(ar, layer);
}
```

might be able to do the job.

beru:
In the file `tiny_dnn/util/serialization_layer_list.h`. There is template function `register_layers`.
And in there, (almost?) all layers are listed manually. But the code is not smart.
[...]

PR #745, opened 2017-06-01
Code repetition

nishnik:
PR for #740 
Questions:
1. If there is any other place we can do this ?
2. Is the location of `layers.inc` right ?


edgarriba:
I guess we should close that

nishnik:
Sure

PR #754, opened 2017-06-07
Introduce parameter class to represent trainable / constant parameters of layer.

karandesai-96:
This PR is the first one to refactor underlying mechanism of representation of trainable weights of a layer. Prior to this PR, we had a class named `edge` which held a `tensor_t` of data and a `tensor_t` of its gradient. `edge` represents the incoming and outgoing data to / from a layer. The end game is to make this edge hold only references to nodes instead of actual tensors with numbers.

* `edge` was a generic class to represent data, labels, weights, biases and auxiliary tensors, but now a separate class `parameter` has been made to represent weights and biases.

* `parameter` class contains members to hold dimensions and number of filters of weights and biases, a boolean to check whether trainable or not and tensors of data and gradients, complete with their getters and setters.

* Layer constructors shall not accept weights and biases in their input type constructors, rather they shall call `layer::add_parameter` method in the constructor to add weights and biases. Same for forward and backward propagation methods.

* All tests except fully connected layer have been skipped temporarily.


karandesai-96:
@edgarriba I think `parameters_refactor` must be rebased to master to include xtensor commit as well.

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,119 @@
...
+*/
+#pragma once
+
+#include "tiny_dnn/util/util.h"
+
+namespace tiny_dnn {
+
+enum class param_type : int8_t { weight = 0x0001, bias = 0x0002 };
+
+class parameter : public std::enable_shared_from_this<parameter> {
```
edgarriba: what's the need of `std::enable_shared_from_this<parameter>`?

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,119 @@
...
+ public:
+  parameter(serial_size_t width,
+            serial_size_t height,
+            serial_size_t depth,
+            serial_size_t n_fmaps,
+            param_type type,
+            bool trainable = true)
+    : param_type_(type),
+      shape_(width, height, depth),
```
edgarriba: take into account that this properties will be part from xtensor

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,119 @@
...
+
+#include "tiny_dnn/util/util.h"
+
+namespace tiny_dnn {
+
+enum class param_type : int8_t { weight = 0x0001, bias = 0x0002 };
+
+class parameter : public std::enable_shared_from_this<parameter> {
+ public:
+  parameter(serial_size_t width,
```
edgarriba: why just 4D? see xtensor constructor

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,119 @@
...
+    n_fmaps_       = n_fmaps;
+  }
+
+  bool is_trainable() { return trainable_; }
+
+  void set_trainable() { trainable_ = true; }
+
+  void freeze_trainable() { trainable_ = false; }
+
+  vec_t *get_data() { return &data_; }
```
edgarriba: follow convention https://github.com/tiny-dnn/tiny-dnn/blob/7d5ce1307a38da18641ee51dba72ecdaba0da4cf/tiny_dnn/layers/layer.h#L167
`get_data()` -> `data()`

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,119 @@
...
+
+  const vec_t *get_data() const { return &data_; }
+
+  void set_data(const vec_t &data) { data_ = data; }
+
+  void set_data(const vec_t *data) { data_ = *data; }
+
+  tensor_t *get_grad() { return &grad_; }
+
+  const tensor_t *get_grad() const { return &grad_; }
```
edgarriba: follow convention https://github.com/tiny-dnn/tiny-dnn/blob/7d5ce1307a38da18641ee51dba72ecdaba0da4cf/tiny_dnn/layers/layer.h#L167
get_grad() -> grad()

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,119 @@
...
+  float_t *grad_at(size_t sample, size_t i) { return &grad_[sample][i]; }
+
+ private:
+  param_type param_type_;
+  shape3d shape_;
+  size_t n_fmaps_;
+  bool trainable_;
+
+  vec_t data_;
+  tensor_t grad_;
```
edgarriba: `data_` and `grad_` should be of type `Tensor`

edgarriba:
@karandesai-96 I would focus this PR just to design the parameter class. Integration can come later.

tiny_dnn/parameter.h
karandesai-96: I had an idea of optimizer catching hold of all the parameters of a network while training, again similar to how pytorch performs updates ( _the `optim.step()` way_ ).

karandesai-96:
@edgarriba I shall have a look at newly integrated Tensor class and make modifications here.

test/test.cpp
```diff
@@ -13,47 +13,47 @@
...
+//#include "test_quantized_fully_connected_layer.h"
+//#endif  // CNN_USE_GEMMLOOP
+//
+//#ifdef CNN_USE_CAFFE_CONVERTER
+//#include "test_caffe_converter.h"
+//#endif  // CNN_USE_CAFFE_CONVERTER
+//
+//#ifdef DNN_USE_IMAGE_API
+//#include "test_image.h"
+//#endif  // DNN_USE_IMAGE_API
```
Randl: Use `/* */` instead. Also why have you commented out all the tests? Does actually not a single things works now?

tiny_dnn/layers/layer.h
```diff
@@ -708,6 +687,53 @@ class layer : public node {
     }
   }
 
+  ////////////////////////////////////////////////////////////////////////////
+  // getters setters for parameters
```
Randl: Please, use Doxygen-style comments

test/test.cpp
karandesai-96: To make things easier, I deleted old `weights()` methods, which broke all the tests. Since things work out well, I will do a little cleanup.

karandesai-96: So I commented all tests but fully_connected ones, and did my work in order to make them pass alone.

tiny_dnn/layers/layer.h
karandesai-96: This is not a function docstring, but enclosing a set of functions, similar to this line. https://github.com/tiny-dnn/tiny-dnn/pull/754/files/d7e647d109f2d40af00f92bbfa0f0977048ce9c4#diff-a1617ac70e8f704e6996e605b9f4b889R374

Shall I modify these as well ?

nyanp:
@karandesai-96
Could you please tell everyone to the merit of this refactoring PR? What have we got to be able to do by this PR? (Maybe is was discussed in gitter, but it is hard to trace it)

karandesai-96:
@nyanp: I had this conversation on Hangouts..

Currently our `edge` holds the weights / biases as well. In frameworks like tf and pytorch, edges are simply connections with references to previous and next nodes, while they hold no "memory" (continguous arrays of numbers). I am making the edges memory-less and introducing a "parameter" class, inspired by Pytorch. Since it is a different class, it might help in overloading the operators `<<` and `>>` to load pretrained weights.

Pointers to newly constructed parameters will be held by a specific layer once it is created, so `fprop` and `bprop` methods shall only have back and forth flow of data. I believe this reduces flow of tensors in the computational graph. This image summarizes it better:

Layer in tensorflow has its own parameters within.
tensorflow layer

And the difference between two configurations:
different configurations

Since our operators are "kernels", they already behave like those in the diagram to the right. As a luxury we are having complete layer classes wrapped over basic kernels so we can design them as per left diagram.

karandesai-96:
I have made a bunch of modifications and scoped out a lot from this PR. I will make smaller PRs to `feat/parameters_refactor` now.

Closing this in favor of #774

nyanp:
@karandesai-96 
Thanks! It's meaningful for future developers to know what the PR is for.

edgarriba:
@karandesai-96 any progress on that?

PR #757, opened 2017-06-08
Tensor integration for fully connected

Randl:


tiny_dnn/core/kernels/fully_connected_grad_op.h
```diff
@@ -22,30 +22,39 @@ class FullyConnectedGradOp : public core::OpKernel {
...
+    const Tensor<float_t> prev_out_t(context.input(0)),
+      weights_t(context.input(1));
+    Tensor<float_t> weights_grads(context.input_grad(1)),
+      bias_grads = params.has_bias_ ? Tensor<float_t>(context.input_grad(2))
+                                    : Tensor<float_t>(),
+      prev_delta_t(context.input_grad(0)), curr_delta_t(context.output_grad(0));
 
     // initialize outputs
-    fill_tensor(prev_delta, float_t{0});
+    prev_delta_t.fill(float_t{0});
```
edgarriba: Why the sufix `_t`?

tiny_dnn/core/kernels/fully_connected_op.h
```diff
@@ -29,25 +29,35 @@ class FullyConnectedOp : public core::OpKernel {
     const tensor_t *bias    = params.has_bias_ ? &context.input(2) : nullptr;
     tensor_t &out_data      = context.output(0);
 
+    // TODO(Randl): Remove once layers forward and backward by themself.
+    const Tensor<float_t> in_data_t(context.input(0)), weights_t(context.input(1)), bias_t = params.has_bias_ ? Tensor<float_t>(context.input(2)) : Tensor<float_t>();
```
edgarriba: You might split this into three lines of code

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -12,59 +12,67 @@
...
+                                        const Tensor<float_t, S2> &weights,
+                                        const Tensor<float_t, S3> &bias,
+                                        Tensor<float_t, S4> &out_data,
                                         const fully_params &params,
                                         const bool layer_parallelize) {
-  for_i(layer_parallelize, in_data.size(), [&](int sample) {
-    const vec_t &in = in_data[sample];
-    vec_t &out      = out_data[sample];
-
+  for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
```
edgarriba: Maybe not now but this is just a matrix multiplication, we could take advantage​ here of xtensor-blas

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -12,59 +12,67 @@
...
       for (serial_size_t c = 0; c < params.in_size_; c++) {
-        out[i] += W[c * params.out_size_ + i] * in[c];
+        out_data.host_at(sample, i) +=
+          weights.host_at(0, c * params.out_size_ + i) *
+          in_data.host_at(sample, c);
       }
 
       if (params.has_bias_) {
-        out[i] += bias[i];
+        out_data.host_at(sample, i) += bias.host_at(0, i);
```
edgarriba: You can just use tensor sum op

tiny_dnn/core/kernels/fully_connected_grad_op.h
Randl: to not overlap with existing variables, will be removed later.

tiny_dnn/core/kernels/fully_connected_op_internal.h
Randl: Sure

bhack:
@SylvainCorlay Here the real first layer integration

tiny_dnn/core/kernels/fully_connected_op_internal.h
bhack: @edgarriba Actually seems that xtensor-blas rely on openblas right?

SylvainCorlay: Yes xtensor-blas requires a BLAS implementation (netlib, openblas or mkl)...

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -12,59 +12,87 @@
...
       // propagate delta to previous layer
       // prev_delta[c] += current_delta[r] * W_[c * out_size_ + r]
-      prev_delta[sample][c] += vectorize::dot(
-        &curr_delta[sample][0], &W[c * params.out_size_], params.out_size_);
+      prev_delta.host_at(sample, c) += vectorize::dot(
+        curr_delta.host_iter(sample, 0),
+        weigths.host_iter(0, c * params.out_size_), params.out_size_);
     }
 
     for_(layer_parallelize, 0, size_t(params.out_size_),
```
edgarriba: params structure doesn't make sense now since the info is encoded to the tensor itself

tiny_dnn/util/product.h
```diff
@@ -286,7 +286,16 @@ inline void double_avx::store<std::false_type>(double *px, const __m256d &v) {
...
-// generic dot-product
+/**
+ * generic dot-product
+ * @tparam T
+ * @tparam f1_aligned
+ * @tparam f2_aligned
+ * @param f1
+ * @param f2
+ * @param size
+ * @return
```
beru: Is it very meaningful to add empty doxygen style comments?

Randl: I've started from adding description, IMHO it's better than nothing.

beru: I wish writing doxygen comments wouldn't be mandatory....

Randl: It isn't, but they're really helpful for reader (at least for me).

Randl:
> Using `std::next`

you're right, `std::next` is unnecessary, I've changed it

> I prefer to use plain pointer type when dealing with raw buffer.

Agreed, but that might cause problems if, for example, view is used as input.

bhack:
I think that we could be in the same case of https://stackoverflow.com/questions/31126823/why-use-stdnext-instead-of-adding-an-integer-to-the-pointer if we will use a view.

beru:
> Agreed, but that might cause problems if, for example, view is used as input.

In that case, I'd add a stride offset to a raw pointer variable.
Capsuling the memory layout detail in class method is generally good but not good for optimization if redundant operation takes place in the method. We should reduce the number of operations in inner loops as much as possible if we want to optimize the programs.

Randl:
> Capsuling the memory layout detail in class method is generally good but not good for optimization if redundant operation takes place in the method. 

As mentioned in the one of the answers from SO, `std:next`is smart enough to optimize if possible. It is used, for example, all over standard library and boost, which wouldn't be the case if there were significant overhead.

beru:
> As mentioned in the one of the answers from SO, std:nextis smart enough to optimize if possible. It is used, for example, all over standard library and boost, which wouldn't be the case if there were significant overhead.

It depends on how iterator class method operates. There are cases when operations should be excluded from inner loops but compilers cannot optimize them out because they can't deduct meanings of algorithms unlike humans.

Using intrinsics instructions is somewhat close to instruction level coding, so mixing C++ style code that abstracts actual operations blurs the implementation.

CMakeLists.txt
```diff
@@ -142,12 +142,17 @@ endif()
 
 # Find NNPACK: Acceleration package for neural networks on multi-core CPUs
 find_package(NNPACK QUIET)
+SET(NNPACK_FOUND ON)
+SET(NNPACK_INCLUDE_DIR "/home/chaimbaskin/evgen/NNPACK/include")
```
edgarriba: remove this?

tiny_dnn/core/kernels/fully_connected_op.h
```diff
@@ -24,30 +24,38 @@ class FullyConnectedOp : public core::OpKernel {
...
+                                           params.has_bias_,
+                                           context.parallelize());
+
+      // TODO(Randl): Remove once layers forward and backward by themself.
+      context.output(0) = out_data.toTensor();
     } else if (engine == core::backend_t::nnpack) {
-      kernels::fully_connected_op_nnpack(
-        in_data, W[0], params.has_bias_ ? (*bias)[0] : vec_t(), out_data,
-        params, context.parallelize());
+      kernels::fully_connected_op_avx(in_data, weights, bias, out_data,
```
beru: In here, this should be `fully_connected_op_nnpack`, not `fully_connected_op_avx`.

edgarriba: right!

beru:
@Randl You should simply use pointers instead of iterators in compute-intensive places, otherwise it'll hard to control generated instructions...

Randl:
@beru Using pointers given strides and views would result in full duplicate of `std::next` functionality. At least I can't see another way to do that.

beru:
@Randl If I'm getting things right, in fully_connected_op routines, we iterate through 2D arrays (in_data, weights, bias). And if we're talking about line strides, I think it's still OK to delegate the task to Tensor implementations. (though overhead should be minimized.)

And there is no point in using iterator class inside inner loops since we don't care about element strides in existing computation (`avx_fully_connected_forward_kernel`). It is written to deal with contiguous buffer. At least last dimension of tensor shouldn't have spaces between elements.

Randl:
@beru Could you elaborate, where exactly should iterator be replaced with pointer?

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,302 @@ namespace kernels {
...
         -1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0, 0,
       };
       __m256i imask =
         _mm256_loadu_si256((__m256i const *)(mask_src + 8 - nremains));
-      for_i(layer_parallelize, in_data.size(), [&](int sample) {
-        const auto &in = in_data[sample];
-        auto &out      = out_data[sample];
+      for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
+        auto in  = in_data.host_iter(sample, 0);
+        auto out = out_data.host_iter(sample, 0);
```
beru: These local variables should be pointers instead of iterators.
You may use `Tensor::host_pointer` method. (Though I wish we can skip creating iterator instance by not using `host_iter` method in there...)

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,302 @@ namespace kernels {
...
+          _mm256_maskstore_ps(&*std::next(out, 8 * nblocks), imask, sum);
         }
       });
     } else {
-      for_i(layer_parallelize, in_data.size(), [&](int sample) {
-        const auto &in = in_data[sample];
-        auto &out      = out_data[sample];
+      for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
+        auto in  = in_data.host_iter(sample, 0);
+        auto out = out_data.host_iter(sample, 0);
```
beru: These `in` and `out` local variables should be pointers instead of iterators.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,302 @@ namespace kernels {
...
       });
     }
   } else {
-    for_i(layer_parallelize, in_data.size(), [&](int sample) {
-      const auto &in = in_data[sample];
-      auto &out      = out_data[sample];
-      for (serial_size_t i = 0; i < params.out_size_; i++) {
+    for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
+      auto in  = in_data.host_iter(sample, 0);
+      auto out = out_data.host_iter(sample, 0);
```
beru: These variables should be pointers instead of iterators.

beru:
Newly added functions that take iterator type arguments inside `tiny_dnn/util/product.h` totally workaround existings TMP implementations. In that way, we can't get benefits from vectorized types.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -54,46 +54,46 @@ inline void avx_fully_connected_forward_kernel(const Tensor<float, S1> &in_data,
           _mm256_maskstore_ps(&*std::next(out, 8 * nblocks), imask, b);
         }
         for (serial_size_t c = 0; c < in_size; c++) {
-          auto in_val = _mm256_set1_ps(*std::next(in, c));
-          auto pW     = W.host_iter(0, c * out_size);
+          auto in_val = _mm256_set1_ps(in[c]);
+          auto pW     = W.host_pointer(0, c * out_size);
```
beru: Please get a pointer from W tensor outside of this loop.
We have no benefits from using host_pointer method inside loop like this.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -54,46 +54,46 @@ inline void avx_fully_connected_forward_kernel(const Tensor<float, S1> &in_data,
...
       });
     } else {
       for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
-        auto in  = in_data.host_iter(sample, 0);
-        auto out = out_data.host_iter(sample, 0);
+        auto in  = in_data.host_pointer(sample, 0);
+        auto out = out_data.host_pointer(sample, 0);
         for (size_t i = 0; i < nblocks; ++i) {
-          __m256 b = _mm256_loadu_ps(&*bias.host_iter(0, 8 * i));
+          __m256 b = _mm256_loadu_ps(bias.host_pointer(0, 8 * i));
```
beru: Please get a pointer from bias Tensor outside of this loop.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -42,8 +42,8 @@ inline void avx_fully_connected_forward_kernel(const Tensor<float, S1> &in_data,
       __m256i imask =
         _mm256_loadu_si256((__m256i const *)(mask_src + 8 - nremains));
       for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
-        auto in  = in_data.host_iter(sample, 0);
-        auto out = out_data.host_iter(sample, 0);
+        auto in  = in_data.host_pointer(sample, 0);
+        auto out = out_data.host_pointer(sample, 0);
         {
           for (size_t i = 0; i < nblocks; ++i) {
             __m256 b = _mm256_loadu_ps(&*std::next(bias.host_begin(), 8 * i));
```
beru: You should get a pointer from bias Tensor outside of this loop.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -54,46 +54,46 @@ inline void avx_fully_connected_forward_kernel(const Tensor<float, S1> &in_data,
           _mm256_maskstore_ps(&*std::next(out, 8 * nblocks), imask, b);
```
beru: You can simply use oprator [] with local pointer variable `out`.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -54,46 +54,46 @@ inline void avx_fully_connected_forward_kernel(const Tensor<float, S1> &in_data,
...
-          __m256 w   = _mm256_maskload_ps(&*std::next(pW, 8 * nblocks), imask);
+          __m256 sum = _mm256_maskload_ps(&out[8 * nblocks], imask);
+          __m256 w   = _mm256_maskload_ps(pW[8 * nblocks], imask);
           sum        = madd256_ps(w, in_val, sum);
-          _mm256_maskstore_ps(&*std::next(out, 8 * nblocks), imask, sum);
+          _mm256_maskstore_ps(&out[8 * nblocks], imask, sum);
         }
       });
     } else {
       for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
```
beru: Maybe we can use `size_t sample` instead of `int sample`.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -42,8 +42,8 @@ inline void avx_fully_connected_forward_kernel(const Tensor<float, S1> &in_data,
       __m256i imask =
         _mm256_loadu_si256((__m256i const *)(mask_src + 8 - nremains));
       for_i(layer_parallelize, in_data.shape()[0], [&](int sample) {
```
beru: Prolly we can use `size_t sample` instead of `int sample` here.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,302 @@ namespace kernels {
...
           if (nblocks & 1) {
             __m256 sum0 = _mm256_loadu_ps(&out[nblocks / 2 * 16]);
-            __m256 w0   = _mm256_loadu_ps(pW + nblocks / 2 * 16);
+            __m256 w0   = _mm256_loadu_ps(&pW[nblocks / 2 * 16]);
             sum0        = madd256_ps(w0, in_val, sum0);
             _mm256_storeu_ps(&out[nblocks / 2 * 16], sum0);
           }
           __m256 sum = _mm256_maskload_ps(&out[8 * nblocks], imask);
-          __m256 w   = _mm256_maskload_ps(pW + 8 * nblocks, imask);
+          __m256 w   = _mm256_maskload_ps(pW[8 * nblocks], imask);
```
beru: You need to insert address-of operator `&` here.
```cpp
          __m256 w   = _mm256_maskload_ps(&pW[8 * nblocks], imask);
```

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,302 @@ namespace kernels {
...
-          _mm256_storeu_ps(&out[8 * i], b);
+          __m256 b = _mm256_loadu_ps(bias.host_pointer(0, 8 * i));
+          _mm256_storeu_ps(&*std::next(out, 8 * i), b);
         }
-        for (serial_size_t c = 0; c < params.in_size_; c++) {
-          auto in_val     = _mm256_set1_ps(in[c]);
-          const float *pW = &W[c * params.out_size_];
+        for (serial_size_t c = 0; c < in_size; c++) {
+          auto in_val = _mm256_set1_ps(*std::next(in, c));
+          auto pW     = W.host_iter(0, c * out_size);
```
beru: You need to use `host_pointer` method here.
```cpp
          auto pW     = W.host_pointer(0, c * out_size);
```

beru:
I see a warning with `get_range` function inside `tiny_dnn/core/framework/tensor.h` file.
Is the implementation alright?

```cpp
  template <typename S>
  xt::xrange<S> get_range(std::initializer_list<S> list) const {
    // if (*(list.begin()) == -1) return xt::all();
    // if(*(list.begin())==*(list.begin()+1)) return *(list.begin());
    if (list.size() == 2) {
      return xt::range(*(list.begin()), *(list.begin() + 1));
    } else {
      // return xt::range(*(list.begin()), *(list.begin() + 1),
      //               *(list.begin() + 2));
    }
  }
```

Randl:
> Is the implementation alright?

There are couple of things I haven't implemented yet. Luckily they aren't necessary as for now.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,304 @@ namespace kernels {
...
-        auto &out      = out_data[sample];
+      for_i(layer_parallelize, in_data.shape()[0], [&](size_t sample) {
+        auto in  = in_data.host_pointer(sample, 0);
+        auto out = out_data.host_pointer(sample, 0);
         {
           for (size_t i = 0; i < nblocks; ++i) {
-            __m256 b = _mm256_loadu_ps(&bias[8 * i]);
-            _mm256_storeu_ps(&out[8 * i], b);
+            __m256 b = _mm256_loadu_ps(&*std::next(bias.host_begin(), 8 * i));
+            _mm256_storeu_ps(&*std::next(out, 8 * i), b);
```
beru: Plase simply use pointers in this loop. We should use iterator only where it should be used.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,304 @@ namespace kernels {
...
-        auto &out      = out_data[sample];
+      for_i(layer_parallelize, in_data.shape()[0], [&](size_t sample) {
+        auto in        = in_data.host_pointer(sample, 0);
+        auto out       = out_data.host_pointer(sample, 0);
+        auto b_pointer = bias.host_pointer(0, 0);
         for (size_t i = 0; i < nblocks; ++i) {
-          __m256 b = _mm256_loadu_ps(&bias[8 * i]);
-          _mm256_storeu_ps(&out[8 * i], b);
+          __m256 b = _mm256_loadu_ps(&b_pointer[8 * i]);
+          _mm256_storeu_ps(&*std::next(out, 8 * i), b);
```
beru: You can simply write as it was before.
```cpp
          _mm256_storeu_ps(&out[8 * i], b);
```

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,304 @@ namespace kernels {
...
         }
-        for (serial_size_t c = 0; c < params.in_size_; c++) {
-          auto in_val     = _mm256_set1_ps(in[c]);
-          const float *pW = &W[c * params.out_size_];
+        auto W_pointer = W.host_pointer(0, 0);
+        for (serial_size_t c = 0; c < in_size; c++) {
+          auto in_val = _mm256_set1_ps(*std::next(in, c));
```
beru: You can simply write
```cpp
          auto in_val     = _mm256_set1_ps(in[c]);
```

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,304 @@ namespace kernels {
...
           }
         }
       });
     }
   } else {
-    for_i(layer_parallelize, in_data.size(), [&](int sample) {
-      const auto &in = in_data[sample];
-      auto &out      = out_data[sample];
-      for (serial_size_t i = 0; i < params.out_size_; i++) {
+    for_i(layer_parallelize, in_data.shape()[0], [&](size_t sample) {
```
beru: Inside this loop should be updated.

tiny_dnn/core/kernels/fully_connected_op_avx.h
```diff
@@ -14,225 +14,304 @@ namespace kernels {
...
+      });
     }
   } else {
-    for (serial_size_t sample = 0; sample < prev_out.size(); sample++) {
-      auto &prev_delta2 = prev_delta[sample];
-      auto &curr_delta2 = curr_delta[sample];
-      auto &prev_out2   = prev_out[sample];
-      auto &dW2         = dW[sample];
-      for (serial_size_t c = 0; c < params.in_size_; c++) {
+    for (serial_size_t sample = 0; sample < prev_out.shape()[0]; sample++) {
```
beru: You can use local variables inside this loop just like it used to be.
There's no point in repeating costly operations when it can be avoided.

tiny_dnn/core/kernels/fully_connected_op_internal.h
```diff
@@ -12,62 +12,90 @@
...
+      out_data.host_at(sample, i) = float_t{0};
+      for (serial_size_t c = 0; c < in_size; c++) {
+        out_data.host_at(sample, i) +=
+          weights.host_at(0, c * out_size + i) * in_data.host_at(sample, c);
       }
 
-      if (params.has_bias_) {
-        out[i] += bias[i];
+      if (has_bias) {
+        out_data.host_at(sample, i) += bias.host_at(0, i);
```
edgarriba: Can we just check the size of bias ? When no bias we are receiving an empty Tensor right?

Randl:
@beru @bhack Is there some problem with VS? Can you take a look?

Randl:
Looks like this one of those glitches since nothing changed between two last commits except formatting

CMakeLists.txt
```diff
@@ -146,8 +146,9 @@ if(USE_NNPACK AND NNPACK_FOUND)
     message(STATUS "Found NNPACK: ${NNPACK_INCLUDE_DIR}")
     add_definitions(-DCNN_USE_NNPACK)
     include_directories(SYSTEM ${NNPACK_INCLUDE_DIR})
-    include_directories(SYSTEM ${NNPACK_INCLUDE_DIR}/../third-party/pthreadpool/include)
+    include_directories(SYSTEM ${NNPACK_INCLUDE_DIR}/../deps/pthreadpool/include)
```
edgarriba: You sure about this path?

tiny_dnn/core/kernels/fully_connected_grad_op.h
```diff
@@ -22,30 +22,40 @@ class FullyConnectedGradOp : public core::OpKernel {
...
     // call the algorithm depending on the selected engine type
 
     const core::backend_t engine = context.engine();
 
     if (engine == core::backend_t::internal) {
       kernels::fully_connected_op_internal(
-        prev_out, W[0], dW, params.has_bias_ ? *db : dummy, curr_delta,
-        prev_delta, params, context.parallelize());
+        prev_out, weights, weights_grads, bias_grads, curr_delta, prev_delta,
+        params.has_bias_, context.parallelize());
```
edgarriba: Has bias should be removed

CMakeLists.txt
Randl: yeah, used by caffe2 too: https://github.com/caffe2/caffe2/blob/master/cmake/External/nnpack.cmake

bhack:
Tests are failing

edgarriba:
cool!

PR #759, opened 2017-06-12
Convolutional layers Tensor integration 

Randl:


tiny_dnn/core/kernels/conv2d_grad_op.h
```diff
@@ -22,28 +22,36 @@ class Conv2dGradOp : public core::OpKernel {
...
+      weights_t(context.input(1));
+    Tensor<float_t> weights_grads_t(context.input_grad(1));
+    Tensor<float_t> bias_grads_t(context.input_grad(2));
+    Tensor<float_t> prev_delta_t(context.input_grad(0));
+    Tensor<float_t> curr_delta_t(context.output_grad(0));
 
-    // initalize outputs
-    fill_tensor(prev_delta, float_t{0});
+    // initialize outputs
+    prev_delta_t.fill(float_t{0});
```
edgarriba: use same convention for `float_t{0}` as in https://github.com/tiny-dnn/tiny-dnn/pull/759/files#diff-fc64083110af88933c8ad786595bf36dR33

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -10,103 +10,114 @@
...
-               vectorize::add(bias[o], out_area, pa);
-             }
-           }
-         }
-       },
-       0);
+  for_(
+    parallelize, 0, in_data.shape()[0],
+    [&](const blocked_range &r) {
+      size_t out_area           = params.out.area();
```
edgarriba: use as much as possible tensor properties for this

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -10,103 +10,114 @@
...
+      serial_size_t iw          = params.in_padded.width_;
+      serial_size_t id          = params.in.depth_;
+      serial_size_t ow          = params.out.width_;
+      serial_size_t oh          = params.out.height_;
+      serial_size_t od          = params.out.depth_;
+      serial_size_t kw          = params.weight.width_;
+      serial_size_t kh          = params.weight.height_;
+      serial_size_t elem_stride = params.w_stride;
+      serial_size_t line_stride = iw * params.h_stride;
+      for (size_t sample = r.begin(); sample < r.end(); sample++) {
```
edgarriba: it's okay for now but this can be simplified just just tensor accessors. I'll handle it later, just add a todo for me :smile:

tiny_dnn/core/framework/tensor.h
```diff
@@ -278,17 +279,17 @@ auto host_data() {
 
   Tensor operator[](size_t index) { return Tensor(storage_[index]); }
 
-  template <typename S>
-  xt::xrange<S> get_range(std::initializer_list<const S> list) const {
-    // if (*(list.begin()) == -1) return xt::all();
-    // if(*(list.begin())==*(list.begin()+1)) return *(list.begin());
+  /*template <typename S>
```
edgarriba: What was wrong with lists?

Randl: They are really problematic, all kind of type problems, you can check compilation errors. Also I want to use, for example, strided views in future and that would be either overcomplicated or impossible

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -72,41 +80,51 @@ inline void conv2d_op_internal(const tensor_t &in_data,
...
-                        tensor_t &curr_delta,
-                        tensor_t &prev_delta,
-                        const core::conv_params &params,
-                        const bool parallelize) {
-  typedef typename vec_t::value_type float_t;
-
-  for_i(parallelize, prev_out.size(), [&](int sample) {
+    const float_t *weight_begin = weigths.host_pointer(0, 0);
     // propagate delta to previous layer
     for (serial_size_t inc = 0; inc < params.in.depth_; inc++) {
```
edgarriba: use tensor shape

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -72,41 +80,51 @@ inline void conv2d_op_internal(const tensor_t &in_data,
...
 
         serial_size_t idx = 0;
-        idx               = params.in.depth_ * outc + inc;
-        idx               = params.weight.get_index(0, 0, idx);
-        const float_t *pw = &W[idx];
 
-        idx                       = params.out.get_index(0, 0, outc);
-        const float_t *pdelta_src = &curr_delta[sample][idx];
+        idx = params.in.depth_ * outc + inc;
+        idx = params.weight.get_index(0, 0, idx);
```
edgarriba: You can also use tensor accessors. Add a todo here, I can fix that

tiny_dnn/core/kernels/conv2d_op_opencl.h
```diff
@@ -46,22 +46,26 @@ class Conv2dOpenCLForwardOp : public core::OpKernel {
...
-
-      auto dev_W =
-        CLCudaAPI::Buffer<float_t>(ctx, queue, W[0].begin(), W[0].end());
+      auto dev_W = CLCudaAPI::Buffer<float_t>(
+        ctx, queue, weights_t.host_pbegin(), weights_t.host_pend());
 
-      auto dev_bias =
-        CLCudaAPI::Buffer<float_t>(ctx, queue, bias[0].begin(), bias[0].end());
+      auto dev_bias = CLCudaAPI::Buffer<float_t>(
+        ctx, queue, bias_t.host_pbegin(), bias_t.host_pbegin());
```
edgarriba: brief comment: we should start thinking about how are going to share memory between devices

tiny_dnn/core/kernels/conv2d_op_internal.h
Randl: that requires change in representation of weights from 1D to 3D, which I believe is ok performance-wise (@beru ? ) since it remains contiguous

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -375,6 +390,7 @@ void avx_conv2d_5x5_kernel(const core::conv_params &params,
         size_t cnt  = (out_area - headSize) / 8;
         double *pa2 = pa + headSize;
         for (size_t i = 0; i < cnt; ++i) {
+          // TODO: segfault here even though adresses are ok
           _mm256_store_pd(&pa2[i * 8 + 0], b2);
```
Randl: I still haven't figured out why segfault happens here. Pointer seems to be legal.

Randl: @beru any ideas?

tiny_dnn/core/framework/tensor.h
edgarriba: sounds good

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -46,15 +46,18 @@ void avx_conv2d_5x5_kernel(const core::conv_params &params,
...
   const __m128 y_bias_scale = _mm_set_ss(bias_scale);
   if (out.height_ == 1 && out.width_ == 1) {
-    const float *pw = (const float *)&W[0];
     for (serial_size_t o = 0; o < out.depth_; ++o) {
       __m256 sum0     = _mm256_setzero_ps();
       __m256 sum1     = _mm256_setzero_ps();
       __m256 sum2     = _mm256_setzero_ps();
       __m128 sum3     = _mm_setzero_ps();
-      const float *pi = (const float *)&in[0];
+      const float *pi = in.host_pbegin();
```
beru: This change increases execution time...

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -78,17 +81,17 @@ void avx_conv2d_5x5_kernel(const core::conv_params &params,
...
+      _mm_store_ss(&a_begin[o], _mm_add_ss(hsum, b));
     }
   } else {
     const serial_size_t nblocks = out.width_ / 4;
     for (serial_size_t o = 0; o < out.depth_; ++o, oidx += out_area) {
-      float *pa = &a[oidx];
+      float *pa = &a_begin[oidx];
       // init to bias value
-      float b = bias[o] * bias_scale;
+      float b = bias.host_at(0, o) * bias_scale;
```
beru: Why in this way??

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -109,11 +112,13 @@ void avx_conv2d_5x5_kernel(const core::conv_params &params,
           pa[i] = b;
         }
       }
+      const float *weights_begin = W.host_pbegin();
+      const float *in_begin      = in.host_pbegin();
```
beru: These pointer variables can be calculated outside of depth loop.

tiny_dnn/core/kernels/conv2d_op_avx.h
Randl: what do you mean?

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -313,55 +324,59 @@ void avx_conv2d_5x5_kernel(const core::conv_params &params,
...
       b            = madd128_sd(b, y_bias_scale, sum6);
-      _mm_store_sd(&a[o], _mm_add_sd(hsum, b));
+      _mm_store_sd(&a_begin[o], _mm_add_sd(hsum, b));
     }
   } else {
     for (serial_size_t o = 0; o < out.depth_; ++o, oidx += out_area) {
-      double *pa = &a[oidx];
-      double b   = bias[o] * bias_scale;
+      double *pa = &a_begin[oidx];
+      double b   = bias.host_at(0, o) * bias_scale;
```
beru: `host_at` method inside loop isn't good for performance and it can be easily avoided in this case.

tiny_dnn/core/kernels/conv2d_op_avx.h
```diff
@@ -21,12 +21,12 @@ namespace kernels {
 #ifdef CNN_USE_AVX
 
 // float ver
-template <typename Allocator>
+template <typename S1, typename S2, typename S3, typename S4>
```
beru: Do we really need to support different storage types for all tensor arguments?
gdb backtrace is hardly readable....

Randl: since view and tensor itself are different times, seems like we do

tiny_dnn/core/kernels/conv2d_op_avx.h
beru: In `conv2d_op_avx.h` file, replace `_mm256_store_ps` to `_mm256_storeu_ps` and you also need to change `_mm256_store_pd` to `_mm256_storeu_pd` in the same file. I locally confirmed these changes will fix the segfault with `tiny_dnn_test` program.

It can be that `xtensor` library doesn't allocate memory with enough alignment (32bytes for AVX, ideally it should be 64bytes though).

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -28,42 +38,40 @@ inline void conv2d_op_internal(const tensor_t &in_data,
          serial_size_t kh          = params.weight.height_;
          serial_size_t elem_stride = params.w_stride;
          serial_size_t line_stride = iw * params.h_stride;
+         // TODO(edgarriba): replace with  tensor accessors
          for (size_t sample = r.begin(); sample < r.end(); sample++) {
-           const vec_t &in = in_data[sample];
-           vec_t &a        = out_data[sample];
+           float_t *out_data_begin      = out_data.host_pointer(sample, 0);
+           const float_t *in_data_begin = in_data.host_pointer(sample, 0);
+           const float_t *weight_begin  = weights.host_pointer(0, 0);
```
beru: `weight_begin` local variable can be assigned outside of sample loop.

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -28,42 +38,40 @@ inline void conv2d_op_internal(const tensor_t &in_data,
...
-                       sum += pw_element[wx] * pin_element[wx];
+                       sum += *(pw_element++) * *(pin_element++);
                      }
-                     pw_element += kw;
-                     pin_element += iw;
                    }
-                   pout[x] += sum;
-                   pin_line += elem_stride;
+                   *(pout++) += sum;
+                   pin_line = std::next(pin_line, elem_stride);
```
beru: If `pin_line` is simple pointer, you don't need to use `std::next` function.
The old way `pin_line += elem_stride;` is preferrable.

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -28,42 +38,40 @@ inline void conv2d_op_internal(const tensor_t &in_data,
...
                      for (serial_size_t wx = 0; wx < kw; wx++) {  // NOLINT
-                       sum += pw_element[wx] * pin_element[wx];
+                       sum += *(pw_element++) * *(pin_element++);
                      }
-                     pw_element += kw;
-                     pin_element += iw;
                    }
-                   pout[x] += sum;
-                   pin_line += elem_stride;
+                   *(pout++) += sum;
```
beru: This is different from original code.

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -28,42 +38,40 @@ inline void conv2d_op_internal(const tensor_t &in_data,
...
+                   pin_line = std::next(pin_line, elem_stride);
                  }
-                 pout += ow;
-                 pin += line_stride;
+                 pin = std::next(pin, line_stride);
```
beru: In original code,
```cpp
                 pout += ow;
                 pin += line_stride;
``
`pout` is advanced with `ow` variable which is `params.out.width_` but newer code increments the `pout` pointer variable inside loop.

I think the old way is better since x86/x64 processor instructions can have address displacements.

File: tiny_dnn/core/kernels/conv2d_op_internal.h [hunk 18]
```diff
@@ -28,42 +38,40 @@ inline void conv2d_op_internal(const tensor_t &in_data,
...
                }
              }
              if (params.has_bias) {
-               vectorize::add(bias[o], out_area, pa);
+               vectorize::add(bias.host_at(0, o), out_area, pa);
```
beru: Using `host_at` method inside loops can be avoided in this case.

tiny_dnn/core/kernels/conv2d_op_internal.h
```diff
@@ -72,41 +80,51 @@ inline void conv2d_op_internal(const tensor_t &in_data,
 }
 
 /******************************************************************/
+template <typename S1,
+          typename S2,
+          typename S3,
+          typename S4,
+          typename S5,
+          typename S6>
```
beru: Is it realistic that all of the tensor arguments have different storage types?

tiny_dnn/core/kernels/conv2d_op_libdnn.h
```diff
@@ -34,42 +34,47 @@ class Conv2dLibDNNForwardOp : public core::OpKernel {
...
 
+    // TODO(Randl): Remove once layers forward and backward by themself.
+    Tensor<float_t> in_data_t(context.input(0));
+    const Tensor<float_t> weights_t(context.input(1)),
+      bias_t = Tensor<float_t>(context.input(2));  // TODO has_bias
+    Tensor<float_t> out_data_t(context.output(0));
+
     // initialize outputs
-    fill_tensor(out_data, float_t{0});
+    out_data_t.fill(0.0f);
```
beru: Maybe `float_t{0}` is somewhat suitable.... `0.0f` reminds me of `float` type.

tiny_dnn/core/kernels/conv2d_op_opencl.h
```diff
@@ -20,14 +20,14 @@ class Conv2dOpenCLForwardOp : public core::OpKernel {
...
 
-    // incoming/outcoming data
-    const tensor_t &in_data = context.input(0);
-    const tensor_t &W       = context.input(1);
-    const tensor_t &bias    = context.input(2);
-    tensor_t &out_data      = context.output(1);
+    // TODO(Randl): Remove once layers forward and backward by themself.
+    Tensor<float_t> in_data_t(context.input(0));
+    const Tensor<float_t> weights_t(context.input(1)),
+      bias_t = Tensor<float_t>(context.input(2));  // TODO has_bias
```
beru: It's just my personal preferrence but declaring multiple variables in a single statement might spoil the readability in some cases.

tiny_dnn/core/kernels/conv2d_op_avx.h
beru: Is it so in `Conv2dOp::compute` method that calls `kernels::conv2d_op_avx` function?
Will `core::OpKernelContext` class have different types (`view` and `tensor`) of in and out data and grad?

tiny_dnn/core/kernels/conv2d_op_avx.h
edgarriba: /cc @SylvainCorlay

tiny_dnn/core/kernels/conv2d_op_avx.h
beru: If the last dimesion of Tensor is contiguous and a position for parent dimension is fixed, we can get a pointer from `bias` tensor outside of the loop.

It looks like `bias` variable changed from 1D to 2D....but only last dimension is specified.

tiny_dnn/core/kernels/conv2d_op_libdnn.h
Randl: @edgarriba

tiny_dnn/core/kernels/conv2d_op_internal.h
Randl: It's realistic that any two will.

beru: I wonder it is needed now...

beru: And can we solve this storage matter dynamically? I don't see a point in increasing static complexity when doing so doesn't contribute to performance/size improvement.

tiny_dnn/core/kernels/conv2d_op_avx.h
SylvainCorlay: In xsimd we have an aligned STL allocator.

We will soon work on plugging xsimd in xtensor.

tiny_dnn/core/kernels/conv2d_op_internal.h
Randl: @edgarriba @beru ?

tiny_dnn/core/kernels/conv2d_op_internal.h
Randl: Yes I've changed logic a bit. Is it worse?

beru: Please check this comment. https://github.com/tiny-dnn/tiny-dnn/pull/759#discussion_r122203461
Well but it is kinda too trivial.

tiny_dnn/core/kernels/conv2d_op_internal.h
beru: @Randl It was just a vague idea. After the integration will be finished, let's consider how we can improve the code.

tiny_dnn/core/kernels/conv2d_op_internal.h
Randl: As for now, Tensor's 2D: first dimension is samples and second is all the data. We can't obtain depth from it, we need 4D Tensor for that

tiny_dnn/core/kernels/conv2d_op_avx.h
Randl: well there is no other way to take address of first element

beru: Can't you simply do this outside of the loop?

edgarriba:
@beru approve and merge

tiny_dnn/core/kernels/conv2d_op_internal.h
beru: It's OK as long as we retrieve pointers before loops and use them inside loops.

tiny_dnn/core/kernels/conv2d_op_libdnn.h
beru: @edgarriba @edgarriba @edgarriba

tiny_dnn/core/kernels/conv2d_op_opencl.h
beru: Why `weights_t` variable is initialized with constructor, while `bias_t` is initialized with `operator =`? Can't both of them be initialized with constructors?

Issue #764, opened 2017-06-16
Implementation of custom loss layer


kvmanohar22:
As part of GSoC'17, I'll need to implement loss function as outlined in [this paper](https://arxiv.org/abs/1612.01051). This would require that I come up with a custom message in `caffe.proto` something quite similar to : 

```proto
message SqueezeDetLoss {
    // All the hyperparameters such as
    // number of classes, anchors, achor-shapes, other biassing parameters
}
```

PR #765, opened 2017-06-16
Adding message field for SqueezeDet loss

kvmanohar22:
In Ref: #764 
I've added the hyper-parameters which should be mentioned while using this loss function.


bhack:
@edgarriba @karandesai-96 Need we also to sync generally the caffe proto again from upstream?

edgarriba:
@kvmanohar22 have you tried this? is there any way that you can provide a test for this?

kvmanohar22:
@edgarriba I haven't tested this. But after reading the paper mentioned in #764, I felt these fields would be required in order to compute the loss.

edgarriba:
@kvmanohar22 we cannot merge without tests

edgarriba:
@kvmanohar22 any progress on that? close this and reopen once you have something more complet

kvmanohar22:
@edgarriba I'm infact working on the loss function for caffe, so I guess we can close this.

PR #768, opened 2017-06-17
Single step recurrent cell layer functionality

prlz77:
This PR extends tiny-dnn to allow recurrent models. List of extensions:
   
**Recurrent Cell**: most basic rnn module implementing equations for a simple rnn cell.
- [x] Inference.
- [x] Training.
- [x] Unit tests.
- [x] Serialization.

Next PRs will integrate Xtensor and add a wrapper on cells that will manage state transitions.

### DIFF
--- a/test/test_serialization.h
+++ b/test/test_serialization.h
@@ -818,19 +818,10 @@ TEST(serialization, sequential_to_json) {
   EXPECT_EQ(net1.in_data_size(), net2.in_data_size());
   EXPECT_EQ(net1.layer_size(), net2.layer_size());
 
-  EXPECT_EQ(net1[0]->in_shape(), net2[0]->in_shape());
-  EXPECT_EQ(net1[1]->in_shape(), net2[1]->in_shape());
-  EXPECT_EQ(net1[2]->in_shape(), net2[2]->in_shape());
-  EXPECT_EQ(net1[3]->in_shape(), net2[3]->in_shape());
-  EXPECT_EQ(net1[4]->in_shape(), net2[4]->in_shape());
-  EXPECT_EQ(net1[5]->in_shape(), net2[5]->in_shape());
-
-  EXPECT_EQ(net1[0]->layer_type(), net2[0]->layer_type());
-  EXPECT_EQ(net1[1]->layer_type(), net2[1]->layer_type());
-  EXPECT_EQ(net1[2]->layer_type(), net2[2]->layer_type());
-  EXPECT_EQ(net1[3]->layer_type(), net2[3]->layer_type());
-  EXPECT_EQ(net1[4]->layer_type(), net2[4]->layer_type());
-  EXPECT_EQ(net1[5]->layer_type(), net2[5]->layer_type());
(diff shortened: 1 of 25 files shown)


test/test_serialization.h
```diff
@@ -769,6 +808,7 @@ TEST(serialization, sequential_to_json) {
   net1 << fully_connected_layer(10, 100) << tanh_layer()
        << dropout_layer(100, 0.3f, net_phase::test)
        << fully_connected_layer(100, 9) << softmax()
+       << recurrent_cell_layer(9, 9, false, new elu_layer)
```
beru: Since you have increased a layer in here, you should probablly update the teste code below.

I'd use a loop like this.
```cpp
for (size_t i=0; i<net1.layer_size(); ++i) {
  EXPECT_EQ(net1[i]->in_shape(), net2[i]->in_shape());
  EXPECT_EQ(net1[i]->layer_type(), net2[i]->layer_type());
}
```

tiny_dnn/activations/elu_layer.h
```diff
@@ -41,3 +41,8 @@ class elu_layer : public activation_layer {
 };
 
 }  // namespace tiny_dnn
+
+// allow upcasting with cereal
+#ifndef CNN_NO_SERIALIZATION
+CEREAL_REGISTER_TYPE(tiny_dnn::elu_layer);
```
beru: I think It would be better if you can move `CEREAL_REGISTER_TYPE(...)` macros into a single header file.
They are scattered now.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -0,0 +1,154 @@
...
+                                       tensor_t &out_h,
+                                       const recurrent_cell_params &params,
+                                       const bool layer_parallelize) {
+  for_i(layer_parallelize, in_data.size(), [&](size_t sample) {
+    const vec_t &in         = in_data[sample];
+    const vec_t &prev_state = prev_h[sample];
+    vec_t &out              = out_data[sample];
+    vec_t &next_state       = out_h[sample];
+
+    for (size_t o = 0; o < params.out_size_; o++) {
```
beru: Inside this loop, you can add a local variable. Accessing to `next_state[o]` should be minimized.

tiny_dnn/core/kernels/recurrent_cell_op_internal.h
```diff
@@ -0,0 +1,154 @@
...
+
+      if (params.has_bias_) {
+        next_state[o] += bias[o];
+      }
+    }
+
+    params.activation_->forward_activation(next_state, next_state);
+
+    // V matrix is out_size_ x out_size_
+    for (size_t o = 0; o < params.out_size_; o++) {
```
beru: In this loop, you can add a local variable. Accessing to `out[o]` should be minimized.

edgarriba:
@beru @nyanp let's merge this

Issue #560, opened 2017-02-10
Rewrite deconvolution (transposed convolution) as inherited from convolution


Randl:
As for now, deconvolution is broken:
https://github.com/tiny-dnn/tiny-dnn/issues/278
https://github.com/tiny-dnn/tiny-dnn/issues/199

Also, as you probably know, classical way is to use transposed convolution instead of real deconvolution. It can be implemented as convolution with minor changes:
https://github.com/soumith/cudnn.torch/blob/master/SpatialFullConvolution.lua
https://groups.google.com/forum/#!topic/lasagne-users/Vc5Jg0qxflE

So I suppose to derive `transposed_convolutional_layer` from `convolutional_layer` and make needed changes there. 

Connected issue with useful links:
https://github.com/tiny-dnn/tiny-dnn/issues/303

PR #770, opened 2017-06-18
Deconvolution Tensor integration

Randl:


beru:
If I remember correctly, having backend methods is kinda outdated.
Instead, we have decided to user KernelOp pattern.

Randl:
@beru That's nice that we gotta review all the code while I'm porting it, but it's definitely out of scope.

bhack:
I agree cause the real hole is in https://github.com/tiny-dnn/tiny-dnn/issues/560

tiny_dnn/core/kernels/deconv2d_grad_kernel_internal.h
```diff
@@ -12,16 +12,27 @@
...
+inline void tiny_deconv2d_back_kernel(const Tensor<float_t, S1> &prev_out,
+                                      const Tensor<float_t, S2> &weights,
+                                      Tensor<float_t, S3> &weights_grads,
+                                      Tensor<float_t, S4> &bias_grads,
+                                      Tensor<float_t, S5> &curr_delta,
+                                      Tensor<float_t, S6> &prev_delta,
+                                      const deconv_params &params) {
+  const float_t *W = weights.host_pointer(0, 0);  // TODO(Randl)
   // propagate delta to previous layer
   for_i(prev_out.size(), [&](int sample) {
```
edgarriba: `size_t`

beru:
@Randl 
> That's nice that we gotta review all the code while I'm porting it, but it's definitely out of scope.

I see, then migrating to KernelOp pattern can be done in a different PR.

What about removing (supposedly) unnecessary avx kernels?
Maybe you don't want to perform that clean-up in this PR.

It looks like you renamed some files, but should they be done in this PR?

Randl:
> Maybe you don't want to perform that clean-up in this PR.

Clean up is ok, but I'd like to avoid rewriting things that aren't connected to `Tensor`.

> What about removing (supposedly) unnecessary avx kernels?

Not sure. Would we need to return them when (if) we want to implement AVX kernel. Does having this basement for infrastructure hurting us? (it might be misleading though). @edgarriba  @bhack ?

edgarriba:
@beru approve and merge

Issue #726, opened 2017-05-24
Add tests for conv with no bias


edgarriba:
A user reported some issues with convolution without bias. We should add tests for this so that we can fix it.

Randl:
https://github.com/tiny-dnn/tiny-dnn/pull/730

bhack:
Also @covteam incurred in this

Issue #665, opened 2017-04-21
Turning off bias on conv layer leads to access violation


kokostek:
Actually, I get this error message while debugging with Visual Studio:
```
Debug Assertion Failed!

Program: C:\Windows\system32\MSVCP140D.dll
File: c:\program files (x86)\microsoft visual studio 14.0\vc\include\vector
Line: 1218

Expression: vector subscript out of range

For information on how your program can cause an assertion
failure, see the Visual C++ documentation on asserts.

(Press Retry to debug the application)
```

It happens in `conv2d_op.h` at this line:
```
    const tensor_t &bias    = context.input(2);
```
Seems like it is still trying to access bias vector even if its off.

Here is my code with nn initalization, feature loading and prediction:
```
    network<sequential> net;

    net 
        << conv(64, 64, 9, 1, 3, padding::valid, false) << relu()
        << max_pool(56, 56, 3, 2)

        << conv(28, 28, 5, 3, 9, padding::same) << relu()
        << max_pool(28, 28, 9, 2)

        << conv(14, 14, 3, 9, 27, padding::same) << relu()
        << max_pool(14, 14, 27, 2)

        << fc(7 * 7 * 27, 300) << relu()
        << fc(300, 30) << relu()
        << fc(30, 3) << softmax();

    vec_t x = convert_image("d:\\work\\video\\dataset\\1\\0000.png", 1.0, 64, 64);  // load features
    vec_t y = net.predict(x);  // predict with with random weights (error happens here)
```

Randl:
Same code with biases works, right?

kokostek:
Yes, it does.

Randl:
Can you provide sample image and also full compiling example?

kokostek:
There is really no need for a sample image. It can just be a random `vec_t` object.

This code runs fine:
``` c++
#include <tiny_dnn/tiny_dnn.h>

using namespace tiny_dnn;

int main()
{
    network<sequential> net;
    
    net << conv(32, 32, 3, 1, 1, padding::same, true);
    vec_t input(32 * 32);
    net.predict(input);

    return 0;
}
```

and this asserts:
``` c++
#include <tiny_dnn/tiny_dnn.h>

using namespace tiny_dnn;

int main()
{
    network<sequential> net;
    
    net << conv(32, 32, 3, 1, 1, padding::same, false);
    vec_t input(32 * 32);
    net.predict(input);

    return 0;
}
```

[...]

PR #772, opened 2017-06-20
Tests without bias

Randl:
https://github.com/tiny-dnn/tiny-dnn/issues/726
https://github.com/tiny-dnn/tiny-dnn/issues/665


edgarriba:
@Randl could you make a quick fix to https://github.com/tiny-dnn/tiny-dnn/pull/730 until we don't merge this to master?

edgarriba:
we could start thinking to have proper classes to generate testing data. Maybe per layer?

PR #774, opened 2017-06-21
Introduce parameter class to represent trainable / constant parameters of layer.

karandesai-96:
This PR is the first one to refactor underlying mechanism of representation of trainable weights of a layer. We have a class named `edge` which holds `tensor_t` of data and grad. `edge` represents the incoming and outgoing data to / from a layer. The end game is to make this `edge` hold only references to nodes instead of actual tensors with numbers.

* `edge` was a generic class to represent data, labels, weights, biases and auxiliary tensors, but now a separate class named `parameter` has been made to represent weights and biases.

* `parameter` class contains members to hold dimensions and number of filters of weights and biases, a boolean to check whether trainable or not and tensors of data and gradients, complete with their getters and setters.

### Roadmap Further:

* Layer constructors shall not accept weights and biases in their input type constructors, rather they shall call `layer::add_parameter` method in the constructor to add weights and biases. Same for forward and backward propagation methods. 

* Layers will have suitable getter - setters to load pre trained weights and save them effectively. 

### Requirement of parameter class:

Please refer the discussion starting from https://github.com/tiny-dnn/tiny-dnn/pull/754#issuecomment-309910155

### DIFF
--- a/test/test_parameter.h
+++ b/test/test_parameter.h
@@ -1,0 +1,59 @@
+/*
+    Copyright (c) 2013, Taiga Nomi and the respective contributors
+    All rights reserved.
+
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
+#pragma once
+#include "gtest/gtest.h"
+#include "testhelper.h"
+#include "tiny_dnn/tiny_dnn.h"
+
+namespace tiny_dnn {
+
+TEST(parameter, init) {
+  parameter p(3, 3, 1, 1, parameter_type::weight, true);
+
+  ASSERT_EQ(p.shape().width_, 3u);
+  ASSERT_EQ(p.shape().height_, 3u);
+  ASSERT_EQ(p.shape().depth_, 1u);
+  ASSERT_EQ(p.size(), 9u);
+  ASSERT_EQ(p.type(), parameter_type::weight);
+
+  ASSERT_TRUE(p.is_trainable());
+}
+
+TEST(parameter, getter_setter) {
+  parameter p(4, 1, 1, 1, parameter_type::bias, false);
+
+  xt::xarray<float_t> xr = {1.0, 2.0, 3.0, 4.0};
+  Tensor<float_t> t{xr};
+
+  p.set_data(t);
+  Tensor<float_t> *pt = p.data();
+
+  for (unsigned int i = 0; i < t.size(); i++) {
+    ASSERT_EQ(pt->host_at(i), t.host_at(i));
+  }
+}
+
+TEST(parameter, merge_grads) {
+  xt::xarray<float_t> xgrad = {{1.0, 2.0, 3.0, 4.0}, {4.0, 3.0, 2.0, 1.0}};
+  Tensor<float_t> grad0{xgrad};
+  Tensor<float_t> gradp{xgrad * 2};
+
+  parameter p(4, 1, 1, 1, parameter_type::bias, false);
+  p.set_grad(gradp);
+  p.merge_grads(&grad0);
+
+  Tensor<float_t> expected{
+    {{10.0, 10.0, 10.0, 10.0}, {10.0, 10.0, 10.0, 10.0}}};
+
+  for (unsigned int i = 0; i < p.size(); i++) {
+    ASSERT_EQ(grad0.host_at(0, i), expected.host_at(0, i));
+    ASSERT_EQ(grad0.host_at(1, i), expected.host_at(1, i));
+  }
+}
+
+}  // namespace tiny_dnn
--- a/test/test_parameter.h
+++ b/test/test_parameter.h
@@ -31,7 +31,7 @@ TEST(parameter, getter_setter) {
   p.set_data(t);
   Tensor<float_t> *pt = p.data();
 
-  for (unsigned int i = 0; i < t.size(); i++) {
+  for (size_t i = 0; i < t.size(); i++) {
     ASSERT_EQ(pt->host_at(i), t.host_at(i));
   }
 }
@@ -46,7 +46,7 @@ TEST(parameter, merge_grads) {
 
   Tensor<float_t> expected{{{6.0, 6.0}, {6.0, 6.0}}};
 
-  for (unsigned int i = 0; i < p.size(); i++) {
+  for (size_t i = 0; i < p.size(); i++) {
     ASSERT_EQ(grad0.host_at(0, i), expected.host_at(0, i));
     ASSERT_EQ(grad0.host_at(1, i), expected.host_at(1, i));
   }
--- a/test/test_parameter.h
+++ b/test/test_parameter.h
@@ -26,9 +26,7 @@ TEST(parameter, init) {
 
 TEST(parameter, getter_setter) {
   parameter p(4, 1, 1, 1, parameter_type::bias, false);
-
-  xt::xarray<float_t> xr = {1.0, 2.0, 3.0, 4.0};
-  Tensor<float_t> t{xr};
+  Tensor<float_t> t{{1.0, 2.0, 3.0, 4.0}};
 
   p.set_data(t);
   Tensor<float_t> *pt = p.data();
@@ -39,16 +37,14 @@ TEST(parameter, getter_setter) {
 }
 
 TEST(parameter, merge_grads) {
-  xt::xarray<float_t> xgrad = {{1.0, 2.0, 3.0, 4.0}, {4.0, 3.0, 2.0, 1.0}};
-  Tensor<float_t> grad0{xgrad};
-  Tensor<float_t> gradp{xgrad * 2};
+  Tensor<float_t> grad0{{{1.0, 2.0}, {2.0, 1.0}}};
+  Tensor<float_t> gradp{{{2.0, 4.0}, {4.0, 2.0}}};
 
-  parameter p(4, 1, 1, 1, parameter_type::bias, false);
+  parameter p(2, 1, 1, 1, parameter_type::bias, false);
   p.set_grad(gradp);
   p.merge_grads(&grad0);
 
-  Tensor<float_t> expected{
-    {{10.0, 10.0, 10.0, 10.0}, {10.0, 10.0, 10.0, 10.0}}};
+  Tensor<float_t> expected{{{6.0, 6.0}, {6.0, 6.0}}};
 
   for (unsigned int i = 0; i < p.size(); i++) {
     ASSERT_EQ(grad0.host_at(0, i), expected.host_at(0, i));
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -46,7 +46,7 @@ class Tensor {
    * @param storage object, can be xtarray.
    * @return
    */
-  explicit Tensor(Storage s) : storage_(s) {}
+  explicit Tensor(const Storage &storage) : storage_(storage) {}
 
   /**
    * Constructor that accepts an initializer list of shape and create a
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -41,6 +41,14 @@ class Tensor {
   Tensor() {}
 
   /**
+   * Constructor that accepts a storage object and create a Tensor housing
+   * that storage. Commonly used by [] operator.
+   * @param storage object, can be xtarray.
+   * @return
+   */
+  explicit Tensor(Storage s) : storage_(s) {}
+
+  /**
    * Constructor that accepts an initializer list of shape and create a
    * Tensor with that shape. For example, given shape = {2,3,4,5,6}, tensor
    * will be of size 2x3x4x5x6. Note: tensor isn't initialized by default
--- a/tiny_dnn/parameter.h
+++ b/tiny_dnn/parameter.h
@@ -7,11 +7,12 @@
 */
 #pragma once
 
+#include "tiny_dnn/core/framework/tensor.h"
 #include "tiny_dnn/util/util.h"
 
 namespace tiny_dnn {
 
-enum parameter_type : int8_t { weight = 0x0001, bias = 0x0002 };
+enum class parameter_type : int8_t { weight = 0x0001, bias = 0x0002 };
 
 class parameter {
  public:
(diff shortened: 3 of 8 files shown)


tiny_dnn/core/framework/tensor.h
```diff
@@ -41,6 +41,14 @@ class Tensor {
   Tensor() {}
 
   /**
+   * Constructor that accepts a storage object and create a Tensor housing
+   * that storage. Commonly used by [] operator.
+   * @param storage object, can be xtarray.
+   * @return
+   */
+  explicit Tensor(Storage s) : storage_(s) {}
```
edgarriba: meaningful names `s -> storage`

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+   *
+   * @param width      [in] filter width
+   * @param height     [in] filter height
+   * @param depth      [in] filter depth / input channels
+   * @param n_fmaps    [in] number of output feature maps
+   * @param type       [in] whether parameter is a weight or a bias
+   * @param trainable  [in] whether parameter will be updated while training or
+   * not
+   */
+  parameter(serial_size_t width,
```
edgarriba: `serial_size_t` doesn't exist anymore. I'll update the branch

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+    : type_(type),
+      shape_(width, height, depth),
+      n_fmaps_(n_fmaps),
+      trainable_(trainable),
+      data_({size()}),
+      grad_({1, size()}) {}
+
+  shape3d shape() { return shape_; }
+
+  size_t size() { return shape_.size() * n_fmaps_; }
```
edgarriba: I guess that `Tensor` has a routine for that, if that's not the case implement it

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+  void set_dims(serial_size_t width,
+                serial_size_t height,
+                serial_size_t depth,
+                serial_size_t n_fmaps) {
+    shape_.width_  = width;
+    shape_.height_ = height;
+    shape_.depth_  = depth;
+    n_fmaps_       = n_fmaps;
+    data_.reshape({size()});
+    grad_.reshape({grad_.shape()[0], size()});
```
edgarriba: what's the motivation to have this routine? are we reshaping tensors on demand?

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+  float_t *data_at(size_t i) { return &data_.host_at(i); }
+
+  float_t *grad_at(size_t sample, size_t i) {
+    return &grad_.host_at(sample, i);
+  }
+
+ private:
+  parameter_type type_;
+  shape3d shape_;
+  size_t n_fmaps_;
```
edgarriba: not sure that `shape_`  and `n_fmaps_` should exist

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+
+enum class parameter_type : int8_t { weight = 0x0001, bias = 0x0002 };
+
+class parameter {
+ public:
+  /**
+   * Initializes an empty parameter taking in the dimensions of weights and
+   * biases of a layer. Currently supported for maximum 4-dimensions, and
+   * stored as a flat ``Tensor``.
+   *
```
edgarriba: you should document what is out memory model: NCHW, NHWC and reorder the constructor arguments according to that. However, I think that as initial design is a mistake to model data as a flat vector. I would do it using the memory model and if somewhere you need a flat vector you can create a flatten view of the data.

test/test_parameter.h
```diff
@@ -0,0 +1,59 @@
...
+  ASSERT_EQ(p.size(), 9u);
+  ASSERT_EQ(p.type(), parameter_type::weight);
+
+  ASSERT_TRUE(p.is_trainable());
+}
+
+TEST(parameter, getter_setter) {
+  parameter p(4, 1, 1, 1, parameter_type::bias, false);
+
+  xt::xarray<float_t> xr = {1.0, 2.0, 3.0, 4.0};
```
edgarriba: I'd would go for a simple vector. We should avoid to use `xt::` outside of `Tensor`

test/test_parameter.h
```diff
@@ -0,0 +1,59 @@
...
+  p.set_data(t);
+  Tensor<float_t> *pt = p.data();
+
+  for (unsigned int i = 0; i < t.size(); i++) {
+    ASSERT_EQ(pt->host_at(i), t.host_at(i));
+  }
+}
+
+TEST(parameter, merge_grads) {
+  xt::xarray<float_t> xgrad = {{1.0, 2.0, 3.0, 4.0}, {4.0, 3.0, 2.0, 1.0}};
```
edgarriba: same as before, use vector

tiny_dnn/parameter.h
edgarriba: updated! please rebase

tiny_dnn/core/framework/tensor.h
beru: The scope of the argument variable is pretty limited. So I don't think it is a must to use full-length word here.

A lot of people have written books about programming styles, argued on internet forums. Maybe it's like pigeons talking about proper manners for their droppings.

But, hey, why the argument isn't a referenced type? We are scheming how to produce heat with computer processors and copying tensor is a hefty operation.

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+ private:
+  parameter_type type_;
+  shape3d shape_;
+  size_t n_fmaps_;
+  bool trainable_;
+
+  Tensor<float_t> data_;
+  Tensor<float_t> grad_;
+};
+}  // namespace tiny_dnn
```
beru: Please add a new line between here like this
```cpp
};  // class parameter

}  // namespace tiny_dnn
```
if doing so dosn't significantly reduce your QOL.

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,125 @@
...
+#pragma once
+
+#include "tiny_dnn/core/framework/tensor.h"
+#include "tiny_dnn/util/util.h"
+
+namespace tiny_dnn {
+
+enum class parameter_type : int8_t { weight = 0x0001, bias = 0x0002 };
+
+class parameter {
```
beru: Should this class carry template parameters for Tensor class instances?
Here, parameter class isn't a template class so its member Tensor variables are like this
```cpp
  Tensor<float_t> data_;
  Tensor<float_t> grad_;
```

If we look at the definition of Tensor class, its first templat parameter is `typename U = float_t` so I wonder why people explitly write `<float_t>` all over the library.

Personally, I don't want to carry template parameters all the time when I write code, so I really wish Tensor class's data type would be dynamic.

tiny_dnn/core/framework/tensor.h
edgarriba: @beru yeah, lot of literature about this but the more context the better. At least is what we teach to the guys at university. Agree, @karandesai-96 pass storage by reference.

tiny_dnn/parameter.h
prlz77: For moving from tiny-cnn to tiny-dnn, I think the dimensionality of the parameters should be unbounded and layers should complain later if they did not receive what they expected. So I would get rid of hardcoded width, depth... and just use an array to represent the dimensionality.
Also using shape3d seems limited to me.

tiny_dnn/parameter.h
edgarriba: @beru agree with you, however not sure how much work implies this feature since xtensor is templated. But I like the idea of dynamic type. /cc @Randl

tiny_dnn/parameter.h
karandesai-96: Dealing with views is not as smooth as it should be, but it should be functional within some time. Besides we are accessing members of data and grad from a flat tensor so it will help me integrate working model of parameter without making radical changes. But yes, I will add a documentation of NCHW.

karandesai-96: Currently **all** the kernels assume parameters to be flat.

tiny_dnn/parameter.h
karandesai-96: @prlz77 @edgarriba I am targeting to do minimal changes to the existing codebase and get parameter class working so I am tactically adopting the older design as much as possible. I agree that `shape3d` limits generalization but I feel it would be easier to generalize once it is integrated everywhere (you know, tracing the failing tests and making them pass again).

I have added two todos for the purpose:
```c++
+   * todo (karandesai) : generalize to n-dimensions
+   * todo (karandesai) : add an n-dimensional view for easy indexing
```

I am adding a lot of todos lately but `feat/parameters_refactor` shall not take any extra todos when the time comes to merge it into master.

tiny_dnn/parameter.h
karandesai-96: I think I should remove this and only add it later on a need-to-have basis.

tiny_dnn/parameter.h
prlz77: Nice!

tiny_dnn/core/framework/tensor.h
beru: @edgarriba C++ is a strongly typed language. In this case, the argument type name `Storage` represents its characteristics so renaming the argument to `storage` does not give much benefit. Moreover, its variable scope is quite short.

Having more context isn't always good. Less is More. Controlling pace and accent is important.

karandesai-96:
/ping: @edgarriba @bhack @Randl There are improvements but they will come up soon. I am ready with next PRs.

test/test_parameter.h
```diff
@@ -0,0 +1,55 @@
...
+  Tensor<float_t> grad0{{{1.0, 2.0}, {2.0, 1.0}}};
+  Tensor<float_t> gradp{{{2.0, 4.0}, {4.0, 2.0}}};
+
+  parameter p(2, 1, 1, 1, parameter_type::bias, false);
+  p.set_grad(gradp);
+  p.merge_grads(&grad0);
+
+  Tensor<float_t> expected{{{6.0, 6.0}, {6.0, 6.0}}};
+
+  for (unsigned int i = 0; i < p.size(); i++) {
```
beru: The type of the loop counter variable `i` should be `size_t`.

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,120 @@
...
+    in the LICENSE file.
+*/
+#pragma once
+
+#include "tiny_dnn/core/framework/tensor.h"
+#include "tiny_dnn/util/util.h"
+
+namespace tiny_dnn {
+
+enum class parameter_type : int8_t { weight = 0x0001, bias = 0x0002 };
```
beru: Do you need 4 hex letters for a byte storage?

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,120 @@
...
+
+  size_t size() { return data_.size(); }
+
+  parameter_type type() { return type_; }
+
+  bool is_trainable() { return trainable_; }
+
+  void set_trainable() { trainable_ = true; }
+
+  void freeze_trainable() { trainable_ = false; }
```
beru: If `is_trainable`, `set_trainable`, `freeze_trainable` methods are this simple, can't we just make `trainable_` member to public and remove these?

edgarriba: I would go just for `trainable()` and `set_trainable(bool trainable)`

nyanp: vote for @edgarriba 's idea.

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,120 @@
...
+
+  Tensor<float_t> *data() { return &data_; }
+
+  const Tensor<float_t> *data() const { return &data_; }
+
+  void set_data(const Tensor<float_t> &data) { data_ = data; }
+
+  void set_data(const Tensor<float_t> *data) { data_ = *data; }
```
nyanp: This overload seems redundant. I would just use ```void set_data(const Tensor<float_t> &data) { data_ = data; }```

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,120 @@
...
+
+  Tensor<float_t> *grad() { return &grad_; }
+
+  const Tensor<float_t> *grad() const { return &grad_; }
+
+  void set_grad(const Tensor<float_t> &grad) { grad_ = grad; }
+
+  void set_grad(const Tensor<float_t> *grad) { grad_ = *grad; }
```
nyanp: Same as ```set_data```.

edgarriba:
@beru @Randl review and merge in favour of #780

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,114 @@
...
+    in the LICENSE file.
+*/
+#pragma once
+
+#include "tiny_dnn/core/framework/tensor.h"
+#include "tiny_dnn/util/util.h"
+
+namespace tiny_dnn {
+
+enum class parameter_type : int8_t { weight = 0x1, bias = 0x2 };
```
beru: I'm not really sure why this enum has to be 1 byte object...
Is it really meaningful to save storage size of `parameter::type_`?

karandesai-96: @beru I kept it to maintain consistency with old `vector_type` ( enum of data, weight, bias, label)

karandesai-96:
@beru Sure, a lot of refactoring coming over soon. It saves a good amount of back and forth of tensors in computational graph.

tiny_dnn/parameter.h
beru: @karandesai-96 This one? What I see is 4 bytes type though.
```cpp
enum class vector_type : int32_t {
  // 0x0001XXX : in/out data
  data = 0x0001000,  // input/output data, fed by other layer or input channel

  // 0x0002XXX : trainable parameters, updated for each back propagation
  weight = 0x0002000,
  bias   = 0x0002001,

  label = 0x0004000,
  aux   = 0x0010000  // layer-specific storage
};
```

edgarriba: parameters should be only weight and bias type

PR #776, opened 2017-06-21
size_t in rnn tests

prlz77:
Substitutes `serial_size_t` and `int` for `size_t` in test recurrent.
Adds it again in `test.cpp'


edgarriba:
@beru @nyanp @Randl approve this

PR #778, opened 2017-06-25
An attempt on BaseTensor

Randl:


tiny_dnn/core/framework/tensor.h
```diff
@@ -138,19 +161,19 @@ class Tensor {
...
    */
-  const auto &shape() const { return storage_.shape(); }
+  const std::vector<size_t> &shape() const override { return storage_.shape(); }
 
   /**
    *
    * @return Tensor's number of dimensions
    */
-  const auto dim() const { return storage_.dimension(); }
+  const size_t dim() const override { return storage_.dimension(); }
```
beru: Should the return type be `const size_t`? I though `size_t` is enough....

Randl:
@SylvainCorlay can you please give your opinion/advice?

bhack:
At which level we will integrate with  https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h? In xtensor?

Randl:
@bhack This is basically different tensor implementation, so either `Tensor` or `xtensor` level

bhack:
@Randl Yes but just cause we invited @SylvainCorlay probably we could also clarify this. There was a thread on slack on this topic

edgarriba:
@Randl close and open the PR, travisCI crashed.

In addition, after debugging Tensor I think that will be very productive to have a `TensorShape` class with and overloaded method to display just the shape. I'd expect something like:

```cpp
Tensor my_tensor({2, 2, 2});
std::cout << my_tensor.size() << std::endl;
```
Expected output:
```bash
[2, 2, 2]
```

Randl:
@bhack FYI it's possible to restart jobs in Travis

edgarriba:
@Randl should we merge or are committing more changes here?

Randl:
I'd like to get review from @SylvainCorlay but no more commits are planned

SylvainCorlay:
I have read the code and have comments but had to leave early. I will post the discussion items on the gitter room.

edgarriba:
@Randl any progress on that?

PR #780, opened 2017-06-27
Replicate methods in layer.h related to old weights for parameters.

karandesai-96:
#### Continuation of parameter integration succeeding #774
This PR adds certain methods in `layer.h` which replicate the functioning of similar old methods used for manipulating weights of the layer. Several todos are inserted at various places indicating future removal of those methods. Duplication is carried out to save Travis from failing.

### Important Additions:

* Getters and Setters for parameter
* init_parameters
* update_parameters
* has_same_parameters


karandesai-96:
@edgarriba Please have a review after #774 is merged, it will reduce the diff to review here.

tiny_dnn/layers/layer.h
```diff
@@ -66,9 +68,13 @@ class layer : public node {
       out_channels_(out_type.size()),
       in_type_(in_type),
       out_type_(out_type) {
+    // todo (karandesai) : remove these after parameter integration
     weight_init_ = std::make_shared<weight_init::xavier>();
     bias_init_   = std::make_shared<weight_init::constant>();
-    trainable_   = true;
+
+    weight_init_f_ = std::make_shared<parameter_init::xavier>();
```
edgarriba: I think by default we should just initialize parameters with vanilla uniform_sampling.

Xavier initialization is just one of the many different extensions of uniform sampling that somehow works for CNN. Since we are not just focusing on CNN, this could lead to errors during training other types of networks. @prlz77 might give us a more theoretical understanding about that ?

tiny_dnn/layers/layer.h
```diff
@@ -347,6 +353,70 @@ class layer : public node {
     return *this;
   }
 
+  /**
+   * @name Parameter Getters and Setters
+   * @{
+   */
+
+  /** @brief Add new parameter to this layer, to be called in constructor. */
+  void add_parameter(size_t width,
```
edgarriba: I'd reorder parameters following the data storage model: NCHW or in this case `n_fmaps, depth, height, width`. 

Besides, I'm not really sure that this vocabulary is the best for this scope. `in_features, ?, height, width`. Take also into account that RNN might have extra dimensionality at some points, not just 4 dimensions.

@prlz77 could you confirm that?

tiny_dnn/layers/layer.h
```diff
@@ -347,6 +353,70 @@ class layer : public node {
...
+  /**
+   * @brief Get pointers to parameters of this layer.
+   *
+   * This can be used during training phase when parameter update is
+   * required.
+   *
+   * @param trainable_only flag to return only the trainable parameters.
+   * @return std::vector of pointers to parameters
+   */
+  std::vector<parameter *> parameters(bool trainable_only = false) {
```
edgarriba: why not using `shared_ptr` ? could you run a small benchmark to see if there's any argument in favour/against ?

tiny_dnn/layers/layer.h
```diff
@@ -347,6 +353,70 @@ class layer : public node {
...
+    std::vector<parameter *> vp;
+    for (size_t i = 0; i < parameters_.size(); i++) {
```
edgarriba: use for range based loop

tiny_dnn/layers/layer.h
```diff
@@ -347,6 +353,70 @@ class layer : public node {
...
+      if (!trainable_only ||
+          (parameters_[i]->is_trainable() && trainable_only)) {
+        vp.push_back(parameters_[i]);
+      }
+    }
+    return vp;
+  }
+
+  parameter *ith_parameter(size_t i) { return parameters_[i]; }
```
edgarriba: same as before, let's evaluate the use of shared_ptr

tiny_dnn/layers/layer.h
```diff
@@ -644,12 +717,57 @@ class layer : public node {
...
+      initialized_ = true;
+      return;
+    }
+
+    // Fill parameter values with data generated by the initialization
+    // function. The pointer to the data is obtained from the
+    // computational graph and the methods fan_in_size() and fan_out_size()
+    // return the number of incoming/outcoming connections for each
+    // input/output unit.
+    for (size_t i = 0; i < parameters_.size(); i++) {
```
edgarriba: you can use range base for loop. As for ` fan_in_size(i)` I guess we could get those from tensor shape, but not really sure what's returning

tiny_dnn/layers/layer.h
```diff
@@ -644,12 +717,57 @@ class layer : public node {
...
+    // layer/node as initialized.
+    initialized_ = true;
+  }
+
   void clear_grads() {
     for (size_t i = 0; i < in_type_.size(); i++) {
       ith_in_node(i)->clear_grads();
     }
+
+    for (size_t i = 0; i < parameters_.size(); i++) {
```
edgarriba: range for loop

tiny_dnn/layers/layer.h
```diff
@@ -670,6 +788,33 @@ class layer : public node {
     post_update();
   }
 
+  void update_parameters(optimizer *o, size_t batch_size) {
```
edgarriba: meaningful names `o -> optimizer_ptr` or whatever alternative

tiny_dnn/layers/layer.h
```diff
@@ -670,6 +788,33 @@ class layer : public node {
...
+    float_t rcp_batch_size  = float_t{1} / float_t{batch_size};
+    Tensor<float_t> &diff_t = parameters_diff_;
+    for (size_t i = 0; i < parameters_.size(); i++) {
```
edgarriba: for range based loop

tiny_dnn/layers/layer.h
```diff
@@ -704,6 +867,10 @@ class layer : public node {
       }
       resize(ith_out_node(i)->get_gradient());
     }
+
+    for (size_t i = 0; i < parameters_.size(); i++) {
+      parameters_[i]->resize_grad(sample_count);
+    }
```
edgarriba: now I see what you meant before. Here you might want to resize or create a new tensor with size NxD1xD2xDn, where N=sample_count. `resize_grad` calls reshape, right? Not sure just ideas :smile:

karandesai-96:
@edgarriba @beru @Randl @bhack @nyanp:

Hi, as I said earlier, I am one step closer to my initial plan of integrating parameter properly. Locally I have got it working with fully connected layer so now I will make some changes (say n dimensional generalization of parameter) and watch the tests of fully connected layer fail. I'll make the tests pass again by any means. This would be repeated over and over.

Let us discuss the possibilities of naming the parameter variables as well as ordering the argument types. One matter of decision can be uniformity with existing popular frameworks.

Here are the docs of Pytorch Conv2D kernel - it accepts a Tensor for weights in NCHW format ( [link](http://pytorch.org/docs/master/nn.html#torch.nn.functional.conv2d) ), but usually when user facing API is concerned (layer constructors in our case as most of the users won't wire up raw kernel methods), the variables are usually ordered as lower dimension first, i.e. `[kernel_shape (width, height), in_channels, out_channels]`.

But since we are thinking on generalizing to n-dimensions, we should accept a `vector<size_t>`. And when we are doing that, I think it is most suitable to accept as NCHW.

tiny_dnn/layers/layer.h
karandesai-96: Yes, I meant this.. Yes I am calling reshape internally.

karandesai-96:
@edgarriba I read this blog and decided to stick with raw pointers but I would definitely be interested in doing a quantitative comparison...

Blog: https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

karandesai-96:
I think range based for loops can be used freely as each member is a pointer after all.

tiny_dnn/layers/layer.h
prlz77: Well, gauss normal maybe is more standard and usually works well (worse than xavier in CNNs). I think normal init this is what other frameworks do.

tiny_dnn/util/parameter_init.h
```diff
@@ -53,10 +53,10 @@ class constant : public scalable {
   constant() : scalable(float_t{0}) {}
   explicit constant(float_t value) : scalable(value) {}
 
-  void fill(parameter *p, size_t fan_in, size_t fan_out) override {
+  void fill(Parameter *parameter, size_t fan_in, size_t fan_out) override {
```
edgarriba: @beru that's a clear example why I'm a fan of camelCase 😄 I can understand that snake_case is attractive at user level but I don't think it's practic. I don't want to start a forever discussion.

tiny_dnn/layers/layer.h
```diff
@@ -378,15 +378,14 @@ class layer : public node {
...
    */
-  std::vector<parameter *> parameters(bool trainable_only = false) {
-    std::vector<parameter *> vp;
-    for (size_t i = 0; i < parameters_.size(); i++) {
-      if (!trainable_only ||
-          (parameters_[i]->is_trainable() && trainable_only)) {
-        vp.push_back(parameters_[i]);
+  Parameters parameters(bool trainable_only = false) {
+    Parameters parameters;
+    for (Parameter *parameter : parameters) {
```
edgarriba: `parameters -> parameters_`

edgarriba: you can use auto: `for (const auto& parameter : parameters_)`

is more C++11 style but at the same time confusing if you don't know that parameters_ is a vector of pointers. At least it will work: http://cpp.sh/82rig

tiny_dnn/layers/layer.h
edgarriba: not sure if it will brake tests this change. It might be worth adding a TODO for this

tiny_dnn/layers/layer.h
```diff
@@ -347,6 +353,67 @@ class layer : public node {
...
+  /**
+   * @brief Get pointers to parameters of this layer.
+   *
+   * This can be used during training phase when parameter update is
+   * required.
+   *
+   * @param trainable_only flag to return only the trainable parameters.
+   * @return std::vector of pointers to parameters
+   */
+  Parameters parameters(bool trainable_only = false) {
```
edgarriba: @karandesai-96 @beru not sure whether we need constness here

tiny_dnn/util/parameter_init.h
karandesai-96: Besides, `TitleCase` for class names and `snake_case` for method names can still be doable. `TitleCase` gives me ability to name `Parameter parameter` while `parameter` class name won't...

karandesai-96: I think atleast Parameter, Tensor, Network and Optimizers should be `TitleCase`

tiny_dnn/layers/layer.h
karandesai-96: @edgarriba right that was a little typo, I fixed it. I preferred being more explicit although shall I replace that ? (You wrote **you "can" use auto** and not "you should" or "you must" xD ).

tiny_dnn/layers/layer.h
```diff
@@ -670,6 +785,33 @@ class layer : public node {
     post_update();
   }
 
+  void update_parameters(optimizer *optimizer_ptr, size_t batch_size) {
+    float_t rcp_batch_size  = float_t{1} / float_t{batch_size};
+    Tensor<float_t> &diff_t = parameters_diff_;
+    for (Parameter *parameter : parameters_) {
+      if (trainable() && parameter->is_trainable()) {
```
edgarriba: you could save some indentation here by doing.
```cpp
if (!trainable() || !parameter->is_trainable()) {
  continue;
}
```
 Up to you, not a request.

tiny_dnn/layers/layer.h
karandesai-96: @edgarriba I thought it might be useful when user is simply pulling out weights for visualizing weights or something. But I am okay with removing it...

tiny_dnn/layers/layer.h
```diff
@@ -670,6 +785,33 @@ class layer : public node {
...
+        parameter->merge_grads(&diff_t);
+
+        for (size_t j = 0; j < diff_t.size(); ++j) {
+          diff_t.host_at(j) *= rcp_batch_size;
```
edgarriba: @Randl just thinking in future now that I see a lots of `host_at`. Should we have a transparent API `at(size_t i)` which internally implements `host_at` and `device_at` ??

tiny_dnn/parameter.h
```diff
@@ -25,34 +25,33 @@ class parameter {
...
    * todo (karandesai) : add an n-dimensional view for easy indexing
    *
-   * @param width      [in] filter width
-   * @param height     [in] filter height
-   * @param depth      [in] filter depth / input channels
-   * @param n_fmaps    [in] number of output feature maps
-   * @param type       [in] whether parameter is a weight or a bias
-   * @param trainable  [in] whether parameter will be updated while training or
-   * not
+   * @param in_channels  filter depth / input channels
```
edgarriba: reorder docs as in the signature

tiny_dnn/util/parameter_init.h
```diff
@@ -0,0 +1,64 @@
...
+  void scale(float_t value) { scale_ = value; }
+
+ protected:
+  float_t scale_;
+};
+
+/**
+ * Use fan-in and fan-out for scaling
+ *
+ * X Glorot, Y Bengio,
```
edgarriba: might be interesting to add proper references to docs

tiny_dnn/layers/layer.h
```diff
@@ -685,6 +827,25 @@ class layer : public node {
     return true;
   }
 
+  bool has_same_parameters(const layer &rhs, float_t eps) const {
```
edgarriba: @karandesai-96 what's the use case of this? I think that will make more sense to overload the bool operator in Tensor.

tiny_dnn/util/parameter_init.h
```diff
@@ -0,0 +1,64 @@
+/*
+    Copyright (c) 2013, Taiga Nomi and the respective contributors
+    All rights reserved.
+
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
+#pragma once
```
edgarriba: https://github.com/alykhantejani/nninit/blob/master/nninit.py
just for reference and future work, you might add a TODO with the link

tiny_dnn/util/parameter_init.h
beru: You guys really like invalidating rules...

tiny_dnn/layers/layer.h
```diff
@@ -347,6 +353,67 @@ class layer : public node {
...
+   *
+   * This can be used during training phase when parameter update is
+   * required.
+   *
+   * @param trainable_only flag to return only the trainable parameters.
+   * @return std::vector of pointers to parameters
+   */
+  Parameters parameters(bool trainable_only = false) {
+    Parameters parameters;
+    for (Parameter *parameter : parameters) {
```
Randl: This is definitely wrong. Did you mean `parameters_`? Why not return `parameters_`?

tiny_dnn/layers/layer.h
edgarriba: up to you

tiny_dnn/layers/layer.h
edgarriba: ops, I'm talking about whether we need to add the const at the end
`Parameters parameters(bool trainable_only = false) const`

karandesai-96: @edgarriba yes we have to, I got an error in `has_same_parameters` .. "passing as non const reference discards qualifiers" or something similar.

tiny_dnn/layers/layer.h
```diff
@@ -49,7 +49,11 @@ class layer : public node {
  public:
   friend void connection_mismatch(const layer &from, const layer &to);
 
-  virtual ~layer() = default;
+  virtual ~layer() {
+    for (auto &p : parameters_) {
+      delete (p);
```
edgarriba: `shared_ptr` :smile:

edgarriba:
@beru @Randl review and approve so that we can move forward

karandesai-96:
Once again to remind, I am agreeing with a few suggestions but working on them would be faster after FC-Parameter integration.

karandesai-96:
I'll add a couple of todos more here in a while.

tiny_dnn/layers/layer.h
karandesai-96: For the current scope of this PR I just replicated `has_same_weights` here

PR #796, opened 2017-07-10
Integration of Tensor into layer

Randl:


tiny_dnn/nodes.h
```diff
@@ -342,15 +342,15 @@ class sequential : public nodes {
...
+    const std::vector<const Tensor<> *> &out) {
     // normalize indexing back to [sample][layer][feature]
     std::vector<tensor_t> normalized_output;
 
     const size_t sample_count = out[0]->size();
     normalized_output.resize(sample_count, tensor_t(1));
 
     for (size_t sample = 0; sample < sample_count; ++sample) {
-      normalized_output[sample][0] = (*out[0])[sample];
+      normalized_output[sample][0] = ((*out[0])[sample]).toVec();
```
prlz77: I think ``toVec()`` has the extra computational cost of making a copy, is there any way to avoid this? Or this is just temporal as it is written in the ``toVec`` method?

tiny_dnn/activations/activation_layer.h
```diff
@@ -66,17 +66,17 @@ class activation_layer : public layer {
     this->in_shape_ = in_shape;
   }
 
-  void forward_propagation(const std::vector<tensor_t *> &in_data,
-                           std::vector<tensor_t *> &out_data) override {
+  void forward_propagation(const std::vector<Tensor<> *> &in_data,
+                           std::vector<Tensor<> *> &out_data) override {
     const tensor_t &x = *in_data[0];
```
karandesai-96: Shouldn't this be changed as well ? Or is this part of code still WIP ?

edgarriba:
@Randl what's the status for this?

edgarriba:
Just a comment for next PR, should we keep the templated type for tensor?

PR #797, opened 2017-07-11
Avepool integration of Tensor

Randl:


edgarriba:
@beru @bhack could you review?

beru:
@edgarriba Getting busy thesedays so no can do....

tiny_dnn/layers/average_pooling_layer.h
```diff
@@ -29,27 +29,28 @@ inline void tiny_average_pooling_kernel(
...
+        for (auto connection : connections) value += in.host_at(connection.second);
         value *= weight;
         value += bias;
-        out[idx] = value;
+        out.host_at(idx) = value;
       }
     }
 
     assert(out.size() == out2wi.size());
+    (*out_data[0])[sample] = out.toVec();
```
prlz77: Is this implying a copy operation?

tiny_dnn/layers/average_pooling_layer.h
```diff
@@ -29,27 +29,28 @@ inline void tiny_average_pooling_kernel(
   float_t scale_factor,
   std::vector<typename partial_connected_layer::wi_connections> &out2wi) {
   for_i(parallelize, in_data[0]->size(), [&](size_t sample) {
-    const vec_t &in = (*in_data[0])[sample];
-    const vec_t &W  = (*in_data[1])[0];
-    const vec_t &b  = (*in_data[2])[0];
-    vec_t &out      = (*out_data[0])[sample];
+    const Tensor<> in((*in_data[0])[sample]);
```
prlz77: Just curiosity, is there a reason for leaving the ``<>`` after ``Tensor``

tiny_dnn/layers/average_pooling_layer.h
```diff
@@ -88,20 +89,26 @@ inline void tiny_average_pooling_back_kernel(
...
-      for (auto o : outs) diff += curr_delta[o];
+      for (auto o : outs) diff += curr_delta.host_at(o);
 
-      db[i] += diff;
+      db.host_at(i) += diff;
     }
+    (*in_grad[1])[sample] = dW.toVec();
+    (*in_grad[2])[sample] = db.toVec();
+    (*in_grad[0])[sample] = prev_delta.toVec();
+    (*out_grad[0])[sample] = curr_delta.toVec();
```
prlz77: I understand all these ``toVec`` won't be necessary in the future because everything will be a Xtensor. Am I right?

edgarriba: right

tiny_dnn/layers/average_pooling_layer.h
edgarriba: it creates an instance of `vec_t` and returns the value
https://github.com/tiny-dnn/tiny-dnn/blob/feat/xtensor_integration/tiny_dnn/core/framework/tensor.h#L383-L388

tiny_dnn/layers/average_pooling_layer.h
edgarriba: it's a workaround to get rid of the templated data type. @Randl new ideas about this ?
we should start thinking about how we define our `BaseTensor` based on dlpack tensor
https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h

PR #803, opened 2017-07-17
Integrate parameter API into fully connected layer.

karandesai-96:
This PR integrates `Parameter` API into fully connected layer. Notable changes made here:

1. `forward_propagation` and `backward_propagation` only receive data as input arguments, and not weights and biases. This implies that `fwd_in_data_`, `fwd_out_data_` will have `std::vector<tensor_t>` of size one.

2. `fully_connected_layer` constructor adds two `Parameter`s in `std::vector<shared_ptr<Parameter>>` of layers (or one, if `has_bias` is not true). Each `Parameter` has its own `initialized` and `trainable` flags.

3. `OpKernel` takes `Parameter`s through `context.ith_parameter()`

Serialization tests of FC layer have been commented out and will be dealt wth later, before it gets merged into master.


test/test_fully_connected_layer.h
```diff
@@ -10,9 +10,51 @@
...
+  vec_t weights = {
+    2, 6, 4,
+    4, 6, 2
+  };
+  vec_t bias = {-4, 22};
+  // clang-format on
+  vec_t expected = {24, 42};
+
+  l.ith_parameter(0).set_data(Tensor<float_t>(weights));
+  l.ith_parameter(1).set_data(Tensor<float_t>(bias));
```
edgarriba: @karandesai-96 @Randl @prlz77 what do you think about having `set_weights(...)` and `set_bias(...)` per layer?
 it will be more informative for the user

karandesai-96: This one is generic should it should stay IMO, but we can definitely add these methods as convenience wrappers.

karandesai-96: And @edgarriba @Randl are we planning to completely eradicate `tensor-t` and `vec_t` after `Tensor` sets in ? If not then I will prefer keeping arguments as `vec_t` / `tensor_t` instead of calling Tensor constructor everywhere.

prlz77: Yep, sounds better as long as it allows for multiple weight matrices inside one layer (set_weights({matrix1, matrix2, ... matrix n}))

karandesai-96:
Please perform rebase of `feat/parameters_refactor` onto `feat/xtensor_integration` to keep mine and @Randl 's work in sync and reduce the diff for reviewing purposes.

test/test_fully_connected_layer.h
```diff
@@ -10,9 +10,51 @@
...
+
+  std::vector<const tensor_t *> out;
+  l.forward({{in}}, out);
```
edgarriba: @karandesai-96 @Randl thinking more stuff, overload forward to return output so that:
`auto out = l.forward({{ in }});`

test/test_parameter.h
```diff
@@ -0,0 +1,69 @@
...
+  EXPECT_EQ(p.shape().depth_, 1u);
+  EXPECT_EQ(p.size(), 9u);
+  EXPECT_EQ(p.type(), parameter_type::weight);
+
+  EXPECT_TRUE(p.is_trainable());
+}
+
+TEST(parameter, getter_setter) {
+  Parameter p(1, 1, 1, 4, parameter_type::bias, false);
+  Tensor<float_t> t{vec_t{1.0f, 2.0f, 3.0f, 4.0f}};
```
edgarriba: `vec_t{1.0f, 2.0f, 3.0f, 4.0f} -> vec_t{1.0, 2.0, 3.0, 4.0}`

test/test_parameter.h
```diff
@@ -0,0 +1,69 @@
...
+  p.set_data(t);
+  Tensor<float_t> *pt = p.data();
+
+  for (size_t i = 0; i < t.size(); i++) {
+    EXPECT_EQ(pt->host_at(i), t.host_at(i));
+  }
+}
+
+TEST(parameter, merge_grads) {
+  Tensor<float_t> gradp{tensor_t{{1.0f, 2.0f},
```
edgarriba: not force float type

tiny_dnn/util/parameter_init.h
```diff
@@ -0,0 +1,69 @@
+/*
```
edgarriba: @karandesai-96 con you add tests for this functionalities ?

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,138 @@
...
+#pragma once
+
+#include "tiny_dnn/core/framework/tensor.h"
+#include "tiny_dnn/util/util.h"
+
+namespace tiny_dnn {
+
+enum class parameter_type : int8_t { weight = 0x1, bias = 0x2 };
+
+class Parameter : public std::enable_shared_from_this<Parameter> {
```
edgarriba: where do you return a shared pointer of Parameter? trying understand why the need of  `std::enable_shared_from_this<Parameter>`

tiny_dnn/parameter.h
```diff
@@ -0,0 +1,138 @@
...
+
+  bool is_trainable() const { return trainable_; }
+
+  void set_trainable(bool trainable = true) { trainable_ = trainable; }
+
+  bool initialized() const { return initialized_; }
+
+  void set_initialized(bool initialized = true) { initialized_ = initialized; }
+
+  Tensor<float_t> *data() { return &data_; }
```
edgarriba: did you try with `std::shared_ptr` ?

edgarriba:
@karandesai-96 would be nice that you could include some benchmarks for each new piece of codes. Less work for the future

tiny_dnn/parameter.h
karandesai-96: I need a shared pointer as I could not take ownership of a shared pointer of weight if I had to set it externally. Doing so required `reset` and it created SIGABRT's whle adding new parameters.

tiny_dnn/parameter.h
karandesai-96: Yes I did, wrote a rough benchmark and concluded that for keeping a hold of parameters, a vector of shared pointer is efficient, since it takes care of memory allocation and destruction when going out of scope. But when we need to manipulate things through method calls, it is best to use raw pointers. Same method with shared pointer was very slightly slow. 

Also it matched with: https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

Hence I kept layer member as `std::vector<std::shared_ptr<Parameter>>` but methods take and return raw pointers.

tiny_dnn/util/parameter_init.h
karandesai-96: @edgarriba sure, althouh I am confused, how can I add tests for xavier I mean they are uniform right ? Fixing random seed ?

test/test_fully_connected_layer.h
edgarriba: @karandesai-96 the idea is remove `tensor_t` and `vec_t`. Since we are not doing a deep copy, it shouldn't be any overload here

edgarriba: @prlz77 I agree, but it can be customized per layer type. Just thinking about user friendly interfaces

tiny_dnn/parameter.h
edgarriba: not sure about const/pros but another solution could be to hold data directly as a pointer

edgarriba: https://stackoverflow.com/questions/1175646/c-when-should-i-use-a-pointer-member-in-a-class

karandesai-96: @edgarriba I had that earlier, we shifted to shared_ptrs to avoid hassles of momery management. We can bring up this discussion separately and think about the possibilities. Luckily it is not a blocker of this PR ;)

tiny_dnn/util/parameter_init.h
karandesai-96: Added :white_check_mark:

test/test_parameter.h
```diff
@@ -62,6 +62,38 @@ TEST(parameter, layer_adder) {
   EXPECT_EQ(parameters[1]->type(), parameter_type::bias);
 }
 
+TEST(parameter, xavier_init) {
+  set_random_seed(42u);
+
+  parameter_init::xavier xvr;
```
edgarriba: use meaningful names `xvr -> xavier`

test/test_parameter.h
```diff
@@ -62,6 +62,38 @@ TEST(parameter, layer_adder) {
...
+                         0.968, 0.341, 0.335, -1.191, -0.187};
+  //clang-format on
+
+  for (size_t i = 0; i < out_result.size(); i++) {
+    EXPECT_NEAR(out_result[i], out_expected[i], 1E-2);
+  }
+}
+
+TEST(parameter, constant_init) {
+  parameter_init::constant cnt(4);
```
edgarriba: use meaningful names `cnt -> constant`

tiny_dnn/util/parameter_init.h
```diff
@@ -0,0 +1,69 @@
...
+ public:
+  xavier() : scalable(float_t{6}) {}
+  explicit xavier(float_t value) : scalable(value) {}
+
+  void fill(Parameter *parameter, size_t fan_in, size_t fan_out) override {
+    const float_t weight_base = std::sqrt(scale_ / (fan_in + fan_out));
+
+    uniform_rand(parameter->data()->host_begin(), parameter->data()->host_end(),
+                 -weight_base, weight_base);
+    parameter->grad()->fill(0);
```
edgarriba: use your own api  `parameter->clear_grads()`

test/test_parameter.h
```diff
@@ -62,6 +62,38 @@ TEST(parameter, layer_adder) {
   EXPECT_EQ(parameters[1]->type(), parameter_type::bias);
 }
 
+TEST(parameter, xavier_init) {
+  set_random_seed(42u);
+
+  parameter_init::xavier xvr;
+  Parameter parameter(1, 1, 1, 10, parameter_type::weight);
+  xvr.fill(&parameter, 1, 1);
```
edgarriba: @karandesai-96 thinking about design, it makes more sense to have a member function `initialize(init_fcn)` in Parameter that receives the function.

tiny_dnn/util/parameter_init.h
karandesai-96: Oops. Right I will

test/test_parameter.h
karandesai-96: Yes it makes sense. I'll try to design that here.

karandesai-96:
There is a huge fall of coverage but I have already started restoring it in subsequent branches.

PR #804, opened 2017-07-17
Convolution parameter integration

karandesai-96:
This PR integrates `Parameter` API into convolution layer. Convolution layer constructor now adds two parameters directly through its constructor and the forward and backward prop method only accept tensors of data, as opposed to data, weights, bias earlier.

Similar to #803.

### DIFF
--- a/test/test_convolutional_layer.h
+++ b/test/test_convolutional_layer.h
@@ -15,29 +15,23 @@ namespace tiny_dnn {
 using namespace tiny_dnn::activation;
 
 TEST(convolutional, setup_internal) {
-  convolutional_layer l(5, 5, 3, 1, 2);
+  convolutional_layer l(5, 5, 3, 1, 2, padding::valid, true);
 
-  EXPECT_EQ(l.parallelize(), true);          // if layer can be parallelized
-  EXPECT_EQ(l.in_channels(), size_t(3));     // num of input tensors
-  EXPECT_EQ(l.out_channels(), size_t(1));    // num of output tensors
-  EXPECT_EQ(l.in_data_size(), size_t(25));   // size of input tensors
-  EXPECT_EQ(l.out_data_size(), size_t(18));  // size of output tensors
-  EXPECT_EQ(l.in_data_shape().size(), size_t(1));   // number of inputs shapes
-  EXPECT_EQ(l.out_data_shape().size(), size_t(1));  // num of output shapes
-  EXPECT_EQ(l.weights().size(), size_t(2));         // the wieghts vector size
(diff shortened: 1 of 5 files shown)


test/test_convolutional_layer.h
```diff
@@ -84,122 +82,128 @@ class tensor_buf {
...
+  vec_t weight = {
+      0.3f,  0.1f,  0.2f,
+      0.0f, -0.1f, -0.1f,
+      0.05f,-0.2f, 0.05f,
 
-  // clang-format off
-  weight[0] = 0.3f;  weight[1] = 0.1f;   weight[2] = 0.2f;
-  weight[3] = 0.0f;  weight[4] = -0.1f;  weight[5] = -0.1f;
-  weight[6] = 0.05f; weight[7] = -0.2f;  weight[8] = 0.05f;
+      0.0f, -0.1f,  0.1f,
```
edgarriba: remove forced float type

test/test_convolutional_layer.h
```diff
@@ -84,122 +78,130 @@ class tensor_buf {
...
 
-  const vec_t &actual_delta = grad.in_at(0)[0];
-  const vec_t &actual_dw    = grad.in_at(1)[0];
-  const vec_t &actual_db    = grad.in_at(2)[0];
+  for (size_t i = 0; i < result_prev_delta.size(); i++) {
+    EXPECT_FLOAT_EQ(expected_prev_delta[i], result_prev_delta[i]);
+  }
 
-  for (size_t i = 0; i < 25; i++) {
-    ASSERT_DOUBLE_EQ(actual_delta[i], expected_prev_delta[i]);
```
edgarriba: this probably breaks precision during tests

test/test_convolutional_layer.h
```diff
@@ -84,122 +78,130 @@ class tensor_buf {
...
-    EXPECT_NEAR(float_t(0.40), out[6], 1E-5);
-    EXPECT_NEAR(float_t(1.15), out[7], 1E-5);
-    EXPECT_NEAR(float_t(0.80), out[8], 1E-5);
+  for (size_t i = 0; i < result.size(); i++) {
+    EXPECT_NEAR(expected[i], result[i], 1E-5);
   }
 }
 
 TEST(convolutional, with_stride) {
-  /*
```
edgarriba: keep this comment just to have an ilustrative picture about what's happening. Would be nice for backprop too

test/test_convolutional_layer.h
karandesai-96: right, but does breaking precision depend on gcc version and os ?

PR #805, opened 2017-07-17
Add OpKernel classes for deconv2d layer and remove old kernels.

karandesai-96:
This PR refactors the underlying kernels of deconvolution layer and makes them uniform with existing kernels such as fully_connected, conv kernels. These kernels are no longer called from backend class, but through an OpKernel during forward and backward propagation.


karandesai-96:
This one is necessary for parameters refactor.

tiny_dnn/layers/deconvolutional_layer.h
```diff
@@ -228,8 +225,17 @@ class deconvolutional_layer : public layer {
...
   void forward_propagation(const std::vector<tensor_t *> &in_data,
                            std::vector<tensor_t *> &out_data) override {
-    // launch deconvolutional kernel
-    layer::backend_->deconv2d(in_data, out_data);
+    dws_.prev_out_ = in_data[0];
+    tensor_t &out  = *out_data[0];
+
+    fwd_ctx_.set_in_out(in_data, out_data);
+    fwd_ctx_.setParallelize(layer::parallelize());
+    fwd_ctx_.setEngine(layer::engine());
```
edgarriba: BTW, keep in mind that `layer::engine()` should be replaced by `layer::backend_type()` once we get rid of the old backend structure

karandesai-96: Sure, noted.

tiny_dnn/core/kernels/deconv2d_op.h
```diff
@@ -0,0 +1,39 @@
...
+
+    // incomimg / outcoming data
+    const Tensor<float_t> in_data(context.input(0));
+    const Tensor<float_t> weights(context.input(1));
+    const Tensor<float_t> bias =
+      params.has_bias ? Tensor<float_t>(context.input(2)) : Tensor<float_t>();
+    Tensor<float_t> out_data(context.output(0));
+
+    // initialize outputs
+    out_data.fill(0.0f);
```
edgarriba: you shouldn't use float literals since `float_t` can be either float or double. Keep it for next PR

tiny_dnn/layers/deconvolutional_layer.h
```diff
@@ -249,7 +255,16 @@ class deconvolutional_layer : public layer {
                         const std::vector<tensor_t *> &out_data,
                         std::vector<tensor_t *> &out_grad,
                         std::vector<tensor_t *> &in_grad) override {
-    layer::backend_->deconv2d(in_data, out_data, out_grad, in_grad);
+    if (params_.pad_type == padding::same)
```
edgarriba: missing if statement brackets

tiny_dnn/core/kernels/deconv2d_op.h
karandesai-96: Okay doing it in #814, it is related to deconv layer too.

tiny_dnn/layers/deconvolutional_layer.h
karandesai-96: Lint pre commit hook has definitely made me careless xD

PR #806, opened 2017-07-17
RNN API

prlz77:
This PR introduces:

## A simple RNN API
- Basic recurrent units are called ``cells``, which inherit from ``cell.h``.
- Cells can perform multiple timesteps.
- A wrapper ``layer`` then wraps the cell to manage the states in multiple time steps. ``recurrent_layer.h`` is the most basic example but ``bidirectional_recurrent_layer.h`` is also possible.
- Tests and serialization are provided.
- ``cells.h`` makes it easier for the user to create new recurrent models: from ``recurrent_layer(std::make_shared<rnn_cell>(rnn_cell(in, out)), timesteps)`` to ``recurrent_layer(rnn(in, out), timesteps)``

## LSTM
- ``lstm_cell.h`` with the corresponding internal backend, serialization and tests.

## GRU
- ``gru_cell.h`` with the corresponding internal backend, serialization and tests.

## Examples
- Add 2 numbers example
- Char Rnn trained on gitter, and gitter bot using python wrappers with SWIG.

## Minor Fixes
- quantized_fully serialize function has been restored.
- gradient checks.

## Side Effects
* ``fprop`` and ``bprop`` have been made public to allow online data augmentation and training on big datasets that do not fit in memory. Example:
    1. Load batch & Augmentation
    2. fprop.
    3. bprop.
    4. Do things to the gradients (like gradient clipping).
    5. Update
    6. Go to 1


test/test_lstm_cell.h
```diff
@@ -0,0 +1,159 @@
...
+  network<sequential> nn;
+  adagrad optimizer;
+
+  nn << recurrent_layer(lstm(3, 2, true), 1);
+  nn.weight_init(weight_init::xavier());
+
+  vec_t a(3), t(2), a2(3), t2(2);
+
+  // clang-format off
+    a[0] = 3.0f; a[1] = 0.0f; a[2] = -1.0f;
```
edgarriba: you are forcing float
`a[0] = 3.0f -> a[0] = 3.0` and so on

prlz77: Aren't the rest of the tests forcing float? Or it is going to change everywhere?

edgarriba: `float_t` can be either `float` or `double`. It's controled at compilation time by `CNN_USE_DOUBLE`
check: https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/config.h#L88-L92

tiny_dnn/layers/gru_cell.h
```diff
@@ -0,0 +1,178 @@
...
+ *
+ **/
+class gru_cell : public cell {
+ public:
+  /**
+   * @param in_dim [in] number of elements of the input
+   * @param out_dim [in] number of elements of the output
+   * @param params [in] optional gru cell parameters
+   **/
+  gru_cell(size_t in_dim, size_t out_dim, const gru_cell_parameters params = gru_cell_parameters()) {
```
edgarriba: pass parameters by reference not by value

test/test_gru_cell.h
```diff
@@ -0,0 +1,165 @@
...
+    network<sequential> nn;
+    adagrad optimizer;
+
+    nn << recurrent_layer(gru(3, 2), 1) << fully_connected_layer(2, 2);
+    nn.weight_init(weight_init::xavier());
+
+    vec_t a(3), t(2), a2(3), t2(2);
+
+    // clang-format off
+a[0] = 3.0; a[1] = 0.0; a[2] = -1.0;
```
edgarriba: wrong identation, be aware of that `// clang-format off`

prlz77: :+1: anyway I put the work in progress  label thinking that you wouldn't look at it yet :wink:

test/test_gru_cell.h
```diff
@@ -0,0 +1,165 @@
...
+
+TEST(gru, train) {
+  network<sequential> nn;
+  adagrad optimizer;
+  recurrent_layer_parameters params;
+  params.reset_state = false;
+  nn << recurrent_layer(gru(3, 2), 1, params);
+  nn.weight_init(weight_init::xavier());
+
+  vec_t a(3), t(2), a2(3), t2(2);
```
edgarriba: same as @karandesai-96 in https://github.com/tiny-dnn/tiny-dnn/pull/834, could implement a class `rnn_test` ? a lot a of duplicated code in tests

examples/recurrent_addition/addition.cpp
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * Example of using recurrent neural networks for adding two
+ * numbers. The dataset consists of random pairs of numbers in
+ * the range -1,1.
+ *
+ */
+template <typename N>
+void construct_net(N &nn,
+                   const std::string rnn_type,
+                   tiny_dnn::core::backend_t backend_type) {
```
edgarriba: if you don't use the `backend_type` remove it

examples/recurrent_addition/addition.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  std::cout << "start learning" << std::endl;
+
+  tiny_dnn::progress_display disp(n_samples);
+  tiny_dnn::timer t;
+
+  optimizer.alpha = learning_rate;
+
+  tiny_dnn::recurrent_layer *layer =
+    static_cast<tiny_dnn::recurrent_layer *>(nn[0]);
```
edgarriba: why do you need this cast here ?

test/test_gru_cell.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <gtest/gtest.h>
+
+#include <functional>
+#include <vector>
+
+#include "test/testhelper.h"
+#include "tiny_dnn/tiny_dnn.h"
+
+namespace tiny_dnn {
```
edgarriba: a test just with a forward pass would help also to understand how to use the api

tiny_dnn/util/gradient_check.h
```diff
@@ -52,8 +52,8 @@ float_t numeric_gradient(layer &layer,
                          std::vector<tensor_t> in_data,  //  copy is safer
                          const size_t in_edge,
                          const size_t in_pos,
-                         std::vector<tensor_t> &out_data,
-                         std::vector<tensor_t> &out_grads,
+                         std::vector<tensor_t> out_data,
```
edgarriba: why by value ?

examples/recurrent_addition/addition.cpp
prlz77: I order to call the following functions afterwards:
```
    // set rnn to element-wise prediction (without forget)
    layer->seq_len(1);
    // clear current recurrent state
    layer->clear_state();
```

tiny_dnn/util/gradient_check.h
prlz77: Because the same out grads might be used in other tests and we don't want this function to modify them. Also, since this function is only called on tests, it is computationally safe to use a pass by value here.

examples/recurrent_addition/addition.cpp
nyanp: We can use ```nn.at<recurrent_layer>(0)``` instead.

edgarriba:
@nyanp @Randl @karandesai-96 please review

test/test_gru_cell.h
karandesai-96: also, I think there should be a `setup` test which clearly highlights the attributes. Similar tests are written for conv layer and other too.

test/test_serialization.h
```diff
@@ -271,37 +271,101 @@ TEST(serialization, serialize_dropout) {
...
+                    "out_size": 20,
+                    "has_bias": true
+                }
+            ]
+        }
+        )";
+
+  net.from_json(json);
+
+  EXPECT_EQ(net[0]->layer_type(), "gru-cell");
```
karandesai-96: We should really standardize whether we are using underscores or hyphens. Layer types give hyphens and serialization uses underscores. But it is not a topic of this PR, as it exists in all layers.
/cc: @edgarriba

test/test_gru_cell.h
prlz77: I agree, for every individual thing of layer. I think this should be done in another PR though, this one is big enough. Should I open an issue or something like that to define the task?

test/test_serialization.h
prlz77: Actually I've had a lot of merge conflicts because of this in both ways (from - to _ and vice-versa)

edgarriba: hyphens for string looks better to me. But open to strongest opinions

costashatz:
Hey guys, 

Thanks for the nice repo and this useful PR. When are you planning to merge it?

prlz77:
Hi @costashatz the main devs are a little bit bussy right now but I think this PR is going to be merged as soon as they have time :)

edgarriba:
@nyanp @karandesai-96 @bhack @beru @Randl please review so that we can merge

bhack:
Can we merge this?

tiny_dnn/tiny_dnn.h
```diff
@@ -100,7 +103,7 @@ using fc = tiny_dnn::fully_connected_layer;
 
 using dense = tiny_dnn::fully_connected_layer;
 
-using recurrent_cell = tiny_dnn::recurrent_cell_layer;
+// using rnn_cell = tiny_dnn::rnn_cell_layer;
```
beru: Is this a dead code?

tiny_dnn/core/kernels/lstm_cell_op_internal.h
```diff
@@ -0,0 +1,313 @@
...
+                                  const vec_t &b_2o,
+                                  tensor_t &out_data,
+                                  tensor_t &h_next,
+                                  tensor_t &c_next,
+                                  tensor_t &i,
+                                  tensor_t &f,
+                                  tensor_t &z,
+                                  tensor_t &c,
+                                  const core::lstm_cell_params &params,
+                                  const bool layer_parallelize) {
```
beru: Wow, so many (24!) arguments.

tiny_dnn/core/kernels/lstm_cell_op.h
```diff
@@ -0,0 +1,76 @@
...
+    fill_tensor(i, float_t{0});
+    fill_tensor(f, float_t{0});
+    fill_tensor(z, float_t{0});
+    fill_tensor(c, float_t{0});
+
+    // call the algorithm depending  on the selected engine type
+
+    const core::backend_t engine = context.engine();
+
+    if (engine == core::backend_t::internal || engine == core::backend_t::avx) {
```
beru: I don't think this condition is needed.

tiny_dnn/core/kernels/gru_cell_op.h
```diff
@@ -0,0 +1,71 @@
...
+    fill_tensor(r, float_t{0});       // reset gate    r(t)
+    fill_tensor(z, float_t{0});       // update gate   z(t)
+    fill_tensor(hr, float_t{0});      // aux state  hr(t)
+    fill_tensor(post_z, float_t{0});  // aux state  - post_z(t) (1-z)
+
+    // call the algorithm depending  on the selected engine type
+
+    const core::backend_t engine = context.engine();
+
+    if (engine == core::backend_t::internal || engine == core::backend_t::avx) {
```
beru: I think this if statement is not necessary.

What if someone tries to add a new backend in the future? Should the person check this statement as well?

test/test_serialization.h
beru: In my opnion, layer type strings should be unified.

And while it's OK to have this virtual method,
```cpp
virtual std::string layer::layer_type() const = 0;
```

It should be possible to retrieve layer type name statically for serialization.
In that way, we don't need to use different names in `serialization_layer_list.h`.

tiny_dnn/core/kernels/gru_cell_op.h
prlz77: I just followed the convention in https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/fully_connected_op.h, https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/core/kernels/conv2d_op.h
I mean, check the backend and if does not match, throw error. Maybe we can open a PR and change all of them.

beru: I totally agree that this should be handled in a different PR.

Maybe you're right. Since you are the one who implemented those ops.
You have probablly tested with internal and avx backends but haven't tested with other backends.
Then above code should be fine.

edgarriba:
nice !

prlz77:
😃

karandesai-96:
Great, one down two to go! :smile:

PR #811, opened 2017-07-19
Shift to container based builds on travis.

kdexd:
This PR introduces changes in `.travis.yml` file. Linux jobs required addition of a line in sources list, as well as pulling a gpg key for llvm. Both were done by a bash script ( `.travis/install.sh` ) which used sudo.

Usage of `sudo` enforces travis to use legacy infrastructure and it is comparatively slower. To quickly start the builds on pushing, we must migrate to container based bulds which are through `sudo: false`.

I turned off sudo and performed the earlier tasks directly by specifying suitable config options in yml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 dist: trusty
 osx_image: xcode8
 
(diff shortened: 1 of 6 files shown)


edgarriba:
@karandesai-96 any update on this

karandesai-96:
@edgarriba No I was working on benchmarks. Let me look into this today, I will implement your idea - will separate out serialization into a separate job as well.

nyanp:
@edgarriba @karandesai-96 
Is it replaced by #837 ?

edgarriba:
@nyanp different approaches I guess. I've been playing with this new feature to build in travis in stages but seems quite new and not very flexible yet since apparently it doesn't allow matrices. Besides, not really sure how we could "dockerize" or at least have a fixed development environment for osx. Additionally, I'd would like to discuss about testing gcc and clang because since the more we increase the backends the bigger is the matrix and we end up will very large timings.

PR #812, opened 2017-07-19
update Dockerfile

edgarriba:
updates the Dockerfile with all the tools needed for development


docker/dev-env/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+# install optional dependencies
+
+RUN apt-get install -y    \
+    libpthread-stubs0-dev \
+    libtbb-dev
+
+# install linters
+
+RUN apt-get install -y    \
+    clang-format-4.0
```
Randl: I think you need to add LLVM 4.0 from http://apt.llvm.org/  to sources

edgarriba: no need for that. I think messed up during push the docker image. Will try again today

edgarriba:
I think that we can merge this

docker/dev-env/Dockerfile
prlz77: Is clang the default compiler in the docker? Otherwise you could set it up as default.

edgarriba: I also prefer  clang but any advantage for doing this ?

nyanp:
Seems good, all we need is rebase 👍

edgarriba:
that's just a proposal, not sure that we really need one image per backend. Probably just providiing a minimal environment with tbb and openmp + linters, and then a another one with all the dependencies included.

docker/dev-env/Dockerfile
prlz77: Didn't you tell me that clang shows more useful debug information and catches more warnings?

edgarriba:
@bhack @Randl @nyanp @karandesai-96 @prlz77 gonna merge this in a few hours

bhack:
It is ok for now we can improve as we need

PR #786, opened 2017-06-30
Remove old backend and convt to ops

edgarriba:

PR #814, opened 2017-07-19
Integrate parameter API into deconvolution layer.

karandesai-96:
This PR integrates `Parameter` API into deconvolution layer. Deconvolution layer constructor now adds two parameters directly through its constructor and the forward and backward prop method only accept tensors of data, as opposed to data, weights, bias earlier.

Similar to #803, #804.


karandesai-96:
Removing those `'f'` and preventing force floats.

test/test_deconvolutional_layer.h
```diff
@@ -41,164 +42,147 @@ TEST(deconvolutional, setup_nnp) {
...
+  EXPECT_EQ(l.ith_parameter(1).size(), 2u);   // size of bias parameter
   EXPECT_STREQ(l.layer_type().c_str(), "deconv");  // string with layer type
 }
 #endif
 
 TEST(deconvolutional, fprop) {
-  network<sequential> nn;
-
   deconvolutional_layer l(2, 2, 3, 1, 2);
+  l.setup(false);
```
edgarriba: I'm wondering here how we could make this lazy, since from
https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/layers/layer.h#L591

I see that all the time allocate memory for the graph

test/test_deconvolutional_layer.h
```diff
@@ -41,164 +42,147 @@ TEST(deconvolutional, setup_nnp) {
...
-  in_data.push_back(&weight_tensor);
-  in_data.push_back(&bias_tensor);
   out_data.push_back(&out_tensor);
-  l.setup(false);
-  {
-    l.forward_propagation(in_data, out_data);
-    vec_t &out = out_tensor[0];
+  l.ith_parameter(0).set_data(Tensor<float_t>(weight));
+  l.ith_parameter(1).set_data(Tensor<float_t>(bias));
+  l.forward_propagation(in_data, out_data);
```
edgarriba: for consistency can we use `forward(...)` and leave `forward_propagation(...)` for internals ?

tiny_dnn/core/kernels/deconv2d_grad_op.h
```diff
@@ -0,0 +1,49 @@
...
+    in the LICENSE file.
+*/
+#pragma once
+
+#include "tiny_dnn/core/framework/op_kernel.h"
+#include "tiny_dnn/core/kernels/deconv2d_op_internal.h"
+
+namespace tiny_dnn {
+
+class Deconv2dGradOp : public core::OpKernel {
```
edgarriba: better name for this: `Conv2dTranposed`

see: https://github.com/tiny-dnn/tiny-dnn/pull/786#discussion_r125084038

tiny_dnn/core/kernels/deconv2d_op_internal.h
```diff
@@ -10,22 +10,77 @@
...
-namespace core {
 namespace kernels {
 
+template <typename S1, typename S2, typename S3, typename S4>
+inline void deconv2d_op_internal(const Tensor<float_t, S1> &in_data,
+                                 const Tensor<float_t, S2> &weights,
+                                 const Tensor<float_t, S3> &bias,
+                                 Tensor<float_t, S4> &out_data,
+                                 const deconv_params &params,
+                                 const bool layer_parallelize) {
```
edgarriba: better naming `layer_parallelize -> parallelize`

tiny_dnn/core/kernels/deconv2d_op_internal.h
```diff
@@ -10,22 +10,77 @@
...
+                pout[(y * params.h_stride + wy) * params.out.width_ +
+                     (x * params.w_stride + wx)] +=
+                  ppw[wy * params.weight.width_ + wx] * (*ppi);
+              }
+            }
+          }
+        }
+      }
+
+      if (params.has_bias) {
```
edgarriba: I would include `has_bias` as part of the signature. The idea is to get rid of this data structure in the future since most of the encoded information is part of the tensor nature now

tiny_dnn/core/kernels/deconv2d_op.h
```diff
@@ -0,0 +1,44 @@
+/*
+    Copyright (c) 2017, Taiga Nomi
```
prlz77: I think it is now _Copyright (c) 2013, Taiga Nomi and the respective contributors_

tiny_dnn/layers/deconvolutional_layer.h
```diff
@@ -198,7 +214,12 @@ class deconvolutional_layer : public layer {
                         size_t w_stride        = 1,
                         size_t h_stride        = 1,
                         backend_t backend_type = core::default_engine())
-    : layer(std_input_order(has_bias), {vector_type::data}) {
+    : layer({vector_type::data}, {vector_type::data}) {
+    layer::add_parameter(out_channels, in_channels, window_height, window_width,
+                         parameter_type::weight);
```
prlz77: Seems strange to have vector_type and parameter_type

tiny_dnn/core/kernels/deconv2d_grad_op.h
```diff
@@ -0,0 +1,52 @@
...
+                                 ? *(context.ith_parameter(1)->grad())
+                                 : Tensor<float_t>());
+
+    // initialize outputs
+    prev_delta.fill(float_t{0});
+
+    kernels::deconv2d_op_internal(prev_out, weights, weights_grads, bias_grads,
+                                  curr_delta, prev_delta, params,
+                                  params.has_bias, context.parallelize());
+    context.ith_parameter(0)->set_grad(weights_grads);
```
prlz77: Why is ``context.ith_parameter(0)->set_grad(weights_grads);`` needed?

tiny_dnn/core/kernels/deconv2d_op.h
edgarriba: right !

tiny_dnn/layers/deconvolutional_layer.h
karandesai-96: Yes, since I am doing parameter integration layer by layer, I am duplicating older methods and data structures with similar names. Once all layers will have it integrate, a huge chunk of code will vanish / get merged with its duplicate counterpart.. This is just a temporary transient phase and not a final API :wink: Please do let me know if you vision some idea about how final Parameter API should look!

tiny_dnn/core/kernels/deconv2d_grad_op.h
karandesai-96: Because `toVec` and `toTensor` do a copy and return a new object. That is passed by reference to ernels and it updated. The older source needs to be updated by copy assignment.

PR #815, opened 2017-07-20
Make API for getting and setting parameters more convenient.

karandesai-96:
This PR introduces certain convenient getters setters for avoiding confusions. Generic getters setters related to parameters continue to be there for internal usage (kernels and optimizers).

### DIFF
--- a/test/test_deconvolutional_layer.h
+++ b/test/test_deconvolutional_layer.h
@@ -19,22 +19,22 @@ TEST(deconvolutional, setup_tiny) {
   deconvolutional_layer l(2, 2, 3, 1, 2, padding::valid, true, 1, 1,
                           core::backend_t::internal);
 
-  EXPECT_EQ(l.parallelize(), true);          // if layer can be parallelized
-  EXPECT_EQ(l.in_channels(), 1u);            // num of input tensors
-  EXPECT_EQ(l.out_channels(), 1u);           // num of output tensors
-  EXPECT_EQ(l.in_data_size(), 4u);           // size of input tensors
-  EXPECT_EQ(l.out_data_size(), 32u);         // size of output tensors
-  EXPECT_EQ(l.in_data_shape().size(), 1u);   // number of inputs shapes
-  EXPECT_EQ(l.out_data_shape().size(), 1u);  // num of output shapes
-  EXPECT_EQ(l.inputs().size(), 1u);          // num of input edges
-  EXPECT_EQ(l.outputs().size(), 1u);         // num of outpus edges
-  EXPECT_EQ(l.in_types().size(), 1u);        // num of input data types
-  EXPECT_EQ(l.out_types().size(), 1u);       // num of output data types
-  EXPECT_EQ(l.fan_in_size(), 9u);            // num of incoming connections
-  EXPECT_EQ(l.fan_out_size(), 18u);          // num of outgoing connections
-  EXPECT_EQ(l.parameters().size(), 2u);      // num of trainable parameters
-  EXPECT_EQ(l.weights_at().size(), 18u);     // size of weight parameter
-  EXPECT_EQ(l.bias_at().size(), 2u);         // size of bias parameter
+  EXPECT_EQ(l.parallelize(), true);           // if layer can be parallelized
+  EXPECT_EQ(l.in_channels(), 1u);             // num of input tensors
+  EXPECT_EQ(l.out_channels(), 1u);            // num of output tensors
+  EXPECT_EQ(l.in_data_size(), 4u);            // size of input tensors
+  EXPECT_EQ(l.out_data_size(), 32u);          // size of output tensors
+  EXPECT_EQ(l.in_data_shape().size(), 1u);    // number of inputs shapes
+  EXPECT_EQ(l.out_data_shape().size(), 1u);   // num of output shapes
+  EXPECT_EQ(l.inputs().size(), 1u);           // num of input edges
+  EXPECT_EQ(l.outputs().size(), 1u);          // num of outpus edges
+  EXPECT_EQ(l.in_types().size(), 1u);         // num of input data types
+  EXPECT_EQ(l.out_types().size(), 1u);        // num of output data types
+  EXPECT_EQ(l.fan_in_size(), 9u);             // num of incoming connections
+  EXPECT_EQ(l.fan_out_size(), 18u);           // num of outgoing connections
+  EXPECT_EQ(l.parameters().size(), 2u);       // num of trainable parameters
+  EXPECT_EQ(l.weights_at()[0]->size(), 18u);  // size of weight parameter
+  EXPECT_EQ(l.bias_at()[0]->size(), 2u);      // size of bias parameter
   EXPECT_STREQ(l.layer_type().c_str(), "deconv");  // string with layer type
 }
 
@@ -107,7 +107,7 @@ TEST(deconvolutional, fprop) {
   };
   // clang-format on
 
-  l.weights_at().set_data(Tensor<float_t>(weight));
+  l.weights_at()[0]->set_data(Tensor<float_t>(weight));
   out        = l.forward({{in}});
   out_result = (*out[0])[0];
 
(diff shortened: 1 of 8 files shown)


test/test_deconvolutional_layer.h
```diff
@@ -10,29 +10,28 @@
...
-
 namespace tiny_dnn {
 
 TEST(deconvolutional, setup_tiny) {
   deconvolutional_layer l(2, 2, 3, 1, 2, padding::valid, true, 1, 1,
                           backend_t::internal);
 
   EXPECT_EQ(l.parallelize(), true);          // if layer can be parallelized
-  EXPECT_EQ(l.in_channels(), 3u);            // num of input tensors
+  EXPECT_EQ(l.in_channels(), 1u);            // num of input tensors
```
edgarriba: @karandesai-96 what makes that difference ?

test/test_deconvolutional_layer.h
```diff
@@ -42,158 +41,125 @@ TEST(deconvolutional, setup_nnp) {
...
+  EXPECT_EQ(l.biases_at(0).size(), 2u);      // size of bias parameter
   EXPECT_STREQ(l.layer_type().c_str(), "deconv");  // string with layer type
 }
 #endif
 
 TEST(deconvolutional, fprop) {
-  network<sequential> nn;
-
   deconvolutional_layer l(2, 2, 3, 1, 2);
+  l.weight_init_f(parameter_init::constant(0.0));
```
edgarriba: sweet ! however, IMHO just `l.weight_init(...)` it's more clean

tiny_dnn/layers/deconvolutional_layer.h
```diff
@@ -198,7 +214,12 @@ class deconvolutional_layer : public layer {
                         size_t w_stride        = 1,
                         size_t h_stride        = 1,
                         backend_t backend_type = core::default_engine())
-    : layer(std_input_order(has_bias), {vector_type::data}) {
+    : layer({vector_type::data}, {vector_type::data}) {
+    layer::add_parameter(out_channels, in_channels, window_height, window_width,
+                         parameter_type::weight);
```
edgarriba: actually, I think that whatever data allocation done in `layer({vector_type::data}, {vector_type::data})` should happen when you add a parameter to the layer. Additionally, we should be able to control whether we use imported wieghts. It's not clear to me what this `add_parameter(...)` does: allocates data?

test/test_fully_connected_layer.h
```diff
@@ -44,8 +44,8 @@ TEST(fully_connected, forward) {
   // clang-format on
   vec_t expected = {24, 42};
 
-  l.ith_parameter(0).set_data(Tensor<float_t>(weights));
-  l.ith_parameter(1).set_data(Tensor<float_t>(bias));
+  l.weights_at(0).set_data(Tensor<float_t>(weights));
```
edgarriba: I wouldn't expose `weights_at(...)` to fully connected since it only has data at 0 position, it might lead to bugs. Expose a single method to get weights and bias:
```cpp
tensor fully_connected_layer::weights() const {
  return weights_at(0);
}
```
same for bias
```cpp
tensor fully_connected_layer::bias() const {
  return biases_at(0);
}
```

tiny_dnn/layers/layer.h
```diff
@@ -404,48 +404,94 @@ class layer : public node {
...
 
-  Parameter &ith_parameter(size_t i) { return *parameters_[i]; }
+  /**
+   * @brief Get pointers to weights of this layer.
+   *
+   * @param trainable_only flag to return only the trainable weights
+   * @return std::vector of pointers to parameters which are weights
+   */
+  // todo (karandesai) : rename after old weights method is removed
+  Parameters weights_p(bool trainable_only = false) {
```
edgarriba: I see a lot of duplicated code, cool solution for this
https://stackoverflow.com/questions/9034388/c-how-to-make-find-if-not-only-return-the-first-object-for-which-applying-pred?answertab=votes#tab-top

tiny_dnn/layers/layer.h
```diff
@@ -404,48 +404,94 @@ class layer : public node {
    * This can be used during training phase when parameter update is
    * required.
    *
-   * @param trainable_only flag to return only the trainable parameters.
+   * @param trainable_only flag to return only the trainable parameters
    * @return std::vector of pointers to parameters
    */
   Parameters parameters(bool trainable_only = false) {
     Parameters parameters;
     for (size_t i = 0; i < parameters_.size(); i++) {
```
edgarriba: why not using foreach ?

tiny_dnn/layers/layer.h
```diff
@@ -404,48 +404,94 @@ class layer : public node {
...
+    for (size_t i = 0; i < p.size(); i++) {
+      if (p[i]->type() != parameter_type::weight) p.erase(p.begin() + i);
+    }
+    return p;
+  }
+
+  ConstParameters weights_p(bool trainable_only = false) const {
+    ConstParameters p = parameters(trainable_only);
+    for (size_t i = 0; i < p.size(); i++) {
+      if (p[i]->type() != parameter_type::weight) p.erase(p.begin() + i);
```
edgarriba: this function is O(2n) while you could do it in O(n) if you specify in parameters the type you want. Besides, since there is no `weights(...)` I don't see the point of naming `weights_p(...)`.

test/test_deconvolutional_layer.h
karandesai-96: @edgarriba:
`in_channels` were 3 earlier - `{vector_type::data, vector_type::weight, vector_type::bias}`. But now only `{vector_type::data}`.

Essentially I would envision all layers to give in and pass out `data` and `aux` vector types later.

test/test_deconvolutional_layer.h
karandesai-96: It is conflicting with existing method of same name and since quantization layers still don't have parameters, I need to keep both methods. I have added a todo to remove old method and rename this one once all layers have parameters integrated or once quantization disabling PR is merged and this one is rebased on top of it.

tiny_dnn/layers/deconvolutional_layer.h
karandesai-96: Yes it essentially allocates memory to data and grad depending on the dimensions. But it allocates memory to `Parameter` struct, I think `xtensor` itself stays uninitialized unless called upon.

Unfortunately I don't clearly remember how it worked out, @Randl a little help here ? 
Here is `add_parameter` body to save you from navigating:
```c++
  /** @brief Add new parameter to this layer, to be called in constructor. */
  void add_parameter(size_t out_channels,
                     size_t in_channels,
                     size_t height,
                     size_t width,
                     parameter_type type,
                     bool trainable = true) {
    parameters_.push_back(std::make_shared<Parameter>(
      out_channels, in_channels, height, width, type, trainable));
  }
```

test/test_deconvolutional_layer.h
edgarriba: IC

tiny_dnn/layers/deconvolutional_layer.h
edgarriba: but also `layer({vector_type::data}, {vector_type::data})` is allocating data right ?

karandesai-96:
I will rename `weights_at()` -> `weights()` and bias as well, right now there are conflicting names so I am not taking a risk.

tiny_dnn/layers/layer.h
```diff
@@ -407,48 +407,77 @@ class layer : public node {
...
 
-  void set_ith_parameter(size_t i, Parameter &p) {
-    if (i >= parameters_.size()) parameters_.resize(i + 1);
-    parameters_[i] = std::make_shared<Parameter>(p);
+  /** Convenience method to access and manipulate weights. */
+  Parameter &weights_at() {
+    Parameter *weights_p = nullptr;
+    bool fetched         = false;
+    for (auto &parameter : parameters_) {
+      if (parameter->type() == parameter_type::weight) {
```
edgarriba: I don't get is this needed

karandesai-96: RNNs have multiple weights. There might be other layers as well. Well we can provide methods to get `W`, `U`, `V` separately which can be implemented in child class (rnn), but if someone calls `rnn_layer.weights()` then he/she should get a warning in my opinion.
CC: @prlz77

prlz77: I think the parameter setter/getter should return a vector of parameters directly

edgarriba: agree with @prlz77

PR #816, opened 2017-07-22
Restructure kernels of average pooling layer.

karandesai-96:
This PR declares `avepool_params` and `average_pooling_worker_specific_storage` and passes them accordingly to kernels. Although we are planning to get rid of `params` since most of the information is in `Tensor`, it is beneficial to keep uniformity for easy migration further.

This helps in integrating parameter API within easily.

### DIFF
--- a/test/test_average_pooling_layer.h
+++ b/test/test_average_pooling_layer.h
@@ -101,12 +101,11 @@ TEST(ave_pool, forward) {
   l.bias_init(weight_init::constant(0.0));
   l.init_weight();
 
-  std::vector<const tensor_t*> out;
-  l.forward({{in}}, out);
+  auto out = l.forward({{in}});
   vec_t res = (*out[0])[0];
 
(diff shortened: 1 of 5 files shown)


test/test_average_pooling_layer.h
```diff
@@ -140,6 +140,60 @@ TEST(ave_pool, forward_stride) {
...
+  // clang-format on
+
+  std::vector<tensor_t> in_grad =
+    l.backward(std::vector<tensor_t>{{curr_delta}});
+  vec_t prev_delta_result = in_grad[0][0];
+  vec_t dw_result         = in_grad[1][0];
+  vec_t db_result         = in_grad[2][0];
+
+  for (size_t i = 0; i < prev_delta_expected.size(); i++) {
+    EXPECT_FLOAT_EQ(prev_delta_result[i], prev_delta_expected[i]);
```
edgarriba: double check that `EXPECT_FLOAT_EQ` doesn't break precision

PR #818, opened 2017-07-23
Generalize gradient check for multiple weight matrices

prlz77:


tiny_dnn/network.h
```diff
@@ -486,33 +486,34 @@ class network {
...
       }
-      vec_t &w     = *current->weights()[0];
-      vec_t &b     = *current->weights()[1];
-      tensor_t &dw = (*current->weights_grads()[0]);
-      tensor_t &db = (*current->weights_grads()[1]);
+      std::vector<vec_t *> weights     = current->weights();
+      std::vector<tensor_t *> grads = current->weights_grads();
 
-      if (w.empty()) continue;
+      if ((*weights[0]).empty()) continue;
```
nyanp: ```if (weights.empty() || (*weights[0]).empty()) continue;``` is preferred in case of ```current->weights()``` returns empty vector.

bhack:
You need to run clang-format

tiny_dnn/network.h
```diff
@@ -486,33 +486,34 @@ class network {
...
       }
-      vec_t &w     = *current->weights()[0];
-      vec_t &b     = *current->weights()[1];
-      tensor_t &dw = (*current->weights_grads()[0]);
-      tensor_t &db = (*current->weights_grads()[1]);
+      std::vector<vec_t *> weights     = current->weights();
+      std::vector<tensor_t *> grads = current->weights_grads();
 
-      if (w.empty()) continue;
+      if (weights.empty() || (*weights[0]).empty()) continue;
```
edgarriba: @nyanp what would make that weights is not empty but yes first indexed element ?

nyanp: @edgarriba
Ah probably ```if (weights.empty())``` is enough in this case. Checking emptiness of each elements can be performed in calc_delta.

prlz77: Should I change it then? I can do it in just a moment but I think it is correct both ways.

edgarriba:
@prlz77 use clang-format so that we can merge

Randl:
Problems with training? Looks strange

prlz77:
@Randl It seems a precission problem, I think this can be solved just by training for more iterations.

edgarriba:
@prlz77 does #846 fix this ? otherwise we can merge it

edgarriba:
@prlz77 does #846 fix this ? otherwise we can merge it

prlz77:
@edgarriba this is complementary to #846, so it is still necessary.

PR #819, opened 2017-07-23
Integrate parameters into avepool layer kernels.

karandesai-96:


edgarriba:
@karandesai-96 any progress on that ?

karandesai-96:
@edgarriba was writing benchmarks for a while. Will look into this now.

test/test_convolutional_layer.h
```diff
@@ -209,27 +209,6 @@ TEST(convolutional, with_stride) {
 // test for AVX backends
 
 #ifdef CNN_USE_AVX
-TEST(convolutional, fprop_avx) {
```
karandesai-96: We didn't need this separately since a long time. I checked other layer tests.. Similar to what they have, `convolution.forward` test should test correct on few AVX jobs of Travis.

nyanp:
I met SEGV in network/manual_init test when I run the PR locally.

https://github.com/tiny-dnn/tiny-dnn/blob/master/test/test_network.h#L140-142

Maybe we need to modify above lines to adopt parameter refactoring.

test/test_convolutional_layer.h
```diff
@@ -210,98 +210,111 @@ TEST(convolutional, with_stride) {
...
 TEST(convolutional, bprop_avx_1x1out) {
   convolutional_layer l(5, 5, 5, 1, 2);
+  l.weight_init_f(parameter_init::xavier(1.0));
+  l.setup(false);
 
-  tensor_buf data(l), grad1(l);
-  tensor_buf grad2(grad1);
-
-  l.set_backend_type(tiny_dnn::core::backend_t::internal);
+  vec_t out_grads(1, 1);
```
nyanp: ```vec_t out_grads(2, 1);``` is correct (shape of output should be 1x1x2).

PR #823, opened 2017-07-23
- added new adamax optimizer.

blackccpie:
- added test for new optimizer.
- updated README.
- corrected adam optimizer implementation : seems like in the original paper the first loop iteration uses b1^1 and b2^1 rather than b1^2 and b2^2 like in the current adam implementation.


blackccpie:
sorry, I missed the clang-formatting step

PR #824, opened 2017-07-23
added new adamax optimizer

blackccpie:
- added test for new optimizer.
- updated README.
- corrected adam optimizer implementation : seems like in the original paper the first loop iteration uses b1^1 and b2^1 rather than b1^2 and b2^2 like in the current adam implementation.

### DIFF
--- a/tiny_dnn/optimizers/optimizer.h
+++ b/tiny_dnn/optimizers/optimizer.h
@@ -7,6 +7,7 @@
 */
 #pragma once
 
+#include <algorithm>
 #include <unordered_map>
 
 #include "tiny_dnn/util/util.h"
--- a/tiny_dnn/optimizers/optimizer.h
+++ b/tiny_dnn/optimizers/optimizer.h
@@ -154,10 +154,9 @@ struct adamax : public stateful_optimizer<2> {
 
     for_i(parallelize, static_cast<int>(W.size()), [&](int i) {
       mt[i] = b1 * mt[i] + (float_t(1) - b1) * dW[i];
-      vt[i] = std::max( b2 * vt[i], std::abs( dW[i] ) );
+      vt[i] = std::max(b2 * vt[i], std::abs(dW[i]));
 
-      W[i] -= alpha * ( mt[i] / (float_t(1) - b1_t)) /
-              ( vt[i] + eps);
+      W[i] -= alpha * (mt[i] / (float_t(1) - b1_t)) / (vt[i] + eps);
     });
 
     b1_t *= b1;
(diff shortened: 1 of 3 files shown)


tiny_dnn/optimizers/optimizer.h
```diff
@@ -135,6 +136,43 @@ struct adam : public stateful_optimizer<2> {
...
+  adamax()
+    : alpha(float_t(0.002)),
+      b1(float_t(0.9)),
+      b2(float_t(0.999)),
+      b1_t(b1),
+      eps(float_t(1e-8)) {}
+
+  void update(const vec_t &dW, vec_t &W, bool parallelize) {
+    vec_t &mt = get<0>(W);
+    vec_t &vt = get<1>(W);
```
edgarriba: use same notation as in the paper `v -> u`

tiny_dnn/optimizers/optimizer.h
```diff
@@ -135,6 +136,43 @@ struct adam : public stateful_optimizer<2> {
...
+
+    for_i(parallelize, static_cast<int>(W.size()), [&](int i) {
```
edgarriba: no need to cast here, just pass `size_t` to parallel for
`for_i(parallelize, W.size(), [&](size_t i)`

tiny_dnn/optimizers/optimizer.h
```diff
@@ -135,6 +136,43 @@ struct adam : public stateful_optimizer<2> {
...
+      mt[i] = b1 * mt[i] + (float_t(1) - b1) * dW[i];
+      vt[i] = std::max(b2 * vt[i], std::abs(dW[i]));
+
+      W[i] -= alpha * (mt[i] / (float_t(1) - b1_t)) / (vt[i] + eps);
```
edgarriba: not sure what trick you applied here, I propose
`W[i] -= alpha * mt[i] / ((1.0 - b1_t) * vt[i] + eps)`
or just
`W[i] -= (alpha / (1.0 - b1_t)) * (mt[i] / (vt[i] + eps))`

tiny_dnn/optimizers/optimizer.h
```diff
@@ -135,6 +136,43 @@ struct adam : public stateful_optimizer<2> {
...
+    });
+
+    b1_t *= b1;
```
edgarriba: add a comment here or something saying about Lp norm update

blackccpie:
I need a little help here, because I have no clue why the recurrent_cell.train test fails in one of the travis builds... this test uses optimizers, but no adam or adamax, that's why I'm puzzled

edgarriba:
v@blackccpie recurrent cell still have some numerical unstabilities. @prlz77 is working on that.  https://github.com/tiny-dnn/tiny-dnn/pull/818 should fix that

blackccpie:
@edgarriba alright, I saw this test fail on other pull requests indeed. Now they seem to have fully passed.

PR #834, opened 2017-07-26
Add basic timing benchmarks for fully connected op kernel.

kdexd:
This PR introduces timing benchmarks for fully connected op kernel. Benchmarks are evaluated with as well as without parallelization. 

Benchmarks include internal as well as avx backends, and are parametrized based on input sizes. Time of execution is found to linearly increase with increasing size of either input or output data keeping other constant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ cache:
   pip: true
   directories:
     - $HOME/.pip-cache/
+    # store old benchmarks binary here for comparison
+    - $HOME/.benchmarks/
 
 addons:
   apt:
@@ -48,6 +50,7 @@ env:
     - USE_TBB=ON
     - BUILD_TESTS=ON
     - BUILD_EXAMPLES=ON
+    - BUILD_BENCHMARKS=OFF
     - COVERALLS=ON
 
   matrix:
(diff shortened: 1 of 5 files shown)


benchmarks/bm_fully_connected.h
```diff
@@ -0,0 +1,225 @@
...
+
+#include "benchmark/benchmark.h"
+#include "tiny_dnn/tiny_dnn.h"
+
+namespace bm = benchmark;
+
+namespace tiny_dnn {
+namespace benchmarks {
+
+void bm_fully_connected_internal(bm::State& state) {
```
edgarriba: make a class, no duplicate code just for one flag

karandesai-96:
@edgarriba yes coming up.

karandesai-96:
@edgarriba not sure how this failure is happening in just one job. :confused:

.travis.yml
```diff
@@ -97,8 +106,24 @@ before_script:
     fi
 
 script:
-  - make -j2
-  - test/tiny_dnn_test
+  - if [ "$BUILD_BENCHMARKS" == "ON" ]; then
```
edgarriba: make a bash script for this to keep clean the travis file

karandesai-96: sure. makes sense.

PR #845, opened 2017-08-01
Fix build error on msvc /w openmp

nyanp:
This PR fix a build error with OpenMP&msvc.

Visual Studio supports older version of OpenMP (v2.0), but OpenMP 2.0 doesn't allow unsigned loop counter (see section 2.4.1 of [API Reference](http://www.openmp.org/wp-content/uploads/cspec20.pdf) (pdf) ). This PR avoid this problem by simply casting loop counter inside ```#ifdef CNN_USE_OMP```.


tiny_dnn/util/parallel_for.h
```diff
@@ -87,8 +87,9 @@ void parallel_for(size_t begin,
                   const Func &f,
                   size_t /*grainsize*/) {
   assert(end >= begin);
+  // unsigned index isn't allowed in OpenMP 2.0
 #pragma omp parallel for
-  for (size_t i = begin; i < end; ++i) f(blocked_range(i, i + 1));
+  for (int i = (int)begin; i < (int)end; ++i) f(blocked_range(i, i + 1));
```
edgarriba: @nyanp why needs to be int type ?

nyanp: @edgarriba It is a restriction from openmp specification. See my comment and this:
https://stackoverflow.com/questions/2820621/why-arent-unsigned-openmp-index-variables-allowed

PR #846, opened 2017-08-02
Blackbox gradient check, power layer fix, integration tests

prlz77:
As agreed with @edgarriba, layer tests should test the layers in isolation. Changes:

- New gradient checker that does not depend on network anymore, can check weightless activation layers.
- Old gradient checks on full networks moved to test_integration.h as well as those tests training whole networks.
- Fixed bug in power layer.
- Corrected SELU backpropagation.

**Please note that the new blackbox gradient check is performing multiple tests on random input/output positions, this is repetitive and should be done in parametrized tests, however i should prioritize RNNs right now as they are the main topic of the GSOC, the same for integration tests**


prlz77:
Upstream master tests are currently failing, and thus, this PR tests are failing after rebasing.

tiny_dnn/util/gradient_check.h
```diff
@@ -0,0 +1,137 @@
...
+  float_t prev_in                 = in_data[in_edge][0][in_pos];
+  in_data[in_edge][0][in_pos]     = prev_in - h;
+  layer.forward_propagation(in_data_, out_data_);
+  float_t out_1               = (*out_data_[out_edge])[0][out_pos];
+  in_data[in_edge][0][in_pos] = prev_in + h;
+  layer.forward_propagation(in_data_, out_data_);
+  float_t out_2               = (*out_data_[out_edge])[0][out_pos];
+  in_data[in_edge][0][in_pos] = prev_in;
+  layer.forward_propagation(in_data_, out_data_);
+  in_data[in_edge][0][in_pos] = prev_in;
```
edgarriba: why is this last needed ?

tiny_dnn/util/gradient_check.h
```diff
@@ -0,0 +1,137 @@
...
+  return (out_2 - out_1) / (2 * h);
+}
+
+/**
```
edgarriba: to follow the reference convention should we call this `compute_analytical_gradient` ?

tiny_dnn/util/gradient_check.h
```diff
@@ -0,0 +1,137 @@
...
+float_t relative_error(const float_t computed_gradient,
+                       const float_t numeric_gradient) {
+  float_t max =
+    std::max(std::abs(computed_gradient), std::abs(numeric_gradient));
+  if (max == 0) {
+    return 0;
+  } else {
+    return std::abs(computed_gradient - numeric_gradient) / max;
+  }
+}
```
edgarriba: `return (max == 0) ? std::abs(computed_gradient - numeric_gradient) / max : 0.0`;

tiny_dnn/util/gradient_check.h
```diff
@@ -0,0 +1,137 @@
...
+                          const size_t in_pos,
+                          std::vector<tensor_t> &out_data,
+                          std::vector<tensor_t> &out_grads,
+                          const size_t out_edge,
+                          const size_t out_pos) {
+  std::vector<tensor_t *> in_data_  = tensor2ptr(in_data);
+  std::vector<tensor_t> in_grads    = in_data;  // copy constructor
+  std::vector<tensor_t *> in_grads_ = tensor2ptr(in_grads);
+  std::vector<tensor_t *> out_data_ = tensor2ptr(out_data);
+  std::for_each(in_grads.begin(), in_grads.end(),
```
Randl: `for (auto &tensor : in_grads) fill_tensor(tensor, 0.0);`

tiny_dnn/util/gradient_check.h
Randl: some comments explaining might definitely be useful

prlz77: @edgarriba you are right, this last was used for debugging and I forgot to remove it

test/test_batch_norm_layer.h
```diff
@@ -165,6 +165,35 @@ TEST(batchnorm, gradient_check) {
   }
 }
 
+/*
```
prlz77: Because it does not pass the check, neither the original check in network.h

PR #848, opened 2017-08-03
xtensor integration for recurrent cell

Randl:


prlz77:
Just activate recurrent tests to see if everything is working properly.

test/test_tensor.h
```diff
@@ -88,16 +70,48 @@ TEST(tensor, view1) {
   EXPECT_EQ(tensor.host_at(0, 0, 0, 0), -1);
 }
 
+TEST(tensor, view2) {
+  Tensor<> tensor({3, 3, 3, 4}, 2);
+  auto t_view = tensor.subView(TensorRange(0, 2), TensorSingleIndex(0),
+                               TensorRange(0, 3), TensorAll());
+  EXPECT_EQ(t_view.shape()[0], size_t(2));
+  EXPECT_EQ(t_view.shape()[1], size_t(3));
+  EXPECT_EQ(t_view.shape()[2], size_t(4));
```
edgarriba: is this a xtensor convention or something you internally implemented? because IMHO, it should return a tensor of shape 2x1x3x4 instead of 2x3x4

tiny_dnn/core/framework/tensor.h
```diff
@@ -312,16 +297,50 @@ auto host_data() {
...
 
+  /**
+   * Fill tensor with another tensor repeated size times over first axis
+   * @tparam S storage of parameter tensor
+   * @param size
+   * @param tensor
+   * @return
+   */
+  template <typename S>
+  Tensor &fill(size_t size, Tensor<U, S> tensor) {
```
edgarriba: we should rename this to `Tensor::repeat(...)` it wasn't clear to me until I read the comments above

tiny_dnn/core/framework/tensor.h
```diff
@@ -312,16 +297,50 @@ auto host_data() {
...
   /**
    * Reshape tensor
    * @param shape new shape
    */
   void reshape(const std::vector<size_t> &shape) { storage_.reshape(shape); }
 
+  /**
+   * Reshapes to shape {}
+   */
+  void clear() { storage_.reshape({}); }
```
edgarriba: is this method also releasing memory ?

test/test_tensor.h
Randl: It is xtensor behavior, and I believe it's pretty intuitive: taking the index you expect to get rid of the dimension. I believe it's possible to use `Range` to keep axis of size 1.

tiny_dnn/core/framework/tensor.h
Randl: Since I haven't acquired memory manually, it should be released automatically too.

test/test_tensor.h
edgarriba: got it

PR #856, opened 2017-08-15
Guess batch size in weight update.

prlz77:
This will be needed for training recurrent neural networks with variable sequence size.


nyanp:
LGTM

PR #859, opened 2017-08-20
Restore previously broken serialization due to parameter integration.

karandesai-96:
Mostly this is chore work, removal of old redundant methods.


test/test_convolutional_layer.h
```diff
@@ -25,7 +25,7 @@ TEST(convolutional, setup_internal) {
   EXPECT_EQ(l.out_data_size(), 18u);             // size of output tensors
   EXPECT_EQ(l.fan_in_size(), 9u);                // num of incoming connections
   EXPECT_EQ(l.fan_out_size(), 18u);              // num of outgoing connections
-  EXPECT_EQ(l.weights_at()[0]->size(), 18u);     // size of weights
+  EXPECT_EQ(l.weights_at()[0]->size(), 36u);     // size of weights
```
nyanp: @karandesai-96 Is it correct? Size of shape should be 3x3x1x2=18.

test/test_deconvolutional_layer.h
```diff
@@ -33,7 +33,7 @@ TEST(deconvolutional, setup_tiny) {
   EXPECT_EQ(l.fan_in_size(), 9u);             // num of incoming connections
   EXPECT_EQ(l.fan_out_size(), 18u);           // num of outgoing connections
   EXPECT_EQ(l.parameters().size(), 2u);       // num of trainable parameters
-  EXPECT_EQ(l.weights_at()[0]->size(), 18u);  // size of weight parameter
+  EXPECT_EQ(l.weights_at()[0]->size(), 36u);  // size of weight parameter
```
nyanp: same as convolutional_layer

test/test_network.h
```diff
@@ -137,17 +137,17 @@ TEST(network, manual_init) {
...
+  EXPECT_EQ(c1_w.size(), 9u);
+  EXPECT_EQ(c1_b.size(), 1u);
+  EXPECT_EQ(f1_w.size(), 2u);
 
-  *c1_w = {0, 1, 2, 3, 4, 5, 6, 7, 8};
-  *c1_b = {1};
-  *f1_w = {1, 2};
+  c1_w.set_data(Tensor<float_t>{vec_t{0, 1, 2, 3, 4, 5, 6, 7, 8}});
+  c1_b.set_data(Tensor<float_t>{vec_t{1}});
+  f1_w.set_data(Tensor<float_t>{vec_t{1, 2}});
```
nyanp: Consider adding initializer-list assignment operator to ```Parameter``` to simplify these codes.

test/test_serialization.h
```diff
@@ -46,8 +46,11 @@ TEST(serialization, serialize_avepool) {
                     "height": 10,
                     "depth": 3
                 },
-                "pool_size_x": 2,
-                "pool_size_y": 2,
+                "window": {
+                    "width": 2,
+                    "height": 2,
+                    "depth": 3
+                },
```
nyanp: We can keep it as is. Do we need to break backward compatibility?

test/test_network.h
Randl: Also, you don't need to write `float_t`, `Tensor<>` is enough and much shorter

karandesai-96:
Coverage restored! (+10%) :smile:

test/test_network.h
karandesai-96: @nyanp you're right!
@Randl yes, thanks!

test/test_serialization.h
karandesai-96: Changed under the hood but user facing API is kept backward compatible now.

PR #864, opened 2017-08-24
Network Tensor integration

Randl:


edgarriba:
@Randl what's the status for this ?

PR #865, opened 2017-08-25
Simplify integration tests

prlz77:
Increases numerical stability in integration tests (better init, remove gradient check of deep networks, i.e. rnns.)


edgarriba:
@prlz77 any idea why coveralls drops -.5% ?

prlz77:
@edgarriba I think because I erased two tests

PR #867, opened 2017-08-26
Restore network tests and caffe converter tests after Parameter integration.

karandesai-96:
This PR restores network, target_cost and caffe_converter tests which were commented out during parameter integration.

### DIFF
--- a/test/test_network.h
+++ b/test/test_network.h
@@ -233,7 +233,7 @@ TEST(network, train_predict) {
 
   net << fully_connected_layer(2, 10) << tanh_layer()
       << fully_connected_layer(10, 2) << tanh_layer();
-  net.init_parameters();
+
   net.train<mse>(optimizer, data, label, 10, 10);
 
   std::vector<tensor_t> parallel_input(tnum);
@@ -287,7 +287,7 @@ TEST(network, train_predict_different_batches) {
 
     net << fully_connected_layer(2, 10) << tanh_layer()
         << fully_connected_layer(10, 2) << tanh_layer();
-    net.init_parameters();
+
     net.train<mse>(optimizer, data, label, batch_sz, 10);
 
     std::vector<tensor_t> parallel_input(data_size);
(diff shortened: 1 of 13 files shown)


test/test_network.h
```diff
@@ -233,7 +233,7 @@ TEST(network, train_predict) {
 
   net << fully_connected_layer(2, 10) << tanh_layer()
       << fully_connected_layer(10, 2) << tanh_layer();
-
+  net.init_parameters();
```
edgarriba: @karandesai-96 so now is it mandatory to explicitly initialize the parameters? It should be automatic unless the user wants an specific initialization or import from whatever resource.

karandesai-96: @edgarriba Well I can call init_parameters of tail layer in `<<` operator. How about that ?

Randl: @karandesai-96 sounds good

test/test_network.h
```diff
@@ -326,12 +326,12 @@ TEST(network, set_netphase) {
 TEST(network, test) {
   network<sequential> net;
   fully_connected_layer fc(30, 1);
-  int data_num = 300;
+  fc.weight_init(parameter_init::constant(1.0));
+  fc.init_parameters();
```
prlz77: Why did you remove this from previous tests but keep it here?

karandesai-96: Layer API has this behaviour since always. `weight_init()` and `bias_init()` set the initialization strategies but not perform initialization. So we need to call `layer.setup(true)` or `layer.init_parameters()` to initialize them.

But for network, we need not require to explicitly call `network.init_parameters()` as initialization happens in layers through `<<` operator. Although we explicitly call `init_parameters` when we set different initialization strategies. (Default ones are `xavier` for weights and `constant(0)` for biases.)

karandesai-96: Previous `layer.setup(true)` was called but calling that during `<<` operator network makes sense as there is data allocation to incoming and outgoing edges of layer. But If I initialize a single layer object, I thought calling `init_parameters()` instead of `setup()` shows more clarity in what we actually intended to do.

PR #868, opened 2017-08-29
Add support to load Parameter data from HDF files.

kdexd:
### Overview

* This PR introduces support to support Parameter data directly from a HDF file.
* As an initial support, HDF files containing model weights, produced by **Keras** on **Tensorflow backend** are supported.
* A new macro has been introduced - `CNN_USE_HDF`. **hdf5** C++ interface is a soft dependency for the purpose.
* Loading of only sequential models is supported yet. Graph models support shall be introduced later.
* Saving facility and more file formats will be supported soon. This is the first step.

### Structure of HDF files required

* Any file produced by [`model.save_weights`](https://keras.io/models/about-keras-models/) keeping tensorflow backend can be loaded into tiny-dnn. File hierarchy is as described:

(**Note**: Just information, this file structure need not be created manually as all of it is handled by keras).
```text
root (filename.h5)   //H5::File
  -- conv_1          // H5::Group
    -- conv_1
      -- kernel:0    // H5::DataSet
      -- bias:0
  -- maxpool_1
  -- conv_2
    -- conv_2
      -- kernel:0
      -- bias:0
  -- maxpool_2
  -- dense_1
    -- dense_1
      -- kernel:0
      -- bias:0
  -- dense_2
    -- dense_2
      -- kernel:0
      -- bias:0 
```
* Each layer shall be a group and shall have one `H5::Attribute` named `weight_names` associated with it.
  * For example `conv_1`'s attribute shall have value `[conv_1/kernel:0, conv_1/bias:0]`
* `H5::File` shall have one `H5::Attribute` named `layer_names` containing names of `H5::Group`s in same order as they occur in Sequential network.
* All `H5::Dataset`s for weights shall have 4D tensors of shape `(width x height x in_channels x out_channels)` for convolution type layers, while `(in_dim x out_dim)` for FC type layers.

### API introduced so far

* `Parameter::load(const std::string &file_path, const std::string &parameter_name)`
* `layer::load(const std::string &file_path, const std::string &layer_name)` 
* `layer::load` iteratively calls `Parameter::load` for each of its parameters.

### Future

Keras was chosen because its HDF files are highly structure and it is widely used, with 4 backends (Tensorflow / Theano / MxNet / CNTK). Although more support would be introduced soon.

* Add `network<sequential>::load` of similar pattern.
* Add saving support just like loading support.
* Extend to HDF files of Keras with different backends.
* Distant future: Support Tensorflow and Pytorch model files too.


karandesai-96:
Probably I should combined `CNN_NO_SERIALIZATION` and `CNN_USE_HDF`.

karandesai-96:
@edgarriba before #869 I think we should discuss things here. This is the first place where I introduce HDF5 stuff. We can limit things here itself and they will have a cascading effect in further PRs.

tiny_dnn/layers/layer.h
```diff
@@ -454,6 +458,52 @@ class layer : public node {
     }
   }
 
+#ifdef CNN_USE_HDF
```
edgarriba: provide this as an example/tutorial. Map data to `std::vector<Tensor<>>` and then overload operators in network, later and parameter to make a tiny API in order to load the data

karandesai-96: This sounds good. I'll do it here.

edgarriba: We should take this project as a reference and try to load that networks in a similar way
https://github.com/szagoruyko/functional-zoo

See that the parameters are fetched to the layers independently to the serializer. It uses numpy array as an intermediate representation to map to `torch.Tensor` , in our case will be directly `std::vector<float>` to `dnn::Tensor`

karandesai-96: So shall we have a `hdf_serializer` class in a file under `tiny_dnn/util` which does this tying of data?

karandesai-96: And I think if we wish to do provide framework specific support for special cases, we can define other serializers inheriting this one. And the API can be much similar like `optimizers` and `parameter_init`.

Just as `parameter.initialize` takes an argument as `parameter_init::function`, we can do `parameter.load(hdf_serializers::serializer &s, args... )` and it depends on use which serializer to use (in his own program).

karandesai-96: We shall have support but internally we will be independent of the file structure and version hassles. What say ?
/cc: @bhack @nyanp @Randl @prlz77 you are welcome to join the discussion :smile:

edgarriba: That approach sounds better but I guess it might take some time to refactor the serializer. I'd go first to the simplest case and load data assuming that parsing is done by the user.

nyanp: @karandesai-96 
I think there are 4 different abstraction levels for importer API in this discussion:

1. Tensor API (Just loading tensor from file, as @edgarriba said)
1. Layer API (#868)
1. Network API (#869)
1. Model API (#870)

Supporting 1 and 4 are good enough for me (providing both flexibility and convenience). But if we provides 2 and 3 as public API, we will need to make a lot of effort to support various user-defined models. In addition to that, we need to write additional code if we add a new layer into tiny-dnn.

Constructing Model API over primitive tensor API seems good for maintainability point of view.

bhack: @nyanp @edgarriba From the last GoogleBrain Ama [seems](https://www.reddit.com/r/MachineLearning/comments/6z51xb/we_are_the_google_brain_team_wed_love_to_answer/#dmtlgdt) that TF team is open on [onnx](https://github.com/onnx/onnx). What do you think?

nyanp: @bhack /cc @edgarriba   @karandesai-96
ONNX is very promising format if TF is open on it. This is one of the reason why I don't want to provide layer/network API with Keras's native format. After supporting popular model with minimum effort, I want to support ONNX (it seems good path for us to survive).

edgarriba: agree !

PR #869, opened 2017-08-29
Add method in network to load pretrained Sequential Models from HDF.

kdexd:
This PR is an extension to #868 works in a similar way as `layer::load` and `Parameter::load`. Note that topologically loading network calls methods of classes **network -> layer -> parameter** in a DFS style. 

Although all of these can be called standalone, example usage can be currently seen in **test_serialization_hdf.h**.

Also, this PR combines macros `CNN_USE_HDF` and `CNN_NO_SERIALIZATION` as decided in #868 .


tiny_dnn/layers/layer.h
```diff
@@ -454,6 +458,33 @@ class layer : public node {
     }
   }
 
+  void load(const std::string &filename, const std::string &layer_name) {
```
edgarriba: I won't include this as part of the API, instead I think that we should limit to provide methods to load just from tensors. I see this as an example how the user can use any serialized file (hdf5 in this case) to fit the layers. I posted some days ago in gitter one possible API:
```cpp
// the user populates our tensor wrapper
std::vector<Tensor<>> data = load(whatever_format_file)

net[i][0] = data[i]  // we assume agnostic data format
net[i][1] = data[i+1]  
```
Another drawback of having hdf5 as part of the core API is that we are tied to it's interface which may vary in different versions and make the whole thing hard to maintain.

karandesai-96: Ah I see. This makes a lot of sense to me. 
One more question. We are allowing users to use `net.load` right ? Say there's a huge network of 50 layers and the HDF5 file is having 100 tensors in a hierarchy. There should be a convenient one line call.

edgarriba: it should be the user who loads from that huge file. Networks are getting more sophisticated and we can never predict how is serialized.

karandesai-96: @edgarriba True, we should have a generalized format, but I am considering the mass of users using tiny-dnn.

I am assuming people to use popular Frameworks like Keras, Tensorflow and Pytorch who currently have good GPU capability to train their networks. They will come to tiny-dnn to load them and deploy them on low compute power devices, say any iOT device.

These frameworks clearly define a way of serializing their network graphs. We currently have an enum named `content_type`. I was envisioning it to be:

```c++
enum content_type { binary, json, hdf, keras_hdf, tf_hdf, t7, mxnet_hdf }  // any any other supported frameworks
```

edgarriba: This frameworks are quite recent and may evolve which will make it hard to maintain in the long term. Caffe converter suffered from this.

karandesai-96: But we can stick to a version right ? Keras HDF has version labelled in their file as an attribute. I am not contradicting your point here, your API should be included, but there should be at least some convenience wrappers to save a lot of lines of code on user end.

Before I read this ^ comemnt, I was typing out a new comment - let me post it here. Not sure how valid it is now...

karandesai-96: Along with a generic API, we should have convenience methods to load well-structured HDFs defined in these very popular frameworks because we intend to bring the audience to us.

I just started with `keras_hdf` first. Here is the net structure:
```c++
net << fc(2, 4) << relu << fc(4, 1) << softmax();
```

Here are a few contents from **hdfview** of pretrained file:
1. While focusing on filename (**xor.h5**)
```text
    Group size = 2  // lets us know how many layers there are serialized
    Number of attributes = 3 
        backend = tensorflow  // lets us know whether NCHW or NHWC
        keras_version = 2.0.0   // lets us know if in future structure changes, what to look for
        layer_names = dense_1,dense_2  // let us know names of H5::Groups under
```

2. While focusing on Group (layer):
```text
dense_1 (1080, 2)
    Group size = 1
    Number of attributes = 1
        weight_names = dense_1/kernel:0,dense_1/bias:0  // lets us know names of tensors inside it
```

3. While focusing on Dataset (parameter):
```text
kernel:0 (6384, 2)
    32-bit floating-point,    2 x 4  // lets us know the shape
```

Hence my point is, we can make out of the structure of file pretty much from various attributes, and can generalize over all network architectures, **provided we have an HDF coming from Keras**.

Similar structures are defined for other frameworks. But at last I still agree with your point, there should also be a way to load a tensor from any type of file in a generic sense.

PR #870, opened 2017-08-29
Add class for VGG16 and VGG19 models.

kdexd:
This PR adds an implementation for VGG16 and VGG19 model. This class can be found as `models::vgg16` / `models::vgg19`. We can choose to include fully connected layers or prefer global average pooling like Network-in-Network architecture.

Once the HDF loading API is tested and merged, it shall be used to load pretrained model from official Keras weights file.

Initialization strategy if Imagenet weights are not loaded, is kept same as that mentioned in the paper.

### References:

1. Keras HDF Files: https://github.com/fchollet/deep-learning-models/releases/tag/v0.1
2. VGG paper: https://arxiv.org/pdf/1409.1556.pdf

### DIFF
--- a/tiny_dnn/models/vgg16.h
+++ b/tiny_dnn/models/vgg16.h
@@ -8,17 +8,29 @@
 #pragma once
 
 #include <string>
+#include "tiny_dnn/tiny_dnn.h"
 
+#include "tiny_dnn/util/nn_error.h"
+#include "tiny_dnn/util/parameter_init.h"
+
+namespace tiny_dnn {
 namespace models {
 
+/**
+ * VGG16 architecture for tiny-dnn model zoo.
+ *
+ * Reference:
+ * Very Deep Convolutional Networks for Large-Scale Image Recognition
+ * - (https://arxiv.org/abs/1409.1556)
+ */
 class vgg16 : public tiny_dnn::network<tiny_dnn::sequential> {
  public:
   explicit vgg16(const std::string &name = "vgg16", bool include_top = true)
     : tiny_dnn::network<tiny_dnn::sequential>(name) {
-    using conv     = tiny_dnn::layers::conv;
-    using relu     = tiny_dnn::activation::relu;
-    using max_pool = tiny_dnn::layers::max_pool;
-    using dense    = tiny_dnn::layers::dense;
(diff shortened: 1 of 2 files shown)


karandesai-96:
**TODO**: We can base it off from master after `feat/xtensor_integration` is tested properly and merged.

tiny_dnn/models/vgg16.h
```diff
@@ -0,0 +1,87 @@
...
+    using conv     = convolutional_layer;
+    using relu     = relu_layer;
+    using max_pool = max_pooling_layer;
+    using dense    = fully_connected_layer;
+
+    // Block 1
+    *this << conv(224, 224, 3, 3, 3, 64, padding::same) << relu();
+    *this << conv(224, 224, 3, 3, 64, 64, padding::same) << relu();
+    *this << max_pool(224, 224, 64, 2) << relu();
+
```
edgarriba: We should be able to do this
```cpp
void create_block(tiny_dnn::network<tiny_dnn::sequential> &net, int in_size, int in_channels, int out_channels) {
  net << conv(in_size, in_size, 3, 3, in_channels, out_channels, padding::same) << relu();
  net << conv(in_size, in_size, 3, 3, out_channels, out_channels, padding::same) << relu();
  net << max_pool(in_size, in_size, out_channels, 2) << relu();
}
```
```cpp
int in_size = 224;
int in_channels = 3;
int out_channels = 64;
for (int i = 0; i < num_blocks; ++i) {
  create_block(*this, in_size, in_channels, out_channels);
  in_channels = out_channels;
  out_channels *= 2;
  in_size /= 2;
} 
```

karandesai-96: Okay so we intend to allow a little modifications over VGG as well ? Sounds good. 

I was planning this thing for Inception and Resnet architecture since Inception and Residual modular blocks are quite popular. Well let's extend this to VGG as well.

Adding :+1:

edgarriba: In this way, I guess you could cover different vggs in a single class and define them programmatically

karandesai-96: Great!

PR #871, opened 2017-08-31
Fix minor mistakes in numeric and analytic gradient checks.

karandesai-96:
Earlier, target gradient was not set as 1 because the value was not referenced, but instead a copy of actual. Hence both numeric and analytic gradients remained zero for many tests, passing them all.

### DIFF
--- a/tiny_dnn/core/kernels/fully_connected_op_internal.h
+++ b/tiny_dnn/core/kernels/fully_connected_op_internal.h
@@ -33,11 +33,11 @@ inline void fully_connected_op_internal(const Tensor<float_t, S1> &in_data,
       out_data.host_at(sample, i) = float_t{0};
       for (size_t c = 0; c < in_size; c++) {
         out_data.host_at(sample, i) +=
-          weights.host_at(0, c * out_size + i) * in_data.host_at(sample, c);
+          weights.host_at(c * out_size + i) * in_data.host_at(sample, c);
       }
 
       if (bias.size() >= out_size) {
-        out_data.host_at(sample, i) += bias.host_at(0, i);
+        out_data.host_at(sample, i) += bias.host_at(i);
       }
     }
   });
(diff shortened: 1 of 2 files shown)


karandesai-96:
Example, `fully_connected.gradient_check`:

**ngrad**: -0.20733282
**cgrad**: -0.207230061

Earlier both were zero.

karandesai-96:
No clue why those two jobs fail, I cannot reprodue that error.

prlz77:
Is this also happening in master?

prlz77:
No clue why those two jobs fail, I cannot reproduce that error.

> I have checked it and no clue as well, it is a function that does not use to fail. Try to repoen this PR to see if errors change

karandesai-96:
@prlz77 Yes, for fully conencted layer gradient check, both `ngrad` and `cgrad` are zero on master.

karandesai-96:
@prlz77 I just restarted the two failing jobs and they passed!! These days some of our tests are getting really very hairy.

prlz77:
@karandesai-96 I know... sometimes travis fails even at compiling, then you restart and it works...

By the way, I don't understand why this is not working in master https://github.com/tiny-dnn/tiny-dnn/blob/master/tiny_dnn/util/gradient_check.h#L113

prlz77:
hmm it seems I fixed it in my PR branch hahaha

https://github.com/prlz77/tiny-dnn/blob/c7f64596e922aa7f9655c81beb35cdcf250d8813/tiny_dnn/util/gradient_check.h

karandesai-96:
Haha, out of two jobs one passed and another failed but the test which failed it earlier passed now and another test failed it XD XD

prlz77:
What did you do?

karandesai-96:
@prlz77 restarted the job that's it, and I feel guilty for it xD

PR #873, opened 2017-09-01
Visual C++ (C++14) does not convert double literal to float

theSundayProgrammer:
The test code fails to compile because C++ "1.0f" when specifying a float rather than "1.0" which is double


edgarriba:
@theSundayProgrammer thanks for the contribtion. Just one thing, what about when we compile with `USE_DOUBLE` ?

beru:
I've opened similar PR #898 since I didn't notice this PR takes care of the same problem.

@edgarriba I confirmed `tiny_dnn_test` passes even with `float` literals and `USE_DOUBLE` defined.
Though I'm not sure if double precision is necessary or not necessary for the tests.

@theSundayProgrammer  You need to apply `clang-format` command to pass Travis-CI build.

edgarriba:
@theSundayProgrammer please apply clang format

beru:
@edgarriba I'd like to close this PR and reopen #898 since there has been no response.

PR #878, opened 2017-09-22
Fix Clazy findings

kecsap:
This is the continuation of my previous PR: https://github.com/tiny-dnn/tiny-dnn/pull/518

1. Add missing refs to some loops.
2. const keyword has no effect on return values.

### DIFF
--- a/tiny_dnn/core/framework/tensor.h
+++ b/tiny_dnn/core/framework/tensor.h
@@ -104,19 +104,19 @@ class Tensor {
    *
    * @return the tensor shape
    */
-  const auto shape() const { return storage_.shape(); }
+  auto shape() const { return storage_.shape(); }
 
   /**
    *
    * @return Tensor's number of dimensions
    */
-  const auto dim() const { return storage_.dimension(); }
+  auto dim() const { return storage_.dimension(); }
 
   /**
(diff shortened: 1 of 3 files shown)


tiny_dnn/core/framework/tensor.h
```diff
@@ -144,11 +144,11 @@ class Tensor {
    */
   auto host_begin() { return storage_.xbegin(); }
 
-  const auto host_begin() const { return storage_.cxbegin(); }
+  auto host_begin() const { return storage_.cxbegin(); }
```
edgarriba: @Randl what's the reason to keep constness here ?
@kecsap could you verify through test that this change doesn't affect the expected behavior ?

beru: I guess the constness can be deduced in these cases.

Out of curiosity, I checked C++ spec draft (around 10.1.7.4 The auto specifier) but couldn't locate the exact guideline.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf

And I wonder about the test edgarriba mentioned. I hope we can let compilers manage the task.
I mean... I don't think additional test is needed for this change.

PR #887, opened 2017-10-07
Fix NNPack kernel syntax and casting issues

noio:
In addition to removing the parens, I got an issue about no implicit casting allowed from float to `size_t`, so Xcode suggested these explicit casts. All runs fine after that. 

Additionally, the `NNPackInitializer` is in the `core` namespace, so I also fixed the complaints about that.


edgarriba:
@noio please format the code using clang. Follow the instructions from README

noio:
done.

PR #889, opened 2017-10-15
 update readme to list all available activation functions 

lakshayg:


karandesai-96:
Looks good, will merge once everything passes - we got two approvals. Thanks @lakshayg !

PR #896, opened 2017-10-20
Parameterization of elu layer

srk97:
### Added the hyperparameter `alpha` to elu activation unit.

The modified forward propagation is:

elu

### DIFF
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -33,6 +33,7 @@ using namespace tiny_dnn::activation;
 #include "test_models.h"
 #include "test_node.h"
 #include "test_nodes.h"
+#include "test_optimizers.h"
 #include "test_power_layer.h"
 #include "test_quantization.h"
 #include "test_quantized_convolutional_layer.h"
@@ -41,7 +42,6 @@ using namespace tiny_dnn::activation;
 #include "test_slice_layer.h"
 #include "test_target_cost.h"
 #include "test_tensor.h"
-#include "test_optimizers.h"
 
 #ifndef CNN_NO_SERIALIZATION
 #include "test_serialization.h"
--- a/test/test_serialization.h
+++ b/test/test_serialization.h
@@ -18,7 +18,7 @@
 namespace tiny_dnn {
 
 inline void check_sequential_network_model_serialization(
-    const network<sequential>& net) {
+  const network<sequential>& net) {
   auto path = unique_path();
   network<sequential> net2;
   net.save(path, content_type::model);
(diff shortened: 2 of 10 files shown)


srk97:
@edgarriba @Randl @karandesai-96 All checks have passed. Could you review it?

test/test.cpp
```diff
@@ -41,7 +42,6 @@ using namespace tiny_dnn::activation;
 #include "test_slice_layer.h"
 #include "test_target_cost.h"
 #include "test_tensor.h"
```
beru: They are in alphabetical order now. No one knows what comes next.  :scream:

tiny_dnn/activations/elu_layer.h
```diff
@@ -33,14 +50,15 @@ class elu_layer : public activation_layer {
                            const vec_t &dy) override {
     for (size_t j = 0; j < x.size(); j++) {
       // dx = dy * (gradient of elu)
-      dx[j] = dy[j] * (y[j] > float_t(0) ? float_t(1) : (float_t(1) + y[j]));
+      dx[j] = dy[j] * (y[j] > float_t(0) ? float_t(1) : (alpha_ + y[j]));
```
beru: It seems correct according to below references.

https://arxiv.org/pdf/1511.07289.pdf
https://en.wikipedia.org/wiki/Activation_function#Comparison_of_activation_functions

Chainer's `ELU::backward_cpu` looks different though.
https://github.com/chainer/chainer/blob/master/chainer/functions/activation/elu.py
It multiplies with alpha. I don't know why...

Caffe's `ELULayer::Backward_cpu` uses addition, so it looks the same.
https://github.com/BVLC/caffe/blob/master/src/caffe/layers/elu_layer.cpp

TensorFlow's `EluGradOp::Compile` does not use alpha.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tf2xla/kernels/elu_op.cc#L49

After checking this gradient code, I wonder if `selu_layer::backward_activation` is OK.

srk97: Why would they change the backward pass and keep forward pass the same? :confused:

srk97: I referred to the paper

edgarriba:
@srk97 something went wrong with style format

test/test_serialization.h
```diff
@@ -1006,7 +1008,7 @@ TEST(serialization, graph_model_and_weights) {
   fully_connected_layer f2(2, 2);
   softmax_layer a2(2);
   fully_connected_layer f3(2, 2);
-  elu_layer a3(2);
+  elu_layer a3((size_t)2);
```
edgarriba: for convention do:
`elu_layer a3(size_t(2))` or `elu_layer a3(static_cast<size_t>(2))`

PR #898, opened 2017-10-22
Add workarounds for Compile Error C2397

beru:
Debug configuration build of `tiny_dnn_test` project fails with lots C2397 errors.
Tested with Microsoft Visual Studio Community 2017 Version 15.4.1.

This PR adds workarounds for it.
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2397
There are several ways to circumvent the problem.

1. Changing compiler option
`Configuration Properties => C/C++ => Code Generation => Floating Point Model`
With default `Precise (/fp:precise)`, the compiler gives the error.
But after changing the option to `Fast (/fp:fast)`, the error disappeared. 
Release configuration build uses `Fast (/fp:fast)` so I guess that's why this problem doesn't occur with Release config.

2. Adding `f` suffixes to floating point literals for test code.
This reduces numerical precision but I'm not sure if any test code requires `double` floating point literals. I confirmed `tiny_dnn_test` Release configuration build runs fine and passed tests with or without `CNN_USE_DOUBLE` preprocessor definition.

3. Adding cast code
I chose this aesthetically unpleasant way and used C-style casts since writing a lot of `static_cast<float_t>(literal)` is ugly and painful.

### DIFF
--- a/test/test_batch_norm_layer.h
+++ b/test/test_batch_norm_layer.h
@@ -220,13 +220,13 @@ TEST(batchnorm, forward) {
     /* y = (x - mean) ./ sqrt(variance + eps) */
     tensor_t expect = {
         {
-            0.0,    0.0,    0.0,   0.0,    // ch-0 of data#0
-           -1.225,  0.408,  0.0,   1.225,  // ch-1 of data#0
-           -0.573, -0.879, -0.573, 0.038,  // ch-2 of data#0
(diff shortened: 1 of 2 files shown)


edgarriba:
@beru given that option 2 works, I'd go for it since gives a clean solution

beru:
@edgarriba I close this issue since the plan 2 was already taken in #873

beru:
The path for supporting VS2017 was closed with #947 so I'll do the same to this PR.

PR #903, opened 2017-11-11
Consistent use of preprocessor macros

dirkgr:
On non-MSVC platforms, `_MSC_VER` is not defined, which results in a warning. When you treat warnings as errors, it doesn't compile at all.

`#if` vs. `#ifdef` is inconsistent in this file anyways, so I changed them all to `#ifdef`.


beru:
I'm not against this PR but it looks like VS2015 supports C99 standard `snprintf` function.
https://msdn.microsoft.com/en-us/library/2ts7cx93.aspx

The support for VS2013 has dropped (due to some reasons),
so I suspect we can remove those pesky preprocessor directives.

PR #931, opened 2018-02-11
homebrew/science deprecated

beru:
Travis CI osx jobs fail so I'm trying to fix it.


beru:
@bhack Thank you for the approval!

PR #935, opened 2018-02-18
Cblas op example and bug-fix

jimbozhang:
1. Support USE_CBLAS in CMakefile.txt;
2. Add an example to show how to use CBLAS backend;
3. Bug fix.


tiny_dnn/core/kernels/fully_connected_op_cblas.h
```diff
@@ -25,14 +25,17 @@ inline void fully_connected_op_cblas(const tensor_t &in_data,
...
+  size_t in_size        = params.in_size_;
+  float_t alpha         = 1;
+  float_t beta          = 1;
+  const float_t *input  = in_data[0].data();
+  const float_t *weight = W.data();
+  float_t *output       = out_data[0].data();
+  if (bias.empty())
+    memset(output, 0, sizeof(float_t) * out_size);
+  else
+    memcpy(output, bias.data(), sizeof(float_t) * out_size);
```
edgarriba: not sure about this. Maybe @beru has another opinion about that.

beru: I guess setting beta to zero instead of zeroing output when `!params.has_bias_` might be more efficient. But to be honest, I have no idea about the BLAS.

jimbozhang: @beru is right. `!params.has_bias_` is better.
I'm doing BLAS support for more ops (e.g. RNN), I'll use `params.has_bias_` in my next commit.

edgarriba: @jimbozhang as for blas (and potentially xsim) it's worth exploring `xtensor-blas` since it has already lot of functionalities implemented. 
https://github.com/QuantStack/xtensor-blas

However, to @Randl still have to finish xtensor integration in https://github.com/tiny-dnn/tiny-dnn/pull/864

PR #917, opened 2017-12-17
xtensor update 0.11.3

blackccpie:
In order to fix #854, this PR updates tiny-dnn's embedded tensor version to 011.3 (latest version not depending on xtl).

PR #939, opened 2018-02-25
update xtensor

beru:
I've noticed that there are build errors with Visual Studio 2017 after merging #917.
So I updated xtensor library to latest version.

xtensor : e3138b1302fd56d9cc63cfc8e6cba9f0057464d3
xtl : c71be890857e87943b41adc48e7f7774d4ce8c48


beru:
@edgarriba Can you write down the instruction then?

beru:
@edgarriba Can we remove `tiny_dnn/core/framework/tensor.h` and related thingies for now from master branch? They aren't actually used so having them just hinders build process. And it seems to me #864 is not active anymore. Has it been abandoned?

edgarriba:
@beru go ahead, in case that @Randl or someone else wants to finish xtensor integration we'll include it again

bhack:
I think it is ok to remove. Want we target to a specific xtensor release tag?

beru:
It seems like recent xtensor does not support old compilers.

Issue #921, opened 2018-01-14
Gsoc 2018


bhack:
After GSCOC 2017 initial support in tinydnn we want to disambiguate a little bit the overlap points between xtensor/xsimd and nnvm/tvm for GSOC 2018 proposals.

/cc @edgarriba @nyanp @Randl @beru @prlz77 

Any upstream feedback it is appreciated /cc @tqchen @SylvainCorlay @JohanMabille

bhack:
Two main topics are:

- [ ] Graph rappresentation and optimizzation pass (NNVM)
- [ ] xtensor/xsimd ops (subset?) overlap with [TOPI](http://docs.tvmlang.org/doxygen/namespacetopi.html)

tqchen:
[some recent result](http://www.tvmlang.org/2018/01/16/opt-mali-gpu.html) that might interests you for mobile gpu deployment

bhack:
@tqchen So I suppose that we to produce an NVVM frontend for [our graph](https://github.com/tiny-dnn/tiny-dnn/blob/master/docs/how_tos/How-Tos.md#graph-model).
What are the differences between Onnx Graph IR and NVVM Graph IR?

bhack:
Also If I understand correctly we could compile the graph only with the python interface (so not with c++) but then we could control the deployment/execution with c++. Is this right?

bhack:
I.e. for reference http://nnvm.tvmlang.org/tutorials/from_onnx.html

tqchen:
C++ interfacing is a WIP, but yes for now python interface is better supported. deployment can be done in any language(c++, python, javascript, java)

bhack:
@tqchen Do you see any difference between ONNX, that has basically a models list "contract", and NNVM? 
So is ONNX really proof only on the [model list](https://github.com/onnx/models) that it has mandatory in the "standard"? And what is the case on NNVM?

tqchen:
Again I don't think one computational graph spec would be that much different from one way or another, we need some common ground, ONNX is a good example of doing that. 

NNVM tries to think what other things can be done once we have the graph, be it onnx or tiny-dnn, can we transform them to optimize for different devices, can we precompute certain part of parameters like the pre-packing optimizations.

bhack:
Ok thanks

xiaoerlaigeid:
Hi! I am new here. I want to particpate gsoc this project. could you give me some advice on this ? Thanks a lot

[...]

PR #947, opened 2018-03-10
remove xtensor

beru:
This PR removes xtensor related code from the library.
#939 is related.


edgarriba:
@beru I'm okay with it, but what's the alternative then ?

edgarriba:
@beru @bhack maybe it's enough with a dlpack like representation ?
https://github.com/dmlc/dlpack

beru:
> what's the alternative then ?

@edgarriba I'd like to find very easy approach. I grepped with `tensor_t` keyword and 1110 lines were printed. So if we choose thorough modification approach, it may take years.

bhack:
Probably could be to adopt Dlpack tensor and use TVM ops. So that we could remove many internal implementations.

bhack:
A more drastic approach is to just design our NNVM network graph and let to use TVM ops. See https://github.com/dmlc/nnvm/blob/master/docs/how_to/deploy.md

bhack:
If you want take a look to this intro https://aws.amazon.com/it/blogs/machine-learning/introducing-nnvm-compiler-a-new-open-end-to-end-compiler-for-ai-frameworks/

beru:
@bhack If I understand correctly, relying on NNVM and TVM means this library will not be header only nor dependency-free. Is that correct?

bhack:
We can see NNVM basically as a NN compiler so tinydnn could be a c++11 header only frontend. If you see here NNVM could consume and deploy already from many frontends https://github.com/dmlc/nnvm/tree/master/tutorials

bhack:
So with this drastic option there could be something like from_tinydnn

beru:
@bhack I briefly checked several `from_*` scripts at https://github.com/dmlc/nnvm/tree/master/tutorials.
Basically what the scripts do is loading famous library/framework's model files then compilation and execution follow. So compatibility through data-formats is there. But is that the thing we're talking about?

NNVM already has Python and C++ bindings so I imagine modifying this library is like rebuilding a trailer truck using small electric car. Let's admit we ain't got no :rugby_football:  :rugby_football: of :metal: to overcome such technical difficulties and debts, shall we?

bhack:
You can check two comments at https://github.com/tiny-dnn/tiny-dnn/issues/921#issuecomment-358707547. We actually don't have graph optimizzation, gpu ops, and many optimized cpu ops on different hardware. So I don't think that we have bandwidth to implement all this things in tinydnn.

beru:
@bhack @edgarriba 

I fully admit having interoperability with NNVM compiler is what must be done in future.

But I fear disposing self-contained nature might be too drastic. So I'd like to reuse current implementation for the sake of preserving header-only, dependency free, stand-alone traits of this library .

Therefore, I suppose we need to separate front-end and back-end of this library further to achieve those different objectives.

As I wrote in previous comments, replacing `tensor_t` to something else within this library seems not very easy. So I say using dlpack's DLTensor looks favourable but not achievable as a short-term goal.

But I think that's OK as long as this library will not directly handle TVM Runtime. If my understanding is correct, only TVM Runtime is DLPack compatible. NNVM compiler is backed by TVM stack but NNVM C/C++ APIs aren't.

And let's not talk about using TVM runtime from this library as we already have many objectives.

beru:
> But I think that's OK as long as this library will not directly handle TVM Runtime. If my understanding is correct, only TVM Runtime is DLPack compatible. NNVM compiler is backed by TVM stack but NNVM C/C++ APIs aren't.

This might be wrong. nnvm python code uses nnvm C-api and tvm python code which means compilation stage is handled by TVM.

beru:
@bhack @edgarriba What I want to say is it's not easy to deeply operate with NNVM and TVM with low-level APIs. NNVM frontends are written in Python but tiny-dnn doesn't have Python API.

bhack:
Another alternative initiative that was backed also by Tensorflow is https://arxiv.org/pdf/1711.03016.pdf. Check the paper to see the differences against TVM and XLA.

beru:
@bhack 

> Probably it is not too far. What do you think of dmlc/nnvm#343?

The movement is great but it's still under development and not ready yet. Even if NNVM/TVM C++ compiler API will be ready in future, that doesn't mean this library can automatically interact with it. 

As I grew older, I've come to realize estimating the amount of work with just wild guess and prior knowledge alone does not always lead to favorable results. 

> Another alternative initiative that was backed also by Tensorflow is https://arxiv.org/pdf/1711.03016.pdf. 
> Check the paper to see the differences against TVM and XLA.

It seems to me NNKit and DLVM IR aren't interoperable from C/C++ by design. 
Aside from that, talented compiler devs are awesome.  :open_mouth:

bhack:
Without being connected with a strong upstream project for scheduling and ops tinydnn will be not practical from an operative point of view. 
So for me we could remove xtensor and merge this but actually I don't see any internal roadmap to cover this topics or bandwidth1 to develop internal solutions. If we remove xtensors what other kind of solution we will have?

bhack:
Another solution to evaluate https://ai.intel.com/intel-ngraph/

bhack:
http://ngraph.nervanasys.com/docs/cpp/

bhack:
This a is a nice overview https://arxiv.org/pdf/1801.08058.pdf

beru:
@bhack Then what about reverting #917 ?

If you anyhow want to keep xtensor in this library, that's fine. But I'm not really sure the meaning of having dependency to other library if it's not actually used (at least in master branch) and doesn't replace `tensor_t` at all.

From what you say I think it sounds pretty amazing that keep using the library will may somehow give us NNVM/TVM interoperability in auto manner even though integration development has stopped more than half an year.

beru:
> If we remove xtensors what other kind of solution we will have?

@bhack Changing `tensor_t` from typedef of `std::vector<std::vector<float_t>>` to custom type that is compatible with current code would be good for first step. We need to carry out this task with least efforts because of the heavy use of `tensor_t`.

Actually, it's enjoyable to see names of many good looking libraries/frameworks/compilers. But without the actual process, I wonder what the outcome would be.

bhack:
Edgar what do you think? Probably xtensor will not solve scheduling and ops at level of this other solutions.

beru:
@bhack I have an impression that there's no strong incentive to remove xtensor so I close this PR.

PR #950, opened 2018-03-23
[char-rnn example] Some fixes to make the code failsafe

prlz77:
1. Ignore gitter messages without username.
2. Joining room before downloading is necessary to avoid error message.
3. Fixed the file existence condition from None to os.path.isfile


examples/char_rnn/python/prepare_dataset.py
```diff
@@ -53,13 +53,17 @@ def download_messages(token, chat_room, msg_path):
...
     print("Preprocessing...")
+    messages_ = []
     for idx, message in enumerate(messages):
-        messages[idx]['text'] = preprocessor.process_text(message['text'])
-        messages[idx]['fromUser']['username'] = preprocessor.process_text(message['fromUser']['username'],
-                                                                          newline=False)
+        if "fromUser" in message:
+            messages[idx]['text'] = preprocessor.process_text(message['text'])
+            messages[idx]['fromUser']['username'] = preprocessor.process_text(message['fromUser']['username'],
+                                                                              newline=False)
```
beru:
```py
        if "fromUser" in message:
            message['text'] = preprocessor.process_text(message['text'])
            message['fromUser']['username'] = preprocessor.process_text(message['fromUser']['username'], newline=False)
```

@prlz77 How about rewriting the code in this way?

PR #968, opened 2018-06-25
Optionally use system-wide installation of gtest/gmock if available

morxa:
Distributions such as Fedora do not allow internet access while building a package. Thus, add an option to use the system's gtest/gmock for testing. The default is to fetch a copy, as before. If `USE_SYS_GTEST` is set to `ON`, look for a system-wide installation of gtest and gmock, and use that if available. Otherwise, fall back to using the local copy.

This is part of an effort to package tiny-dnn for Fedora.

The PR is based on the latest alpha release, I can rebase if necessary.


CMakeLists.txt
```diff
@@ -401,6 +402,7 @@ if(BUILD_EXAMPLES)
 endif(BUILD_EXAMPLES)
 
 if(BUILD_TESTS)
+    enable_testing()
```
beru: I think this change isn't needed since it's already written on https://github.com/tiny-dnn/tiny-dnn/blob/a8de26ad40d955908c5ec4fc946c3c67dd381c6c/test/CMakeLists.txt#L2

CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ option(USE_IMAGE_API  "Build tiny-dnn with Image API support"       ON)
 option(USE_GEMMLOWP   "Build tiny-dnn with gemmlowp support"       OFF)
 
 option(BUILD_TESTS      "Set to ON to build tests"              OFF)
+option(USE_SYS_GTEST  "Set to ON to use a system installation of gtest" OFF)
```
beru: @morxa I like when people ignore the beauty of indentation. That's how they optimize their precious time.

CMakeLists.txt
morxa: But `enable_testing()` needs to be called in the root CMakeLists.txt:
```
$ git checkout v1.0.0a3
[...]
$ mkdir build; cd build
$ cmake .. -DBUILD_TESTS=ON -DCMAKE_INSTALL_PREFIX=$PWD/install
[...]
$ make
[...]
$ make install
[...]
$ ctest .
Test project /tmp/tiny-dnn/build
No tests were found!!!
$ git cherry-pick b1bef
$ cmake ..
$ ctest .
Test project /tmp/tiny-dnn/build
    Start 1: all_tests
[...]
```

CMakeLists.txt
morxa: As I said in the original PR, I can rebase onto master if necessary.

CMakeLists.txt
beru: > But enable_testing() needs to be called in the root CMakeLists.txt

I didn't know about that. It seems you're correct according to the cmake manual. 

https://cmake.org/cmake/help/latest/command/enable_testing.html
> Therefore, this command should be in the source directory root.

Then how about removing duplicating command in `test/CMakeLists.txt`?
But I checked that isn't the the cause of multiple `run_tests` projects. :(
Need to investigate more...

CMakeLists.txt
beru: It would be great if you would do that. I see `CMakeLists.txt` is conflicting, not sure why though.

Unfortunately (or fortunately?), there're many tests required to pass before merging pull requests. So it can be a daunting process. Please rely on your coffee anytime.

morxa:
I rebased onto master. The Travis failure does not seem to be related to this PR, is this working on master? I'm not sure about AppVeyor, but I don't have access to a Windows machine, so I can't really test this.

beru:
@morxa 

> I rebased onto master. The Travis failure does not seem to be related to this PR, is this working on master? I'm not sure about AppVeyor, but I don't have access to a Windows machine, so I can't really test this.

I mostly use Windows 10 and Visual Studio 2017. I wish Microsft open sources them.

By the way, I tweaked some CMakeLists.txt files based on your branch.

[CMakeLists.txt](https://github.com/tiny-dnn/tiny-dnn/files/2295623/CMakeLists.txt)

- Removed `enable_testing()` command inside `if(BUILD_TESTS)` in the root `CMakeLists.txt` file. It may probablly be against the rule but that's how it used to be.

[test/CMakeLists.txt](https://github.com/tiny-dnn/tiny-dnn/files/2295624/CMakeLists.txt)

- Revived double `enable_testing()` commands. I'm not really sure how they're necessary. But that's how it  was originally written...
- Changed 2 spaces indentation to 4 spaces indentation for uniformity. I aplogize if this against your religion.

beru:
Talking about travis-ci failed osx build jobs.
https://travis-ci.org/tiny-dnn/tiny-dnn/jobs/416831235#L4655

This can be the solution. But meh.... why it has to be this hard to prepare environment on osx.
https://superuser.com/questions/1044130/why-am-i-having-how-can-i-fix-this-error-shell-session-update-command-not-f

edgarriba:
@morxa @beru can we try to mimic what the xtensor guys are doing for testing with GTest ?
https://github.com/QuantStack/xtensor/blob/master/test/CMakeLists.txt

beru:
@edgarriba 
> can we try to mimic what the xtensor guys are doing for testing with GTest ?

It would be helpful if you could explain which part of the CMakeLists.txt should be updated.
In other words, more details instructions are needed.

edgarriba:
I close this PR in favor of https://github.com/tiny-dnn/tiny-dnn/pull/987

Issue #949, opened 2018-03-15
Error C3203: unspecialized alias template can't be used as a template argument


fuchstraumer:
This is an issue that came up in xtensor not so long ago, however it's easily fixed as that's what
this PR acheives: https://github.com/QuantStack/xtensor/pull/492

Copy and pasting the code from that PR into `xcontainer.hpp` fixes the issue, otherwise the following is the output of MSVC:
```
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(650): error C3203: 'layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(659): error C3203: 'layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1tinydnn\tiny_dnn\xtensor\xcontainer.hpp(668): error C3203: 'const_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(674): error C3203: 'const_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(680): error C3203: 'const_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(689): error C3203: 'const_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(698): error C3203: 'reverse_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(707): error C3203: 'reverse_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
1>tinydnn\tiny_dnn\xtensor\xcontainer.hpp(717): error C3203: 'const_reverse_layout_iterator': unspecialized alias template can't be used as a template argument for template parameter '_Ty2', expected a real type
[...]

PR #987, opened 2018-08-26
Change test framework to Catch2

beru:
Hi, this PR changes test framework to Catch2.

I also removed unfinished tensor implementation.


beru:
@edgarriba @hzxie Please review this PR.

test/test_quantized_convolutional_layer.h
```diff
@@ -93,24 +88,24 @@ TEST(quantized_convolutional, fprop) {
...
+        EXPECT_NEAR(0.8273983, out[8], 5e-2);
+        EXPECT_NEAR(-0.192101, out[9], 5e-2);
+        EXPECT_NEAR(1.2309504, out[10], 5e-2);
+        EXPECT_NEAR(2.2292108, out[11], 5e-2);
+        EXPECT_NEAR(0.5300441, out[12], 5e-2);
+        EXPECT_NEAR(1.7407004, out[13], 5e-2);
+        EXPECT_NEAR(1.6345025, out[14], 5e-2);
+        EXPECT_NEAR(0.6362420, out[15], 5e-2);
+        EXPECT_NEAR(3.4186277, out[16], 5e-2);
+        EXPECT_NEAR(-0.489455, out[17], 5e-2);
```
beru: Please note that I had to adjust these `abs_error` values to pass tests.

beru:
@edgarriba 

> can you move the in house tensor files to a /root/graveyard ? There's not a final decision on what we are doing with tensors so I wouldn't just through away the code.

How about splitting the commit? So that reverting the change would be easy.

I'm not sure if adding `root/graveyard` folder and putting tensor implmentation files in there is a good idea.
The git repository contains past information so I thought leaving the files is unnecessary.

beru:
Maybe this PR should not change in-house Tensor implementations as it makes this PR complicated.

I removed it since I can't build the test project with VS2017 (which I use on Windows) and I've always thought having unfinished and unused implementation is meaningless.

Adding this patch https://github.com/tiny-dnn/tiny-dnn/issues/949#issuecomment-391408664
can be an workaround. I can send another PR to do so. But again, why keep having the files?

edgarriba:
is work and effort from other people. We cannot remove it as simple as that, even when there's no a decision yet on this. Yes, please split PRs.

tiny_dnn/xtensor/xcontainer.hpp
```diff
@@ -124,6 +124,26 @@ class xcontainer : private xiterable<D> {
...
+
+  template <class S, layout_type L>
+  using broadcast_iterator = xiterator<typename iterable_base::stepper, S, L>;
+  template <class S, layout_type L>
+  using const_broadcast_iterator = xiterator<typename iterable_base::const_stepper, S, L>;
+  template <class S, layout_type L>
+  using reverse_broadcast_iterator = std::reverse_iterator<broadcast_iterator<S, L>>;
+  template <class S, layout_type L>
+  using const_reverse_broadcast_iterator = std::reverse_iterator<const_broadcast_iterator<S, L>>;
+#else
```
beru: Please note this change is necessary for VS2017.

It's taken from https://github.com/tiny-dnn/tiny-dnn/issues/949#issuecomment-391408664 and adjusted.

beru:
> is work and effort from other people. We cannot remove it as simple as that, even when there's no a decision yet on this.

Their hard efforts should be respected but just keeping the unused files will not change the situation.

Looking back the past development, we discussed a lot about tensor implementation but we still couldn't finish its development. Later, you introduced third party xtensor library [here](https://gitter.im/tiny-dnn/developers?at=58ce5f0cf7f7d48104320934), and we've decided to use it.

The adoptation was even [announced](https://twitter.com/SylvainCorlay/status/871798514304548869) but still the the development didn't progress and later, Randl took part during gsoc2017 but he couldn't finish it in time without much help and no one has continued it for about a year.

I wish to see some changes in 2018. So let me implement in-house tensor class. I'll make its base class abstract so that we can use third party libraries such as xtensor or eigen.

beru:
@edgarriba 

It's really difficult to estimate amount of work beforehand. Having failure is a common thing in development process. The task had to be divided. I added a lot of pointer handling code so the fault should be mine.

Discussing plans with third-party collaborators is fine. But in the end, somebody has to implement it. Will they implement tiny-dnn library code for us? If so, you must be really good at finding philanthropists.

beru:
@edgarriba Please review this PR again. If it's alright, please merge it.

PR #1013, opened 2019-02-20
Create softclip Activation Layer

dmilos:
Added softclip Activation Layer.
See: https://en.wikipedia.org/wiki/Activation_function, Soft Clipping.


edgarriba:
@dmilos thanks for the contribution. Some tests are failing. Besides, you should provide some sort of unit tests to ensure that forward and backprop works as expected.

dmilos:
You may discard this PR.
New version is available as #1014

dmilos:
Closed due to new version #1014

PR #1014, opened 2019-02-24
Create Soft Clipping Layer

dmilos:
Added Soft Clipping Activation Layer. V2.
See: https://en.wikipedia.org/wiki/Activation_function, Soft Clipping.


test/test_softclip_layer.h
```diff
@@ -0,0 +1,77 @@
+/*
+    Copyright (c) 2019, Dejan Milosavljevic and the respective contributors
+    All rights reserved.
+
+    Use of this source code is governed by a BSD-style license that can be found
+    in the LICENSE file.
+*/
```
beru, line 7: I think it's preferrable to use uniformed copyright notice throughout library files.
Can you use below copyright notice?

```cpp
/*
    Copyright (c) 2013, Taiga Nomi and the respective contributors
    All rights reserved.
    Use of this source code is governed by a BSD-style license that can be found
    in the LICENSE file.
*/
```

By the way, you can update `AUTHORS` file with `scripts/update-authors.sh` file.

dmilos: OK, I'll fix that.
~~Also I'll discard this PL after create new one for this feature.~~
Appears that I have small bug in JSON string in test_serialization.h.

Fixed:Copyright Notice and JSON.
Now it is ready.

PR #1048, opened 2022-04-17
Add new vec2rgbimage function.

TrevorBlythe:
Turns a vector into rgb image simple as. Nice for anyone looking to output images from their neural networks

example:
image<float> img = vec2rgbimage<float>(vec,50,50);

I know it works because this code turns an image to a vector and then turns it back to an image with the function:
 image<float>* imgf = new image<float>("Quandale_dingle.png", image_type::rgb);
   vec_t list;

   for(int i = 0;i<(*imgf).data().size();i++){
      list.push_back(imgf->data()[i]);
   }

   image<float> img = vec2rgbimage<float>(list,50,50);
   img.save("clone.png");

### DIFF
--- a/tiny_dnn/util/image.h
+++ b/tiny_dnn/util/image.h
@@ -457,6 +457,35 @@ inline image<T> vec2image(const vec_t &vec, const index3d<size_t> &maps) {
   return img;
 }
 
+/**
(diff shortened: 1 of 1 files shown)


TrevorBlythe:
this should be default tbh

